PR #5, opened 2016-11-24
Documentation: learning resources

SaschaMann:
Added some learning resources to start with. The Julia website already has an updated list of resources, so it mostly just links to that.


kytrinyx:
This is great. I like that you have the warning about how quickly Julia is evolving at the moment. That might help mitigate some frustration.

PR #8, opened 2017-01-02
CI Test suite (Fixes #3)

### LINKED ISSUE #3: Implement Continuous Integration
Implement a test suite that can run both locally and on Travis CI, that verifies that the example code passes the test suite for each exercise.

Things to consider:

If some tests are marked as skipped or pending in some way, then CI should remove these annotations so that all the tests run against the example, not just the first test.

If we need to rewrite the files in any way, we should copy the exercise to a tmp directory first, so that we don't accidentally end up committing changes that are just for continuous integration or testing purposes.

We should make it possible to test just one exercise at a time, because testing everything can take a while, and very often people are just working on one exercise and will only need to run the full test suite before committing the change.

SaschaMann:
This implements a test suite that can be used to run tests locally and on Travis CI. You can specify the example solutions to be tested via command line arguments. The default behaviour is to test all example solutions.

I've tested it both locally and on Travis CI for Julia v0.5 and the latest v0.6-dev build. It seemed to work but please check if I made any mistakes in `.travis.yml` as I don't have much experience with Travis.

(ref. #3)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,19 @@
-language: bash
+language: julia
+
+os:
+  - linux
+  - osx
+
+julia:
+  - release
+  - nightly
+  - 0.5
+
+matrix:
+  allow_failures:
+    - julia: nightly
 
 script:
-  - bin/fetch-configlet
-  - bin/configlet .
+  - julia runtests.jl
+  - ./bin/fetch-configlet
+  - ./bin/configlet .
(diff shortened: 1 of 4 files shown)


SaschaMann:
@kytrinyx Could you take a look?

kytrinyx:
@SaschaMann This looks excellent! Thank you.

PR #11, opened 2017-01-23
Add exercise: difference of squares

SaschaMann:


SaschaMann:
I will merge this tomorrow if nobody brings up objections or suggestions to this PR until then.

PR #13, opened 2017-01-23
Add exercise: hello world

SaschaMann:


config.json
```diff
@@ -5,7 +5,14 @@
   "active": false,
   "test_pattern": "TODO",
   "exercises": [
-
+      {
+          "slug": "hello-world",
+          "difficulty": 1,
+          "topics": [
```
andrej-makarov-skrt: Maybe Generics and Variables here to add?

SaschaMann: I think those are so general that they would apply to practically every exercise and adding them here doesn't add useful info. It is assumed that a user already has some basic Julia knowledge, at least to the point that they understand how to write a method/function.

I'm going to add `string interpolation` though, which covers one aspect of `variables`. Thanks for the suggestion!

PR #14, opened 2017-01-24
Add exercise: leap

andrej-makarov-skrt:


config.json
```diff
@@ -5,7 +5,13 @@
   "active": false,
   "test_pattern": "TODO",
   "exercises": [
-
+    {
+      "slug": "leap",
+      "difficulty": 1,
+      "topics": [
+        "generators"
```
SaschaMann: The example solution doesn't have any generators, perhaps change this to `control-flow (if-else statements)` or `comparisons` or similar?

SaschaMann:
Good idea to add the exercise text in the example! I just added a minor comment about the config.

config.json
andrej-makarov-skrt: @SaschaMann yes, I agree, just used this from your pull request.
Are these three okay: Control-flow (conditionals), Integers, Mathematics?

andrej-makarov-skrt:
I close this due to wrong actions from my side.

config.json
andrej-makarov-skrt: @SaschaMann I made a new one pull request, please check it out.

PR #16, opened 2017-01-25
Add exercise: leap

andrej-makarov-skrt:


SaschaMann:
Looks good! Thanks, @andrej-makarov-skrt

PR #20, opened 2017-01-27
Add exercise: raindrops

andrej-makarov-skrt:


config.json
```diff
@@ -6,6 +6,14 @@
...
     {
+      "slug": "raindrops",
+      "difficulty": 1,
+      "topics": [
+        "control-flow (loops)",
+        "arrays",
+        "strings"
+      ]
+    },
       "slug": "rna-transcription",
```
SaschaMann: You missed a `{` here

exercises/raindrops/runtests.jl
```diff
@@ -0,0 +1,73 @@
+using Base.Test
+
+#include("raindrops.jl")
+include("example.jl")
```
SaschaMann: Needs to be `include("raindrops.jl")` again.

If you run `julia runtests.jl raindrops` in the main directory, it tests the example solution. That way you don't have to manually change it in the exercise-specific test file.

SaschaMann:
Thanks for adding this! 👍

PR #22, opened 2017-01-31
Add exercise: scrabble-score

andrej-makarov-skrt:


SaschaMann:
LGTM, as always ;) Thanks!

PR #23, opened 2017-01-31
Add exercise: sieve

andrej-makarov-skrt:


SaschaMann:
`trues(limit)` instead of `fill(true, limit)` would work here as well. It creates a `BitArray` instead which uses less memory. Not really relevant for this exercise, though, but might be helpful to know.

Thanks for adding this exercise!

andrej-makarov-skrt:
@SaschaMann I thought about BitArray, but went further without using it. Thank you!

SaschaMann:
Could you adjust the indentation for this one, too? Thanks!

andrej-makarov-skrt:
Done! Please check it out.

SaschaMann:
Looks good, thank you.

PR #18, opened 2017-01-25
Add code formatting guidelines & typo fixes

SaschaMann:
Fixes #2

PR #24, opened 2017-01-31
Add exercise: trinary

andrej-makarov-skrt:


SaschaMann:
Nice solution!

Just as a headsup, #18 introduced some code formatting guidelines for the examples. If you add more exercises in the future, please adjust your indentations to those. I'll merge this anyway and then fix all examples at once because it's a less of a hassle this way.

Thanks a lot for adding so many exercises!

andrej-makarov-skrt:
@SaschaMann Alright, saw this guidelines earlier and didn't take care of Atom's default indentation (two spaces for Julia's files).

PR #25, opened 2017-02-01
Add exercise: secret-handshake

andrej-makarov-skrt:


SaschaMann:
Nice solution, thanks!

PR #26, opened 2017-02-02
Fix indentation to follow guide rules

andrej-makarov-skrt:


SaschaMann:
Having consistency for the example code is very nice. Thanks a lot for fixing all of these!

PR #29, opened 2017-02-05
Add exercise: roman-numerals

aawray:


exercises/roman-numerals/runtests.jl
```diff
@@ -0,0 +1,36 @@
...
+
+@testset "basic transformations" begin
+    for sample in samples
+        @test to_roman(sample[1]) == sample[2]
+    end
+end
+
+@testset "error handling" begin
+    @test_throws ErrorException to_roman(0)
+    @test_throws ErrorException to_roman(3001)
```
SaschaMann: A solution that handles numbers higher than 3000 would still be valid, even if it's not required and we don't test for it. Imo failing the tests if a solution goes beyond 3000 isn't ideal.

I agree that there should be some sort of error for numbers that it can't handle, though. Perhaps having a hint or a constant for the maximum number would be a bit less restrictive?

exercises/roman-numerals/runtests.jl
```diff
@@ -0,0 +1,36 @@
...
+    163 => "CLXIII",
+    402 => "CDII",
+    575 => "DLXXV",
+    911 => "CMXI",
+    1024 => "MXXIV",
+    3000 => "MMM"
+)
+
+@testset "basic transformations" begin
+    for sample in samples
```
SaschaMann: Instead of the `begin` with a loop inside, the output of this might be more readable if each sample had its own testset so one can immediately see which test fails in the test output.

exercises/roman-numerals/runtests.jl
aawray: Yes, these tests I made a bit confused me, but then I thought to take a risk and to see what will happen, so that's okay, thank you for opinion. I'll make some changes later. (:

andrej-makarov-skrt: Removed 3001 test and marked second comment as hint. I think that for zero and negative numbers there should be an error.

exercises/roman-numerals/runtests.jl
andrej-makarov-skrt: Changed this one, please check it out.

Just in case of confuse: aawray is also my account here.

exercises/roman-numerals/runtests.jl
SaschaMann: I meant something along the lines of "throw an error if `number` is too high for your function to handle". Your hint is already part of the exercise description.

The example solutions aren't downloaded if somebody fetches an exercise through the exercism CLI, so the hint in `example.jl` would not be immediately visible for a user. Instead you can put it in either a new file `HINTS.md` in the same directory as the exercise or put it in the `roman-numerals.jl` code skeleton.

> I think that for zero and negative numbers there should be an error.

I agree. Perhaps add an additional test for negative numbers, too?

exercises/roman-numerals/runtests.jl
SaschaMann: That works a bit nicer 👍

SaschaMann:
I added a comment to the review thread: https://github.com/exercism/xjulia/pull/29#discussion_r99623099

github hides it as outdated by default for some reason so had to link it here

exercises/roman-numerals/runtests.jl
aawray: Moved hint and added more tests. Thank you!

SaschaMann:
Looks good! Thanks for adding this :)

PR #30, opened 2017-02-05
Add exercise: custom set

SaschaMann:


SaschaMann:
This needs to be updated for v0.6 but since that's still a few months away, I'll update it once v0.6 is a bit more stable. See #37

PR #31, opened 2017-02-05
Add exercise: nucleotide count

SaschaMann:


exercises/nucleotide-count/example.jl
```diff
@@ -0,0 +1,8 @@
+function count_nucleotides(strand::AbstractString)
+    counter = Dict('A' => 0, 'C' => 0, 'G' => 0, 'T' => 0)
+    for sym in strand
+        sym in ('A', 'C', 'G', 'T') || error("Invalid nucleotide in strand")
```
SaschaMann: `sym in keys(counter)` would be cleaner indeed. I'll change it.

SaschaMann:
Changed it to `sym in keys(counter)` as suggested, thanks for feedback!

PR should be squashed on merge.

PR #33, opened 2017-02-07
Add exercise: transpose

andrej-makarov-skrt:


exercises/transpose/example.jl
```diff
@@ -0,0 +1,20 @@
...
+    result = fill("", rows)
+    for i = 1:rows
+        for j = 1:cols
+            sym = ""
+            if i <= length(input[j])
+                delta = j-length(result[i])-1
+                if delta > 0
+                    sym = " " ^ delta
+                end
+                sym = sym * string(input[j][i])
```
SaschaMann: `sym *= string(...)` would work as well, looks a bit cleaner imo but that's personal taste.

SaschaMann:
Looks good! Thanks for adding so many exercises :)

aawray:
So that is what I do just using regular expressions parsing JSON file. :)

exercises/transpose/example.jl
aawray: Yes, thank you!

SaschaMann:
Could you upload/share that script? Would be very useful to have.

andrej-makarov-skrt:
Here's Python script that must be changed per exercise: https://gist.github.com/andrej-makarov-skrt/492f133372e8c3d42c5cbad168253375

I use regexp in Atom editor for fast string replacements.

SaschaMann:
Thanks!

andrej-makarov-skrt:
@SaschaMann updated the script, I used it for atbash-cipher exercise, so you can check it out as an extended example.

PR #35, opened 2017-02-07
Add exercise: luhn

SaschaMann:


exercises/luhn/example.jl
```diff
@@ -0,0 +1,16 @@
+function luhn(id::AbstractString)
+    id = split(replace(id, ' ', ""), "")
+    length(id) < 2 && return false
+    all(isdigit, id) || return false
```
andrej-makarov-skrt: Lines 3-4 can be merged in one line with one return statement, but it will be harder to read.

SaschaMann: True. I'll leave the more readable version.

PR #43, opened 2017-02-15
Add exercise: robot-name

andrej-makarov-skrt:


exercises/robot-name/robot-name.jl
```diff
@@ -0,0 +1,11 @@
+type Robot
```
SaschaMann: This exercise is a bit more advanced, so I think providing less help in the skeleton would make sense.

This is one of the first exercises that introduces types though, so some guiding towards them is probably a good idea. Perhaps just having the function skeleton for `reset(instance::Robot)` is enough of a hint that the user needs to implement his own type?

SaschaMann:
This will definitely need to be ported to v0.6 in the future due to changes to the type system. I'll create a project/milestone to collect those exercises later, so that we can easily find them.

exercises/robot-name/robot-name.jl
```diff
@@ -0,0 +1,11 @@
...
+    name::AbstractString
+
+    # Constructor
+    Robot() = new("")
+end
+
+function reset(instance::Robot)
```
SaschaMann: This changes `instance` in-place so it should be called `reset!` by convention.

exercises/robot-name/robot-name.jl
andrej-makarov-skrt: I decided to leave empty structure and method skeletons. Created guidelines with small help. :)

exercises/robot-name/HINTS.md
```diff
@@ -0,0 +1,17 @@
+This exercise introduces work with Julia's [type system](http://docs.julialang.org/en/stable/manual/types/)
+and random [numbers](http://docs.julialang.org/en/stable/stdlib/numbers/).
+
+Let's create a structure that will describe our robots
+and then use it to create new instances and to work with them.
+
+Generate the robot's name using randomness and don't worry about
+collisions, tests sometimes will fail and that's alright.
```
SaschaMann: Perhaps add a comment that one should not worry about collisions at first but then try to resolve it later?

SaschaMann:
Thanks @andrej-makarov-skrt

PR #44, opened 2017-02-15
Add exercise: rotational-cipher

SaschaMann:


SaschaMann:
Should be squashed on PR to remove the empty WIP commit

exercises/rotational-cipher/runtests.jl
```diff
@@ -0,0 +1,51 @@
...
+end
+
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+# Additional exercises                                                #
+# Remove the comments for the optional bonus exercises from HINTS.md  #
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
+# Bonus A
+# @testset "string literal R13" begin
+#     @test R13"The quick brown fox jumps over the lazy dog." == "Gur dhvpx oebja sbk whzcf bire gur ynml qbt."
```
aawray: You can use @test_skip as it is used in other language's test sets, just to conform common style (may it be?).

aawray: Oh, just saw the issue regarding @test_skip and Travis.

SaschaMann: I will try it but it threw an Error anyway on my end, so I thought Travis will fail the build.

SaschaMann: Indeed, it throws a LoadError and fails the build.

andrej-makarov-skrt: @test_skip behavior is a bit confusing, there must be discussion or we should make another macro for test skipping like @skip_test :)

SaschaMann: Good idea, I'll leave this PR open until we have a solution to #45

SaschaMann:
I will merge this now with the commented out tests and will revisit it in v0.6 once `@test_skip` behaves properly. Would be good to get some user feedback on how they'd like optional tests to be handled, too.

PR #48, opened 2017-02-18
Add: logo files

alemelis:
png and svg in img folder


SaschaMann:
Looks good, although the outer circle isn't there in most Julia logos.

@kytrinyx could you take a look if this fits the usual exercism-logo requirements?

kytrinyx:
This works, either with or without the outer circle.

alemelis:
indeed, the outer circle was an artistic license I took
I'm happy to modify the logo and take it out

SaschaMann:
I'm fine with the circle. It looks good, thanks for creating the logo! :)

Just to be sure: There shouldn't be any license issues because it's so different to the original Julia logo, right?

kytrinyx:
> There shouldn't be any license issues because it's so different to the original Julia logo, right?

I'm pretty sure _loosely inspired by_ is fine. For future reference, though, what is the license on the Julia logo?

SaschaMann:
I can't find any info on the logo itself but the julialang.org contents (does that include the logo?) is licensed under the MIT license: https://github.com/JuliaLang/julialang.github.com/blob/master/LICENSE.md

In the past people have just asked Stefan Karpinski (creator of the logo) directly for permission to use it, so if we want to be 100% sure, I can send a mail and ask.

kytrinyx:
> In the past people have just asked Stefan Karpinski (creator of the logo) directly for permission to use it, so if we want to be 100% sure, I can send a mail and ask.

I think it would be worth asking, making sure to mention that that this doesn't represent the language Julia, but the language track on Exercism.

I might be overly cautious.

SaschaMann:
He's given us the permission to use it.

@kytrinyx I forwarded you the email in case you want a record of it.

kytrinyx:
Sweet, thanks so much!

alemelis:
julia's wikipedia [page](https://commons.wikimedia.org/wiki/File:Julia_prog_language.svg) links to [this discussion](https://groups.google.com/forum/#!topic/julia-dev/g2fe8L9dkwg) about the logo and its license 

it seems they are using the same policy of python

SaschaMann:
I've added the attribution/licensing info the README & will merge this on tuesday unless there are any objections.

Thanks a lot for creating a logo for our track, @alemelis

alemelis:
@SaschaMann my pleasure :)

SaschaMann:
@alemelis The SVG appears to have unused whitespace around it. The logo appears smaller than the other icons on the language list. Could you adjust that?

alemelis:
@SaschaMann you're right, sorry about that! The svg should now be fixed

PR #49, opened 2017-02-18
Add exercise: etl

aawray:


andrej-makarov-skrt:
Not sure if input and output must be Dicts. @SaschaMann, is this example okay?

SaschaMann:
@andrej-makarov-skrt I'll take a look in a bit, I was busy over the weekend.

SaschaMann:
@andrej-makarov-skrt I think Dicts make more sense for out- and input here. Also Chars instead of 1-letter strings.

aawray:
@SaschaMann changed test sets and example, please check it out.

exercises/etl/runtests.jl
```diff
@@ -0,0 +1,36 @@
...
+end
+
+@testset "multiple scores with multiple letters" begin
+    input = Dict(1=>["A", "E"], 2=>["D", "G"])
+    output = Dict("g"=>2, "e"=>1, "a"=>1, "d"=>2)
+    @test transform(input) == output
+end
+
+@testset "multiple scores with differing numbers of letters" begin
+    input = Dict(1=>[ "A", "E", "I", "O", "U", "L", "N", "R", "S", "T" ],
```
SaschaMann: There are only single letters as key dictionaries, so they should be Chars 'A' etc. instead of Strings

aawray: Done.

SaschaMann:
That works, thanks for adding this exercise! :)

PR #53, opened 2017-03-04
fix the icon file having the wrong name

catb0t:
the icon file should be called icon.png in order for it to
display correctly on http://exercism.io/languages.

the SVG isn't needed and is wasting bandwidth, unless it's
part of a move to SVG I'm not caught up on.


SaschaMann:
Good catch, thanks for letting us know. :) I must have missed that when merging #48, only checked for the directory.

---

In the launch checklist it said that a SVG is optional but doesn't hurt and [this issue](https://github.com/exercism/exercism.io/issues/2910) has some intention to moving to SVG or at least having support for it.

PR #54, opened 2017-03-06
Add v0.5 to Travis testing

SaschaMann:
v0.5.1 has been released. While it should be compatible with v0.5, we should still test v0.5 to be sure as it will likely take some time for people to update.


SaschaMann:
Specified the wrong version in the travis file, will reopen with the correct one.

PR #68, opened 2017-06-18
Remove leading 'x' from repository name

kytrinyx:
The leading 'x' is kind of arbitrary. Especially now that we can set
topics on the repositories, we don't need a pattern to distinguish what
is a track or not.

The repository itself has already been renamed. GitHub redirects from
the old name to the new name, so we do not have to rush to fix links to
the old repository name, though we should update them for the sake of
clarity.

See exercism/meta#1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Exercism Julia Track
 
-[![Join the chat at https://gitter.im/exercism/xjulia](https://badges.gitter.im/exercism/xjulia.svg)](https://gitter.im/exercism/xjulia?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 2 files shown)


SaschaMann:
@kytrinyx Will this not cause issues for people who also contribute to [Julia itself](https://github.com/julialang/julia)? The forks would then both be named `username/julia` or can GitHub handle this automatically?

kytrinyx:
> Will this not cause issues for people who also contribute to Julia itself?

By default GitHub will fork it to your account and append a numeral to the repo name (e.g. `username/julia-1` if you already have `username/julia`. Once you have the fork, you can click the "settings" tab and rename it to `username/exercism-julia`.

PR #79, opened 2017-08-05
Add introductory blurb to config.json

SaschaMann:
See https://github.com/exercism/discussions/issues/173.

@robphoenix or @kytrinyx Could you proof-read this, please? :)


config.json
```diff
@@ -1,6 +1,7 @@
 {
   "language": "Julia",
   "active": false,
+  "blurb": "Julia is an open-source high-level, high-performance dynamic programming language designed for technical and scientific computing while also being effective for general-purpose tasks. It is convenient to use for daily work but also runs fast enough to be deployed for high-performance applications.",
```
kytrinyx: It's a little bit on the long side for the blurb—we're aiming at about 30-ish words (this is around 40).

How about moving the high-performance + general purpose out of the first sentence, and focusing the second sentence on those two things?

> Julia is an open-source high-level, dynamic programming language whose sweet spot is technical and scientific computing. It is convenient for day-to-day work and packs a high-performance punch.

docs/ABOUT.md
```diff
@@ -1,11 +1,9 @@
...
 - Easy to use interfaces to call libraries written in other languages, such as [C, Fortran](http://docs.julialang.org/en/stable/manual/calling-c-and-fortran-code/) and [Python](https://github.com/JuliaPy/PyCall.jl), directly.
 - [Multiple dispatch](http://docs.julialang.org/en/stable/manual/methods/#man-methods)
 - A dynamic, nominative and parametric [type system](http://docs.julialang.org/en/stable/manual/types/).
-- Homoiconicity: Julia code can be represented in Julia itself, making it a good language to learn about [metaprogramming](http://docs.julialang.org/en/stable/manual/metaprogramming/).
+- Homoiconicity: Julia code can be represented in Julia itself, which allows for powerful [metaprogramming](http://docs.julialang.org/en/stable/manual/metaprogramming/) uses.
 
 The first public release was in 2012. You can find out more about the motivation behind it in the blog post ["Why We Created Julia"](http://julialang.org/blog/2012/02/why-we-created-julia) by the core developers.
```
kytrinyx: That is a fantastic blog post.

I'd love it if the opening of the ABOUT could lead with the sentiments from the blog post.

Maybe something like:

> The creators of Julia want to eat their cake and have it too. As they describe in their blog post [Why We Created Julia](http://julialang.org/blog/2012/02/why-we-created-julia) they want the speed of C, the dynamism of Ruby, the familiar mathematical notation of Matlab. They want it to be their favorite things from their favorite languages. String processing like Perl. Glue like the shell. Powerful but not impenetrably complex. Country _and_ western.

In short, they want _[s]omething that is dirt simple to learn, yet keeps the most serious hackers happy_.

I also think that if we could rework the bullet points to be a couple of paragraphs of prose, that would read more smoothly.

SaschaMann:
It took me a while to get back to this due to holidays and exams, sorry!

Thanks for your feedback, I've tried to update the texts accordingly.

SaschaMann:
Thanks for proofreading!

> Also, that Julia blog post is seriously the best I've ever read by language creators about why they created it.

Indeed. And the best part is that they even manage to achieve many of those super greedy goals with their language :)

PR #80, opened 2017-08-05
Add progression through exercises

SaschaMann:
See https://github.com/exercism/discussions/issues/175.

First attempt at a progression order through the track. Looks like this:

image

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,7 +64,7 @@
     {
       "uuid": "9474a13b-09c1-4e4d-90e3-1a14fde5eb62",
       "slug": "bob",
-      "core": true,
+      "core": false,
       "unlocked_by": null,
       "difficulty": 2,
       "topics": [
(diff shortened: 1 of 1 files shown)


jonmcalder:
Cool! FYI, my understanding is that (for now at least) only core exercises should unlock other exercises. 

e.g. from exercism/discussions#75:

>  Aside from those initial optional exercises, all other optional exercises (core: false) should be unlocked_by the slug of a (a single) core exercise.

You've got sequential unlocking in `nucleotide-count` -> `scrabble-score` -> `etl`

SaschaMann:
@kytrinyx' PR #73 mentioned the following:

> Non-core exercises depend on only one exercise (unlocked_by: ). At the moment we are assuming that this is a core exercise, but there is no reason that it needs to be.

so I thought such progression would also be fine. Javascript also uses it according to the visualisation example.

--- 

Added some non-core, non-bonus exercises so we have the following:
> If possible, leave 3 or 4 simple exercises as (core: false, unlocked_by: null), as this will provide new participants with some exercises that they can tackle even if they have not finished the first core exercise.

jonmcalder:
Ah ok cool. I must have overlooked that somehow! My apologies.

kytrinyx:
I'm pretty sure what you have here will work. Once we get the prototype launched we'll be able to test it out and tweak it if not.

PR #82, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,3 @@
+function hello()
+    return "Hello, World!"
+end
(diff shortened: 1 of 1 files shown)


SaschaMann:
Changed it to be more Julian.

Hello World looks so boring and doesn't show anything, would a more complex snippet that actually shows the language be allowed as well?

Along the lines of
```julia
function julia_set(z, c)
    n = 0
    while abs2(z) < 4
        z = z^2 + c
        n+= 1
    end
    n
end
```

kytrinyx:
This is much, much better!

PR #83, opened 2017-08-27
resources: fix link to Style Guide

fsouza:


SaschaMann:
Good catch, thanks for fixing it! :)

Issue #87, opened 2017-10-05
Implement missing exercises


SaschaMann:
There are [many exercises](http://exercism.io/languages/julia/todo) that still need to be added to the Julia track. If you want to work on one, you can leave a comment here to "claim" it or submit an empty WIP Pull Request. More info is also available in the README of this repo. If you have any questions, don't hesitate to ask! :)

leo60228:
#170

PR #89, opened 2017-10-07
Implement Grains exercise

HarrisonGrodin:
Contributes to exercism/julia#87. Loosely based on [Python implementation](https://github.com/exercism/python/tree/master/exercises/grains).

### DIFF
--- a/config.json
+++ b/config.json
@@ -362,7 +362,7 @@
       "uuid": "2ed6e171-04e1-b080-f383-a801b2564b3e0f4f1ee",
       "slug": "grains",
       "core": false,
-      "unlocked_by": null,
+      "unlocked_by": "difference-of-squares",
       "difficulty": 1,
       "topics": [
         "exceptions"
--- a/config.json
+++ b/config.json
@@ -2,8 +2,7 @@
   "language": "Julia",
   "active": true,
   "blurb": "Julia is an open-source high-level, dynamic programming language whose sweet spot is technical and scientific computing. It is convenient for day-to-day work and fast enough for high performance computing.",
-  "exercises": [
-    {
+  "exercises": [{
       "uuid": "a668410d-41aa-4710-a68f-54521da6486d",
       "slug": "hello-world",
       "core": true,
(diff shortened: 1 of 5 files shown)


exercises/grains/runtests.jl
```diff
@@ -0,0 +1,27 @@
+using Base.Test
+
+include("grains.jl")
+
+@testset "On squares" begin
+    @testset "On square $s" for s = 1:64
+        @test on_square(s) == 2^(s-1)
+        @test total_after(s) == 2^s - 1
```
SaschaMann: This overflows for s = 64:

```julia
julia> 2^64-1
-1
```

This should be Int128 or UInt64 instead

config.json
```diff
@@ -358,6 +357,16 @@
         "mathematics",
         "iterators"
       ]
+    },
+    {
+      "uuid": "2ed6e171-04e1-b080-f383-a801b2564b3e0f4f1ee",
+      "slug": "grains",
+      "core": false,
+      "unlocked_by": null,
```
SaschaMann: I think this could be unlocked by difference-of-squares because they're somewhat similar.

exercises/grains/README.md
```diff
@@ -0,0 +1,27 @@
...
+## For bonus points
+
+Did you get the tests passing and the code clean? If you want to, these
+are some additional things you could try:
+
+- Optimize for speed.
+- Optimize for readability.
+
+Then please share your thoughts in a comment on the submission. Did this
+experiment make the code better? Worse? Did you learn anything from it?
```
SaschaMann: Please leave the source and the comment about incomplete solutions in here.

exercises/grains/example.jl
```diff
@@ -0,0 +1,18 @@
...
+end
+
+"""Calculate the total number of grains after square `square`."""
+function total_after(square)
+    check_square_input(square)
+    sum(map(on_square, 1:square))
+end
+
+"""Validate an arbitrary square."""
+function check_square_input(square)
```
SaschaMann: You could leave both the 0.7 and the 0.6 version in and check which version is currently used. That'll save some time later on when we migrate to 0.7/1.0

SaschaMann:
Few minor things need to be fixed but overall it looks great, thanks for contributing!

SaschaMann:
LGTM, thanks for adding this! :)

PR #94, opened 2017-10-20
Add exercise: run-length-encoding

Teo-ShaoWei:
Another call-to-arm to #87.

### DIFF
--- a/exercises/run-length-encoding/example.jl
+++ b/exercises/run-length-encoding/example.jl
@@ -1,0 +1,22 @@
+function encode(s)
+    (length(s) == 0) && (return s)
+
+    count = 0
+    for char in s
+        (char == s[1]) ? (count += 1) : break
+    end
+    count_str = (count == 1) ? "" : dec(count)
+    return string(count_str, s[1], encode(s[count + 1 : end]))
+end
+
+
+
+function decode(s)
+    function expand(count_str, char)
+        count = (count_str == "") ? 1 : parse(Int, count_str)
+        return char ^ count
+    end
+
+    result = ""
+    return string((expand(rm.captures...) for rm in eachmatch(r"(\d*)(\D)", s))...)
+end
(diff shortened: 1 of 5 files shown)


exercises/run-length-encoding/example.jl
```diff
@@ -0,0 +1,26 @@
+function encode(s)
+    (length(s) == 0) && (return s)
+
+    count = 0
+    for char in s
+        if char == s[1]
```
SaschaMann: This could also use the ternary operator.

SaschaMann:
This would fit fairly well in the "count letters or words"-section of the progression to be unlocked by nucleotide-count, I think. Your thoughts on that?

exercises/run-length-encoding/example.jl
Teo-ShaoWei: Wow thanks for being detailed with the check! =)

This one I intended for the `break` to be visible and thus settled on this structure, but I was perplexed as I spent quite a while thinking about which is better, without a clear winner for me in the end. Ternary with control flow like `break` will make the code terse and appeal to stronger coders but might be off-putting to beginners who will find it too hard to read, kinda like how Haskell is...

I think I can respect your opinion on this. Will edit.

Teo-ShaoWei:
Yes, I agree with you that this is of a similar nature. Let's park it that way. Sorry but how should I correctly implement that? Is it in `config.json`?

SaschaMann:
Yes. You just set `"unlocked_by"` to the exercise slug that unlocks it. You can check if it worked with `confliget tree .`.

exercises/run-length-encoding/example.jl
SaschaMann: Imo the syntax highlighter will make it visible enough but I see your point. In the end it's mostly personal preference.

Teo-ShaoWei:
Ok done as suggested.

PR #96, opened 2017-10-22
Add exercise: triangle

Teo-ShaoWei:
This question has been marked with the topic `mathematics`, but to be able to fully make use of the mathematical beauty of this question, a good understanding of how to chop up the procedure into smaller parts is necessary. So I add in `functional abstraction` too.

### DIFF
--- a/exercises/triangle/runtests.jl
+++ b/exercises/triangle/runtests.jl
@@ -1,0 +1,64 @@
+using Base.Test
+
+include("triangle.jl")
+
+@testset "check equilateral triangles" begin
+    @testset "true if all sides are equal" begin
+        @test is_equilateral([2, 2, 2]) == true
+        @test is_equilateral([0.5, 0.5, 0.5]) == true
+    end
+
+    @testset "false if any side is unequal" begin
+        @test is_equilateral([2, 3, 2]) == false
+    end
+
+    @testset "false if no sides are equal" begin
+        @test is_equilateral([5, 4, 6]) == false
+    end
+
+    @testset "false if invalid triangle" begin
+        @test is_equilateral([0, 0, 0]) == false
+    end
+end
+
+@testset "check isosceles triangles" begin
+    @testset "true if at least 2 sides are equal" begin
+        @test is_isosceles([3, 4, 4]) == true
+        @test is_isosceles([4, 3, 4]) == true
+        @test is_isosceles([4, 4, 3]) == true
+        @test is_isosceles([4, 4, 4]) == true
+        @test is_isosceles([0.4, 0.5, 0.5]) == true
+        @test is_isosceles([0.5, 0.4, 0.5]) == true
+        @test is_isosceles([0.5, 0.5, 0.4]) == true
+    end
+
+    @testset "false if no sides are equal" begin
+        @test is_isosceles([2, 3, 4]) == false
+    end
+
+    @testset "false if invalid triangle" begin
+        @test is_isosceles([1, 1, 3]) == false
+    end
+end
+
+
+@testset "check scalene triangles" begin
+    @testset "true if no sides are equal" begin
+        @test is_scalene([5, 4, 6]) == true
+        @test is_scalene([0.5, 0.4, 0.6]) == true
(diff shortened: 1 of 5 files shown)


SaschaMann:
I'll take a closer look at this tomorrow, don't have time today. Sorry for the delay.

Teo-ShaoWei:
No worries, small thing. I will perform rebasing first.

exercises/triangle/runtests.jl
```diff
@@ -0,0 +1,64 @@
...
+
+    @testset "false if any side is unequal" begin
+        @test is_equilateral([2, 3, 2]) == false
+    end
+
+    @testset "false if no sides are equal" begin
+        @test is_equilateral([5, 4, 6]) == false
+    end
+
+    @testset "false if invalid triangle" begin
```
SaschaMann: IMO this should throw a DomainError, not return false. I know that problem-specs has it differently, but returning something when the function isn't actually defined for the input is bad style imo.

SaschaMann: I am not sure how other tracks handle it, though. Might be worth raising an issue in problem-specifications.

Teo-ShaoWei: Sorry took me a while to reply, needed to think awhile to make sure I know what I am saying.

Hmm well... we are on the same view that if an invalid input is given, a DomainError or equivalent in the language should be thrown. However, imo that is not the case here.

From the point of mathematics, the problem is well defined with answer `false`. The answer specification is done correctly in that sense. But this is up to interpretation it seems. There are languages that the implementer supports this view ([Haskell](https://github.com/exercism/haskell/blob/master/exercises/triangle/examples/success-standard/src/Triangle.hs), [Ruby](https://github.com/exercism/ruby/tree/master/exercises/triangle/.meta/solutions)) and also those who go against ([python](https://github.com/exercism/python/blob/master/exercises/triangle/example.py), [javascript](https://github.com/ThomasZumsteg/exercism-javascript/blob/master/triangle/triangle.js))

From the point of software development, I am actually taking the anti-establishment view that we do not always need to throw unless required via specification even for invalid inputs. I know that for a long time, stronghold like C#, Java, etc encourage to always throw when the input is invalid, but that make things very hard to manage. Even if I am a 3rd-degree user of a piece of code, I might trigger an error under very unforeseen circumstances because the 2nd-degree code I use did not handled it well. Not to mention that the code is diluted by all the boundary controls. Norms usually sound logical when they were first adopted, but not all stand the practicality test of time.

So some modern languages like [Elixir](https://github.com/exercism/elixir/blob/master/exercises/triangle/example.exs) and [Golang](https://github.com/rlmcpherson/exercism-go/blob/master/triangle/triangle.go) and [Elm](https://github.com/exercism/elm/blob/master/exercises/triangle/Triangle.example.elm) etc just do away with throwing exceptions. This design decision resulted in much more reliable web apps. But still, seems like the implementer of Elm and Elixir uses error codes nevertheless >_<

Anyway, I guess this is going to be decided based on opinion rather than facts. So from what I stated above, I would prefer to stick to the current implementation, at least for this question. Would that be ok?

SaschaMann: > From the point of mathematics, the problem is well defined with answer false. The answer specification is done correctly in that sense. But this is up to interpretation it seems.

My view is that those functions are only defined on the set of valid triangles, that's why I thought it should raise a DomainError.

---

> Anyway, I guess this is going to be decided based on opinion rather than facts. So from what I stated above, I would prefer to stick to the current implementation, at least for this question. Would that be ok?

Yeah, that's okay. Your arguments are convincing and I guess this is such a situation where `false` instead of throwing an error doesn't do any harm while it simplifies things immensely. I didn't think of that before. Thanks for looking into how other tracks handle it and writing such a detailed explanation :)

Teo-ShaoWei: Thanks for accepting my suggestion =)

Haha yes I want to make sure I am making the best decisions for us that I can. Let’s keep up the good work!

PR #101, opened 2018-01-15
Rename deprecated hints files

kytrinyx:
The ./exercises/:slug/HINTS.md file has been deprecated in favor of the
./exercises/:slug/.meta/hints.md file.

See https://github.com/exercism/meta/issues/94


kytrinyx:
@SaschaMann I did something gross to get the runtests.jl to not recurse... I'm thinking maybe a better approach would be to use `readdir`, now that I've had a moment to think things through. Let me give that a whirl.

kytrinyx:
Ok, I redid the tweak to the `runtests.jl` and I think this is better... at least not embarrassing :-)

SaschaMann:
`readdir` seems better here indeed. I'll take a closer look tomorrow.

runtests.jl
```diff
@@ -13,32 +13,37 @@ macro test_broken(ex)
     @test eval(current_module(), ex)
 end
 
-for (root, dirs, files) in walkdir("exercises")
-    for exercise in dirs
-        # Allow only testing specified execises
-        if !isempty(ARGS) && !(exercise in ARGS)
-            continue
-        end
+if !isdir("tmp/")
```
SaschaMann: In L32 a temp directory is created for each exercise, this shouldn't be needed or am I missing something?

runtests.jl
```diff
@@ -13,32 +13,37 @@ macro test_broken(ex)
...
+end
+
+for exercise in readdir("exercises")
+    # Allow only testing specified execises
+    if !isempty(ARGS) && !(exercise in ARGS)
+        continue
+    end
+
+    exercise_path = joinpath("exercises", exercise)
+    if !isdir(exercise_path)
```
SaschaMann: For readability, I'd prefer short-circuiting with `||` here, but that's personal preference. This works, too.

runtests.jl
kytrinyx: It might not be anymore, actually. When I was messing with the `walkdir` variation of this the temp directory was being created within `exercises`, and then the code would try to recurse into the temp dir. Let me try this again without it (and also, what on earth was I thinking with the indentation here?)

runtests.jl
kytrinyx: I'll change it to a short circuit. Thanks!

kytrinyx:
It didn't need that top-level `tmp/`, and is much nicer with the short circuit. Thanks, @SaschaMann!

SaschaMann:
I don't understand why the build fails on travis. Locally it works.

```julia
ERROR: LoadError: SystemError: mktempdir: No such file or directory
Stacktrace:
 [1] #systemerror#44 at ./error.jl:64 [inlined]
 [2] systemerror(::Symbol, ::Bool) at ./error.jl:64
 [3] mktempdir(::String) at ./file.jl:329
 [4] macro expansion at /home/travis/build/exercism/julia/runtests.jl:26 [inlined]
 [5] anonymous at ./<missing>:?
 [6] include_from_node1(::String) at ./loading.jl:576
 [7] include(::String) at ./sysimg.jl:14
 [8] process_options(::Base.JLOptions) at ./client.jl:305
 [9] _start() at ./client.jl:371
while loading /home/travis/build/exercism/julia/runtests.jl, in expression starting on line 16
```

This error doesn't make sense to me. Do you have an idea?

kytrinyx:
Oh! I think I know what it is.

CNOT:
`mktempdir` manual says this:

> Create a temporary directory in the parent directory and return its path. If
>   parent does not exist, throw an error.
> 

I assume if you don't have the privilege to access the parent directory and make a file, it will throw an error. That's my guess.

kytrinyx:
Nah, I think it's because I was still trying to create the tempdir within `tmp/` but on CI we never made `tmp/` whereas locally it existed because we had run the script before I removed the line that created it.

SaschaMann:
Ah okay, that explains it. :)

Thanks for the PR!

kytrinyx:
Thanks for your patience :-)

PR #108, opened 2018-04-07
Nucleotide Count: expect DomainError instead of ErrorException

daveyarwood:
In [my first cut of a solution](http://exercism.io/submissions/8ee2cc5ec49a464ba3d553c5a46a78ff) to this problem, I complained in a comment about the fact that the tests expect specifically an ErrorException, when a more specific error (KeyError) would be thrown otherwise.

Then @SaschaMann offered that maybe a DomainError is more appropriate, if we have to specify the type of the exception being thrown. I like that idea, so here is a PR that makes the change. I'm open to further discussion if anyone has thoughts on this.


SaschaMann:
Looks good! Thanks for bringing this up in the comment and improving the exercise via a PR! :)

daveyarwood:
:beers:

PR #114, opened 2018-06-20
Rewrite ABOUT.md to be less technical

SaschaMann:
* Change the introduction to be less technical and more friendly (#112)
* Extend text mentioning base library with standard library (v0.7 change)

I didn't change a lot but the reader isn't hit by quite as much type-system-lingo anymore.

@kytrinyx Do you think this is better suited as an introduction or should I cut it down a bit more?


SaschaMann:
Thanks for the feedback 👍

PR #115, opened 2018-06-20
added spiral-matrix

CNOT:
Added the files for the spiral-matrix question.


exercises/spiral-matrix/example.jl
```diff
@@ -0,0 +1,9 @@
+#Credit to u/SingularInfinity on Redit: https://www.reddit.com/r/dailyprogrammer/comments/6i60lr/20170619_challenge_320_easy_spiral_ascension/djb8po0
+
+function spiral_matrix(n::Int)
+    n < 0 && throw(DomainError("n should be positive"))
```
SaschaMann: You could add a test case for this. Even if it isn't in the `canonical-data.json`, it definitely makes sense to have this.

CNOT: I'm not convinced it's necessary to check that error. In most cases a negative matrix size should throw an error of some sorts.

SaschaMann: Okay, fair enough. I'll merge it, thanks for adding this! :)

PR #118, opened 2018-08-01
[RFC] Add IJulia notebook version of the exercises

SaschaMann:


SaschaMann:
The conversion package is ready, but I'll wait for https://github.com/JuliaLang/IJulia.jl/issues/693 to be fixed before merging as IJulia currently doesn't work on Windows using Julia 1.0.

PR #120, opened 2018-08-04
Add 'math' topic to relevant exercises in track config

kytrinyx:
The new site lets people filter optional exercises by topic, and this
makes it easy for people to skip past them.

This is important since the math-y exercises are not key to learning
the language, and a lot of people find them intimidating and demotivating.

This updates any 'mathematics' topics to be 'math' since it's shorter
and won't get cropped in the UI.

It also removes the math topics for any exercises that are not in the
canonical math-y list.

exercism/exercism.io#4110

### DIFF
--- a/config.json
+++ b/config.json
@@ -144,7 +144,8 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "generators"
+        "generators",
+        "math"
       ]
     },
     {
@@ -157,7 +158,7 @@
         "control_flow_conditionals",
         "exceptions",
         "integers",
-        "mathematics"
+        "math"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


SaschaMann:
It's really helpful that you add the PRs to all language tracks for changes like this, makes it much easier to keep up. Thanks for that!

---

My thoughts on removing mathsy exercises from the core path, just for future reference:

> This is important since the math-y exercises are not key to learning
the language

> I don't want to make it a rule, because some languages are very math heavy, so they should be free to do so if it makes sense, but I would like to recommend to avoid it in general.

I think this exception definitely applies to Julia. The language is primarily used in scientific and technical computing, which often happen to be quite maths heavy. It would be nonsensical to not have any core exercises that teach the maths features. I've also had the opposite feedback before, where people didn't enjoy all the string processing at the start for the same reasons.

So I'm glad that you've considered that and that it's not a strict rule.

Complex numbers also happens to be the only exercise teaching conversion and promotion of types, which is a pretty important feature to understand and "deserves" being part of the core path, imo.

PR #130, opened 2018-11-07
Add exercise: binary-search

SaschaMann:
Left to do:

- [x] Add tests for bonus tasks


SaschaMann:
For future reference: the expected results are intentionally different from the canonical data in order to stay consistent with Julia conventions.

Issue #140, opened 2019-04-02
grains overflow


dhpollack:
The tests for the grain problem do not properly check whether the solution has accounted for the fact that 64th place causes an overflow.  The user should use an Int128 or UInt128 type.  I adjusted the tests as follows:

```
...
@testset "On squares" begin
    @testset "On square $s" for s = 1:64
        @test on_square(s) == Int128(2)^(s-1)
        @test total_after(s) == Int128(2)^s - 1
    end
end

@testset "Overflow Test" begin
    @test total_after(64) > total_after(30)
end
```

Having said that, I think you should change the tests to not give away the answer.  You don't need to check every square.  Just the first few, and the last few.  The canonical-data.json file does something like this rather than the exhaustive check used here.

SaschaMann:
Thanks for the bug report!

> Having said that, I think you should change the tests to not give away the answer.

That's definitely true, and not something I thought of before. Usually I prefer more general tests over specific examples (like canonical-data tends to have) but giving away the solution is obviously not the right way to handle this.

I probably won't have time to fix this properly this week, but the changes you suggested would fix the immediate issue of overflow checking and strengthen the test suite in general. Could you submit a PR with those changes?

dhpollack:
@SaschaMann I sort of misread your message before I created the pull request.  Do you want me to change it to be exactly what I suggested above?  I actually changed them to the canonical-data tests.

dhpollack:
I actually just created a second pull.  Choose the one you want to use.  :-)

SaschaMann:
The canonical-data tests are better, indeed. Sometimes there are so many it's tedious to manually add them (didn't check the file when I wrote that comment) and the 2nd option would be a quick hotfix, that's why I suggested it.

Thanks for fixing the issue! :)

PR #141, opened 2019-04-03
replaced grains tests with canonical-data tests and fixed example script

dhpollack:
fix for bug related to #140

### DIFF
--- a/exercises/grains/runtests.jl
+++ b/exercises/grains/runtests.jl
@@ -3,10 +3,14 @@ using Test
 include("grains.jl")
 
 @testset "On squares" begin
-    @testset "On square $s" for s = UInt64(1):64
-        @test on_square(s) == 2^(s-1)
-        @test total_after(s) == 2^s - 1
-    end
+    @test on_square(1) == 1
+    @test on_square(2) == 2
+    @test on_square(3) == 4
+    @test on_square(4) == 8
+    @test on_square(16) == 32768
+    @test on_square(32) == 2147483648
+    @test on_square(64) == 9223372036854775808
+    @test total_after(64) == 18446744073709551615
 end
 
 @testset "Invalid values" begin
--- a/exercises/grains/runtests.jl
+++ b/exercises/grains/runtests.jl
@@ -1,14 +1,22 @@
 using Test
 
+# canonical data version: 1.2.0
+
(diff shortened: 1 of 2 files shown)


exercises/grains/runtests.jl
```diff
@@ -3,10 +3,14 @@ using Test
 include("grains.jl")
```
SaschaMann:
```suggestion
# canonical data version: 1.2.0

include("grains.jl")
```

This isn't documented and not consistent among all test suites yet, but adding a version comment makes updating easier later on.

exercises/grains/runtests.jl
```diff
@@ -3,10 +3,14 @@ using Test
...
-        @test on_square(s) == 2^(s-1)
-        @test total_after(s) == 2^s - 1
-    end
+    @test on_square(1) == 1
+    @test on_square(2) == 2
+    @test on_square(3) == 4
+    @test on_square(4) == 8
+    @test on_square(16) == 32768
+    @test on_square(32) == 2147483648
+    @test on_square(64) == 9223372036854775808
```
SaschaMann: Perhaps for the sake of teaching the cases that require an `Int128` should be in their own testset, so that there's some "progression": first one has to solve the algorithm part, then making sure it works for the entire domain.

 I'll leave this decision up to you, though, I'm fine with it as it is, too.

dhpollack: that's actually a good idea.  Do you mind if I change the names of the testsets?  Perhaps calling the first set, "beginning squares" and then "ending squares" for the later tests?

dhpollack: also, if I add a test that's not in the canonical-data should I note that?  I was thinking `total_after(1)` as a possible "edge" case and perhaps `total_after(3)` as a non-edge case since currently the only test for that function is an edge case with the Int128.

SaschaMann: Feel free to rename them.

Additional test cases don't need to be documented, though if they're potentially relevant to other languages  you can open a PR in the problem-spec repo and add them. Both tests sound reasonable to me.

dhpollack:
@SaschaMann alright, should be good to go

SaschaMann:
LGTM, thanks for improving the exercise!

PR #147, opened 2019-05-04
Fix argument of ETL exercise template

marcoesposito1988:


exercises/etl/etl.jl
```diff
@@ -1,4 +1,4 @@
-function transform(input::AbstractArray)
+function transform(input::Dict)
```
SaschaMann: This should be an abstract type, i.e. `AbstractDict`

marcoesposito1988: Sounds good

SaschaMann:
Thanks for fixing it!

PR #155, opened 2019-09-04
New exercise: Cats and dogs

SaschaMann:


SaschaMann:
This exercise is worked on in a different repo to prepare for v3 (https://github.com/exercism/exercism/issues/5088) so I'll close it here.

PR #159, opened 2019-10-29
rotational cipher links fix 

### LINKED ISSUE #160: Julia track - Rotational Cipher exercise has broken links
Exercise's introduction has the following broken links:
[blog post](https://docs.julialang.org/en/stable/manual/metaprogramming/)
[Julia Manual](https://iaindunning.com/blog/julia-unicode.html)

I'm new to metaprogramming those tips could be very useful.

Daniel-EST:
Followed the instructions suggested at the [issue](https://github.com/exercism/exercism/issues/5101) I opened.
Closes #160


SaschaMann:
Thanks for the PR! 

The Travis build is failing because the notebook version of the exercise hasn't been updated. You can either run `julia bin/generate_notebooks.jl` from the root of the repo yourself to do so or I'll do it before I merge the PR tomorrow (can't do it on this PC right now).

SaschaMann:
Thanks!

Teo-ShaoWei:
Oh no sorry I am late to the party!

I refer us to [the README template section](https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/readmes.md).

What happened is that the README for rotaional-cipher is generated using `exercises/rotational-cipher/.meta/hints.md`. Editing the README directly is not recommended as it is a generated document via `bin/configlet generate .`.

If we run that command now, it will keep asking for that line to be reset. Regeneration help to keep the problem statement updated with the latest from https://github.com/exercism/problem-specifications, so we should not skip it.

So the solution should have been to edit the raw `exercises/rotational-cipher/.meta/hints.md`. Then use `bin/configlet generate .` to make the README.

Anyway this is software and there is nothing a PR can't fix =P I will go draft up a new one.

Issue #161, opened 2019-10-31
Generation of exercise READMEs by configlet is not idempotent


Teo-ShaoWei:
From the [instruction to regenerate exercise READMEs](https://github.com/exercism/docs/blob/master/maintaining-a-track/regenerating-exercise-readmes.md), it seems to me that the following
```
bin/configlet generate .
```
should be an idempotent operation. However, this is not the case, and many exercise READMEs will be changed upon running this command on the current state of this repo.

This means that if we are creating a new exercise, we have to always run
```
bin/configlet generate . --only <stub>
```
which is ok but defeat the purpose of the simpler version.

SaschaMann:
The way readmes are generated will change with v3, and I believe there will be better built-in checks for this

PR #162, opened 2019-10-31
Fix exercise README template

Teo-ShaoWei:
Refers to #161.

Update `config/exercise_readme.go.tmpl` to add in the missing newline in the generated README before `## Source`, and remove the extra newline after that block. The new template ensure that the generated README have consistent single newline in the event of missing `.Hints`, `.TrackInsert`, and/or `.Spec.Credits`.

Most READMEs will manually cater for it after generation, but that will make `bin/configlet generate .` be non-idempotent.

Also update READMEs that didn't cater for the missing/additional newline.


Teo-ShaoWei:
This is seriously delicate... I deem it to be the limitation of the templating language =P

SaschaMann:
Thanks! I'm not familiar with the template language, so I'll have to trust you that it works as you outline it :D The output definitely looks better.

Could you run `julia bin/generate_notebooks.jl` to update the README embedded in the Jupyter notebooks, please?

Teo-ShaoWei:
Ah I see, I guess that's why Travis CI complained. Ok, I've done the fixup as per your recommendation. Thanks much for the review!

SaschaMann:
> I guess that's why Travis CI complained.

Yep. It's a bit hidden in the log if you don't know that the check exists.

PR #163, opened 2019-10-31
Update rotational-cipher meta

Teo-ShaoWei:
This is continuation of #159.

Propagate the link update to the raw `.meta/hints.md` so that `bin/configlet generate .` will make the README correctly.

Also took the chance to remove the additional newline.


SaschaMann:
Thanks for the fix! I forgot that the readmes are generated when reviewing #159.

Perhaps we should add a CI check to make sure that the generated readmes are identical to the comitted ones, similar to [bin/check_notebooks](https://github.com/exercism/julia/blob/master/bin/check_notebooks)? (That could go in a separate PR, merging this already)

Teo-ShaoWei:
Thanks for the review =) No worries it happens, we have too much repo to juggle, can't always remember everything about them, we can always fix it afterward.

Adding CI there will be a good idea. Meanwhile, I will continue to work on #161 to get us back to a clean slate.

PR #164, opened 2019-11-01
Regenerate READMEs from problem specifications

Teo-ShaoWei:
Creating a draft PR first, documenting the WIP.

@SaschaMann would like your help to look through only the commit on `Regenerate READMEs from problem specifications` and see whether any of the changes are intended and local to the Julia track. For those cases we can consider moving them into their `<slug>/.meta/hints.md` first while the [formatting is being sorted out](https://github.com/exercism/problem-specifications/issues/1560). If it makes more sense for the description to be edited instead, we can try our luck and do a PR there, then see how haha.


SaschaMann:
Thanks for the PR! 🎉 

Before reviewing it: Not all test suites are updated to the latest problem-spec version (see https://tracks.exercism.io/julia). Regenerating all the READMEs could lead to a divergence between testsuite and exercise description, so we have to be a bit careful. I'll take a look at the changes this weekend.

I've been meaning to add problem-specs as a submodule to make it easier to keep track of versions and changes to problem specs but haven't gotten around to it. This PR looks like a good opportunity to do that, to make updates in the future easier. What do you think?

exercises/isbn-verifier/README.md
```diff
@@ -37,8 +37,6 @@ Now, it's even trickier since the check digit of an ISBN-10 may be 'X' (represen
 
 ## Bonus tasks
 
-* Implement `show(io::IO, isbn::ISBN)` to print the ISBN in the standard format with dashes.
```
SaschaMann: This needs to go into hints or some other local file.

SaschaMann:
Other than that one bonus task, the changes seem fine to me

exercises/isbn-verifier/README.md
Teo-ShaoWei: I see, yes that is true... I see how we can do that

Teo-ShaoWei:
Thanks for the check!

Yes I agree that it is possible for README to diverge from the original, and one way we can document that intention is to add the problem spec repo as a submodule, and then create commit on the change. I am however sadly inadequate in this workflow as I haven't done it before. I think this one if wrongly implemented can really spoil the repo... If you don't mind please give me a general idea on how I should implement it?

SaschaMann:
Sorry for the late response.

I thought about it a bit more, and a problem-specs submodule would make it annoying to only update one exercise at a time to the latest version. It's probably better if we just keep track of the problem-specs version in a comment on top of the testsuite. Some exercise have this already, others will receive it once they're updated.

exercises/isbn-verifier/README.md
SaschaMann: Putting it in the hints file should work, I think

SaschaMann:
@Teo-ShaoWei is this ready?

cmcaine:
Merging with #245.

PR #171, opened 2019-12-02
dnd-character: init

### LINKED ISSUE #170: Implement dnd-character
For Google Code-In

leo60228:
Closes #170


leo60228:
I noticed that most/all of the existing exercises have Jupyter notebooks. I've never used Jupyter before, is this required?

leo60228:
Hm, it looks like the birthday paradox is biting my tests.

SaschaMann:
> I noticed that most/all of the existing exercises have Jupyter notebooks. I've never used Jupyter before, is this required?

You can run `julia bin/generate_notebooks.jl` to generate them from the normal exercises files.

leo60228:
Thanks.

exercises/dnd-character/example.jl
```diff
@@ -0,0 +1,37 @@
+function modifier(ability)
+    return floor((ability - 10) / 2)
+end
+
+function ability()
+    return rand(3:18)
+end
+
+mutable struct DndCharacter
```
SaschaMann: The DND should be capitalised to be consistent with Julia style conventions

exercises/dnd-character/example.jl
```diff
@@ -0,0 +1,37 @@
...
+        return new(
+            strength,
+            dexterity,
+            constitution,
+            intelligence,
+            wisdom,
+            charisma,
+            hitpoints,
+        )
+    end
```
SaschaMann: I think this should be an outer constructor. It doesn't enforce an invariant but rather it's an additional convenience constructor.

That said, it doesn't matter too much for the example here, so I'll leave it up to you.

exercises/dnd-character/runtests.jl
```diff
@@ -0,0 +1,64 @@
+# canonical data version: 1.1.0
+using Test
+
+include("dnd-character.jl")
+
+function ischaracter(dnd)
```
SaschaMann: The parameter is a character, so I think `c` or `character` instead of `dnd` would be a better name.

exercises/dnd-character/runtests.jl
```diff
@@ -0,0 +1,64 @@
...
+    @test modifier(17) == 3
+    @test modifier(18) == 4
+end
+
+@testset "random ability is within range" begin
+    global abilities = []
+    global history = []
+
+    for i=1:3
+        push!(abilities, ability())
```
SaschaMann: The history is saved in `history` and the tests only access `abilities[i]`, I think the `abilities` array isn't needed and could be just a single variable, or am I missing something?

exercises/dnd-character/runtests.jl
```diff
@@ -0,0 +1,64 @@
...
+    @test modifier(14) == 2
+    @test modifier(15) == 2
+    @test modifier(16) == 3
+    @test modifier(17) == 3
+    @test modifier(18) == 4
+end
+
+@testset "random ability is within range" begin
+    global abilities = []
+    global history = []
```
SaschaMann:
```suggestion
    history = Int[]
```

for better performance and following best practices

exercises/dnd-character/runtests.jl
```diff
@@ -0,0 +1,64 @@
...
+            @test abilities[i] >= 3 && abilities[i] <= 18
+            @test !in(abilities[i], history)
+        end
+
+        push!(history, abilities[i])
+    end
+end
+
+@testset "random character is valid" begin
+    global characters = DndCharacter[]
```
SaschaMann:
```suggestion
    characters = DndCharacter[]
```

exercises/dnd-character/example.jl
```diff
@@ -0,0 +1,37 @@
...
+end
+
+mutable struct DndCharacter
+    strength::Integer
+    dexterity::Integer
+    constitution::Integer
+    intelligence::Integer
+    wisdom::Integer
+    charisma::Integer
+    hitpoints::Integer
```
SaschaMann: Ideally you want to avoid fields that have an abstract type (see the [Julia Performance Guide](https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-fields-with-abstract-type-1)), `Int` or some other specific Integer type would be fine.

config.json
```diff
@@ -469,6 +469,22 @@
                 "integers",
                 "math"
             ]
+        },
+        {
+            "uuid": "fcb9ef0e-ed97-4420-a38f-9bf690c23a3b",
+            "slug": "dnd-character",
+            "core": false,
+            "unlocked_by": "robot-name",
+            "difficulty": 3,
```
SaschaMann: `robot-name` unlocks this exercise and has a difficulty of 5, so I think should be at least as high as `robot-name`

PR #189, opened 2019-12-08
Adding Volume-of-sphere Exercise

FrancisJ000:
Added an exercise to find the volume of a sphere.

### DIFF
--- a/exercises/volume of sphere/README.md
+++ b/exercises/volume of sphere/README.md
@@ -1,0 +1,17 @@
+# Volume of Sphere
+
+To find the volume of a sphere
+
+For example:
+input: "1"
+output: "4.1887902047863905"
+
+## Source
+
+The official Julia doc https://juliabyexample.helpmanual.io/
+
+## Version compatibility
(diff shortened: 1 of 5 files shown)


exercises/volume of sphere/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+To find the volume of a sphere
+
+For example:
+input: "1"
+output: "4.1887902047863905"
+
+## Source
+
+The official Julia doc https://juliabyexample.helpmanual.io/
```
logankilpatrick: This is not the official Julia docs.

logankilpatrick: These are: https://docs.julialang.org/en/v1/

exercises/volume of sphere/volume.jl
```diff
@@ -0,0 +1,4 @@
+function sphere_vol(r)
```
logankilpatrick: Why is this function blank?

SaschaMann:
Good call. It's also a bit too simple for an exercise, there's not really anything to discuss in its current state.

PR #198, opened 2019-12-12
Add minesweeper exercise

abhishalya:
This add a new exercise specified at
https://github.com/exercism/problem-specifications/tree/master/exercises/minesweeper.

### DIFF
--- a/config.json
+++ b/config.json
@@ -527,6 +527,24 @@
                 "math",
                 "integers"
             ]
+        },
+        {
+            "slug": "minesweeper",
+            "uuid": "9a47bffd-4dd2-48ce-b47b-753271305966",
+            "core": false,
+            "unlocked_by": "pangram",
+            "difficulty": 2,
+            "topics": [
+                "loops",
+                "conditionals",
+                "variables",
+                "strings",
+                "integers",
+                "lists",
+                "booleans",
+                "arrays",
+                "text_formatting"
+            ]
         }
     ]
 }
--- a/exercises/minesweeper/example.jl
+++ b/exercises/minesweeper/example.jl
@@ -1,0 +1,51 @@
+# function to check if the current character represents a mine or not
+ismine(ch) = (ch == '*' ? 1 : 0)
+
+function annotate(minefield)
+
+    # ans will store our final output of the program
+    ans = []
+
+    # i represents row number
+    for i in 1 : length(minefield)
+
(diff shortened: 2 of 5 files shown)


abhishalya:
Small question: Is there any way we could check if `check_notebook` passes or not?
Since until we add the notebook file to track, the diff will always be emitted by the script. Also, I think we should only be checking diff for `.ipynb` files only rather than checking for all in this script.

exercises/minesweeper/example.jl
```diff
@@ -0,0 +1,51 @@
+# function to check if the current mine represent a mine or not
```
logankilpatrick:
```suggestion
# function to check if the current character represents a mine or not
```

exercises/minesweeper/example.jl
```diff
@@ -0,0 +1,51 @@
...
+            else
+                ans_row *= '*'
+            end
+        end
+
+        # push the calculated row_string to our ans
+        push!(ans, ans_row)
+    end
+
+    ans
```
logankilpatrick:
```suggestion
return ans
```
I guessing this needs to be returned though I am not certain.

abhishalya: I'll add this change to the commit.

exercises/minesweeper/example.jl
abhishalya: Yep :sweat_smile:

exercises/minesweeper/example.jl
logankilpatrick: Cool, you can commit directly from the suggestion if that helps

abhishalya: I'll prefer the old approach :P

config.json
```diff
@@ -527,6 +527,24 @@
                 "math",
                 "integers"
             ]
+        },
+        {
+            "slug": "minesweeper",
+            "uuid": "9a47bffd-4dd2-48ce-b47b-753271305966",
+            "core": false,
+            "unlocked_by":"nucleotide-count",
```
```suggestion
            "unlocked_by": "pangram",
```

`nucleotide-count` already unlocks a lot of exercises and arrays haven't been introduced to the core progression path yet, so let's move it to be unlocked by pangram.

abhishalya:
Thanks, I've included the change requested.

SaschaMann:
Thanks!

PR #201, opened 2019-12-16
Adding in accumulate exercise

logankilpatrick:


logankilpatrick:
Hey @SaschaMann , How do I create a new UUID? Also, is there anything else that I need to change? I was a little confused by the contributing docs.

logankilpatrick:
I am missing the notebook, how do I generate that?

exercises/accumulate/example.jl
```diff
@@ -0,0 +1,23 @@
...
+        new_colletion = collection .* collection
+    elseif operation == +
+        new_colletion = collection .+ collection
+    elseif operation == -
+        new_colletion = collection .- collection
+    elseif operation == /
+        new_colletion = collection ./ collection
+    elseif operation == %
+        new_colletion = collection .% collection
+    end
```
SaschaMann: Untested, but I think this can be simplified to

```julia
new_collection = operation.(collection, collection)
```

exercises/accumulate/runtests.jl
```diff
@@ -0,0 +1,25 @@
+# canonical data version: 1.1.0
+
+ using Test
+
+ include("accumulate.jl")
+
```
SaschaMann: The description uses squares as an example, it would be a good candidate for a test that uses an anon function and not a pre-defined operator.

logankilpatrick: I think I addressed this

exercises/accumulate/example.jl
logankilpatrick: Yes, this is the better way indeed

exercises/accumulate/example.jl
```diff
@@ -0,0 +1,23 @@
+"""
+    accumulate(collection, operation)
+Takes in a collection and applies the operation on that collection then returns
+a new collection.
+"""
+function accumulate(collection, operation)
+
+    new_colletion = []
```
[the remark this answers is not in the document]
logankilpatrick: done!

PR #203, opened 2019-12-19
Add the allergies exercise

DarkoGNU:
I hope everything's gone well
I'm open to suggestions

### DIFF
--- a/exercises/allergies/allergies.jl
+++ b/exercises/allergies/allergies.jl
@@ -1,0 +1,7 @@
+function allergic_to(score, allergen)
+
+end
+
+function allergy_list(score)
+    
+end
--- a/config.json
+++ b/config.json
@@ -572,6 +572,23 @@
                  "integers",
                  "arrays"
              ]
-          }
+        },
+        {
+            "uuid": "4eced190-062f-4a7b-8c92-87ff02f1e343",
+            "slug": "allergies",
+            "core": false,
+            "unlocked_by": "pangram",
+            "difficulty": 1,
+            "topics": [
+                "loops",
+                "conditionals",
+                "integers",
+                "lists",
+                "booleans",
+                "enumerations",
+                "arrays",
+                "bitwise_operations"
+            ]
+        }
     ]
 }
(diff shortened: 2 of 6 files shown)


DarkoGNU:
Oh, sorry for that error

DarkoGNU:
I've just squashed that commit which repaired the notebook with the other commit. I hope I've done the right thing

exercises/allergies/allergies.jl
```diff
@@ -0,0 +1,7 @@
+function allergic_to(score, allergen)
+
+end
+
+function list(score)
```
SaschaMann: Maybe `allergy_list` would be a better name. List is quite generic.

SaschaMann:
Thanks for adding the exercise!

I think the name of `list` should be changed but everything else looks good to me.

> I've just squashed that commit which repaired the notebook with the other commit. I hope I've done the right thing

Yep, you did. Thanks!

DarkoGNU:
Is there anything left to do?

SaschaMann:
> Is there anything left to do?

Nope, looks nice! Thanks again for adding the exercise :)

@logankilpatrick Previously `allergy_list` was called `list`

DarkoGNU:
> Looks good to me. @SaschaMann what do you mean by the List?

@logankilpatrick one question - is force push a good way to make changes to a pull request like this? I'm new to Git and I see that maybe your confusion above wouldn't have arosen if I had done it another way

logankilpatrick:
PM me on slack and we can talk about it!

SaschaMann:
Ideally you should wait with squashing and forcepushing until the reviewing process is complete or until the maintainers ask for it. It's easier to do reviews on short commits a preserved complete dev history, even when the PR will be squashed before or on merging it to master.

And in general, try to use `--force-with-lease` instead of `--force` when possible (it has some safety features) and *never* force push branches that others work on (not the case for most PRs, though).

DarkoGNU:
Understood, thanks!

PR #205, opened 2019-12-27
Add JSONtoTests.jl

soumitradev:
WARNING: Works only for some JSON test data. If you're lucky, this will work flawlessly and might save you hours of work. Else, it will throw an error and stop. You will have to delete the generated runtests.jl in this case. This was actually requested by @SaschaMann as a part of my task, but I didn't find time then, so I added it now.

### DIFF
--- a/generators/JSONtoTests.jl
+++ b/generators/JSONtoTests.jl
@@ -1,0 +1,52 @@
+using JSON, HTTP
+
+currentExercise = ARGS[1]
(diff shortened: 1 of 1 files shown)


SaschaMann:
Thanks for adding the generator!

It'll likely take me a few days before I can try it out and review it, sorry.

soumitradev:
@SaschaMann I'll also try adding comments to the code by this Sunday.  These comments will explain what each block does so it'll be easier for you later. If I find the time, I might try generalizing this a bit so it works on most exercises.

soumitradev:
I deleted my fork :(

See #215

PR #211, opened 2019-12-31
Matching brackets

chinglamchoi:


config.json
```diff
@@ -589,6 +589,24 @@
                 "arrays",
                 "bitwise_operations"
             ]
-        }
+        },
+	{
+	    "uuid":"bf5c1f49-963a-4397-bb67-b58e1b7d45e3",
+	    "slug":"matching-brackets",
+	    "core":true,
```
```suggestion
	    "core":false,
```

For now, we don't want to add any core exercises to not put even more work on the mentors. I'll have to think about where this fits into the progression. I think we can leave it as floating exercise that the students can do whenever they want for now.

exercises/matching-brackets/README.md
```diff
@@ -0,0 +1,5 @@
+# Matching Brackets
```
SaschaMann: To keep a standard format and make maintenance in the future easier, the README should be generated using the configlet. You can find instructions on how to do that here: https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/readmes.md#generating-a-readme

Let me know if you run into an issues or have any questions about it

exercises/matching-brackets/runtests.jl
```diff
@@ -0,0 +1,23 @@
+using Test
```
SaschaMann: Please add a comment `# canonical data version: x.y.z` to the very top of the file. Replace `x.y.z` with the canonical data version you used.

exercises/matching-brackets/runtests.jl
```diff
@@ -0,0 +1,23 @@
...
+
+include("matching-brackets.jl")
+
+@testset "Determine if brackets in expression are correctly matched" begin
+    @test matching_brackets("[]")::Bool == true
```
SaschaMann: - `::Bool` at the end is not necessary
- `matching_brackets("[]")` already returns a `Bool`. The `== true` or `== false` are not necessary.

Two examples:

```julia
    @test  matching_brackets("[]")
    @test !matching_brackets("[[")
```

I like aligning the function name in cases like this to make it easier to see which tests are expected to evaluate to false, and which are expected to be true.

exercises/matching-brackets/runtests.jl
```diff
@@ -0,0 +1,23 @@
...
+    @test matching_brackets("{[]}")::Bool == true
+    @test matching_brackets("{}[]")::Bool == true
+    @test matching_brackets("([{}({}[])])")::Bool == true
+    @test matching_brackets("{[)][]}")::Bool == false
+    @test matching_brackets("([{])")::Bool == false
+    @test matching_brackets("[({]})")::Bool == false
+    @test matching_brackets("{}[")::Bool == false
+    @test matching_brackets("[]]")::Bool == false
+    @test matching_brackets("(((185 + 223.85) * 15) - 543)/2")::Bool == true
+    @test matching_brackets("\\left(\\begin{array}{cc} \\frac{1}{3} & x\\\\ \\mathrm{e}^{x} &... x^2 \\end{array}\\right)")::Bool == true
```
SaschaMann: Please group similar tests together in testsets. There are a few ways to do it. The [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/matching-brackets/canonical-data.json) has some suggestions.

For example you could have testsets for

- tests that contain one pair of brackets
- tests with nested brackets
- tests with brackets behind each other but not nested
- etc.

There's no right or wrong way but at least some grouping should be done here. That also enables the student to work on the exercise in a TDD-inspired way, by making their solution pass simple tests first and then expanding it to work on more complex cases.

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,33 @@
+function matching_brackets(a)
```
SaschaMann: General comments on this file:

- Please set your editor to replace tabs with 4 spaces to follow the style guide from the README.
- Perhaps use more descriptive variable names. It's hard to remember what `a`, `b` or `c` etc. mean when reading the code that's further down. For example, `b` and `c` could be `opening_brackets` and `closing_brackets`.
- Related to the previous point: `a` means a lot of different things throughout the function body. First it's the input string, then it's turned into an array and eventually it's filtered. This causes two issues:
  1. Readability: when reading the function, this confused me at L7-L9 because I still thought `a` was the input. 
  2. Performance: changing the type of variables should be avoided for performance reasons: https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-changing-the-type-of-a-variable-1
  - I would suggest that you instead create multiple variables, or combine the steps into one and rename `a` to be more descriptive. The argument should probably be called `input` regardless

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,33 @@
...
+			end
+		end
+	end
+	if valid == 0 || count != 0
+		return false
+		# parantheses not balanced
+	else
+		return true
+		# parantheses balanced
+	end
```
SaschaMann: You could tidy this up a bit and combine it to one line by using `return <condition>`.

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,33 @@
+function matching_brackets(a)
+	a = split(a, "")
+	b,c = ["(", "[", "{"], [")", "]", "}"]
+	a = filter(i -> (i in b || i in c), a)
+	stack = Array{String}(undef, 0)
+	global d, count, valid = Dict([("(", ")"),(")", "("), ("[", "]"), ("]", "["), ("{", "}"), ("}", "{")]), 0,1
```
SaschaMann: These are only used within the function body, they don't need to be globals.

Perhaps also move `d` into a separate line to make it easier to read.

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,33 @@
+function matching_brackets(a)
+	a = split(a, "")
+	b,c = ["(", "[", "{"], [")", "]", "}"]
+	a = filter(i -> (i in b || i in c), a)
```
SaschaMann: I'm fine with it as it is, but you could make it slightly more readable by using `i in b ∪ c` or `i in union(b, c)`

exercises/matching-brackets/example.jl
chinglamchoi: Implemented the changes about coding style! 

On the 3rd comment, I don't think I've changed the type of `a` (now: `inputt`)? In lines 7-9, I check the length of a to see if it matches a special test cases which allows early exiting. My logic is that if the length of the brackets expression is not divisible by 2, then there is no chance of the expression being balanced.

exercises/matching-brackets/runtests.jl
```diff
@@ -0,0 +1,31 @@
...
+
+@testset "Determine if brackets in expression are correctly matched" begin
+    @test  matching_brackets("[]") 
+    @test !matching_brackets("[[")
+    @test !matching_brackets("}{")
+    @test !matching_brackets("{]")
+    @test  matching_brackets("{ }")
+end
+
+@testset "Determine if brackets in expression are correctly matched" begin
```
SaschaMann: Could you give these sets a unique name? Would be nicer and more useful to the student if the test summary shows what's being tested. Currently it's all the same:

```
[ Info: Testing matching-brackets
Test Summary:                                             | Pass  Total
Determine if brackets in expression are correctly matched |    5      5
Test Summary:                                             | Pass  Total
Determine if brackets in expression are correctly matched |    9      9
Test Summary:                                             | Pass  Total
Determine if brackets in expression are correctly matched |    3      3
```

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,28 @@
+function matching_brackets(inputt)
+    inputt = split(inputt, "")
+    open_brackets,close_brackets = ["(", "[", "{"], [")", "]", "}"]
+    inputt = filter(i -> (i in union(open_brackets, close_brackets)), inputt)
+    stack = Array{String}(undef, 0)
+    d = Dict([("(", ")"),(")", "("), ("[", "]"), ("]", "["), ("{", "}"), ("}", "{")]) #global
```
SaschaMann: A more descriptive name would be nice here as well

exercises/matching-brackets/example.jl
SaschaMann: > On the 3rd comment, I don't think I've changed the type of a (now: inputt)?

It changed from String to an Array:

```julia
julia> function matching_brackets(a)
           @show a, typeof(a)

           a = split(a, "")
           @show a

           b,c = ["(", "[", "{"], [")", "]", "}"]

           a = filter(i -> (i in b || i in c), a)
           @show a
       end
matching_brackets (generic function with 1 method)

julia> matching_brackets("()")
(a, typeof(a)) = ("()", String)
a = SubString{String}["(", ")"]
a = SubString{String}["(", ")"]
2-element Array{SubString{String},1}:
 "("
 ")"
```

`@code_warntype` also shows it and marks the change red (`@_2` are line numbers):

grafik

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,28 @@
...
+        for i in inputt
+            if i in open_brackets
+                push!(stack, i)
+                count += 1 #global
+            elseif count == 0 # & downward: assumes i not in open_brackets
+                valid = 0
+                break
+            elseif stack[end] != d[i] #& downward: assumes count != 0
+                valid = 0
+                break
```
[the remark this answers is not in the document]
chinglamchoi: Not exactly. I was thinking more like XOR instead of OR

exercises/matching-brackets/example.jl
```diff
@@ -0,0 +1,28 @@
+function matching_brackets(inputt)
+    inputt = split(inputt, "")
+    open_brackets,close_brackets = ["(", "[", "{"], [")", "]", "}"]
+    inputt = filter(i -> (i in union(open_brackets, close_brackets)), inputt)
+    stack = Array{String}(undef, 0)
+    d = Dict([("(", ")"),(")", "("), ("[", "]"), ("]", "["), ("{", "}"), ("}", "{")]) #global
+    count, valid = 0,1 #global 
+    if length(inputt) % 2 != 0
+        valid = 0
```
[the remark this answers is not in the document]
chinglamchoi: I'm not sure I completely understand, but i think even if I returned false immediately, I would still need an else statement for cases where the length is not divisible by 2? Apologies for the bad style, I learned programming via competitive programming, where all everyone cares about is speed!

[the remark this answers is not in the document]
SaschaMann: If you `return false`, it stops the execution of the function right there. Everything below it is never reached which is effectively the same as having everything below in an `else` block but more readable:

```julia
julia> function f(x)
           if x == 0
               return false
           end

           println("This is never reached if x == 0")
           return x^2
       end
f (generic function with 1 method)

julia> f(2)
This is never reached if x == 0
4

julia> f(0)
false
```

Using large `else` blocks can easily end up with several nested blocks if there are multiple conditions like this.

> Apologies for the bad style, I learned programming via competitive programming, where all everyone cares about is speed!

No worries, that's what code reviews are for :)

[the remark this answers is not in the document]
SaschaMann: In Julia, you'll often see people use short-circuiting for these conditions, e.g.

```julia
length(inputt) % 2 != 0 && return false
```

or 

```julia
length(inputt) % 2 == 0 || return false
```

exercises/matching-brackets/example.jl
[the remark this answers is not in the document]
SaschaMann: You could use `xor`/`⊻` if you want: `count == 0 ⊻ stack[end] != d[i]`. Check out the help text for it if you want to know more about it.

PR #215, opened 2020-01-11
[WIP] Add JSONtoTests.jl

soumitradev:
> WARNING: Works only for some JSON test data. If you're lucky, this will work flawlessly and might save you hours of work. Else, it will throw an error and stop. This was actually requested by @SaschaMann as a part of my Google Code-in task, but I didn't find time then, so I added it now.

This is actually an addition to #205. I closed #205 because I deleted my fork. :(

Currently, this script only works with 42/48 of the exercises. This is still WIP.

You can check the exercises on which this doesn't work in the latest comment.

Now, the code has comments and a coverage tester is also added. This tester tests the converter on all the 48 (at time of writing) exercises.

NOTE: If you plan to use this, make sure to double-check the runtests.jl file. The code is not very well written IMO, so it could have thousands of bugs.


soumitradev:
Improved code, and formatting. Next up: big change to code that should massively increase coverage for exercises.

generators/JSONtoTests.jl
```diff
@@ -0,0 +1,138 @@
+# This code generates the runtets.jl file automatically for some lucky exercises that work with this script.
```
SaschaMann:
```suggestion 
# This code generates the runtests.jl file automatically for some lucky exercises that work with this script.
```

generators/JSONtoTests.jl
```diff
@@ -0,0 +1,138 @@
...
+
+# NOTE: This is a very experimental version of this idea. The code is not all that well written,
+# and may not work on most exercises. If it works, it might save you hours of time writing that exercise.
+# If it doesn't it will throw an error and quit.
+
+using JSON, HTTP
+
+# Convert an array to a array in string form
+# eg. 4-element Array{Int64,1}: 1 2 3 4 --> "[1, 2, 3, 4]"
```
SaschaMann: Ideally write this as a docstring so that it shows up in the REPL help.

generators/JSONtoTests.jl
```diff
@@ -0,0 +1,138 @@
...
+                push!(str, "$item, ")
+            end
+        end
+
+        # Make it a string and remove the last extra comma and space from array
+        truncated_str = join(str)[1:end-2]
+
+        # Return final array string by closing with bracket
+        return string(truncated_str, "]")
+    end
```
SaschaMann: No need to wrap it in an `else` block due to the `return` in L16.

generators/JSONtoTests.jl
```diff
@@ -0,0 +1,138 @@
...
+    # Get the exercise the user is working on from command line argument
+    current_exercise = exercise_name
+
+    # Download the test data for the exercise the user is working on and parse the JSON.
+    get_json = HTTP.get("https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/$current_exercise/canonical-data.json")
+    json_code = String(get_json.body)
+    data = JSON.parse(json_code)
+
+    # Create an array to push our julia code text to.
+    lines = []
```
SaschaMann:
```suggestion
    lines = String[]
```

generators/JSONtoTests.jl
```diff
@@ -0,0 +1,138 @@
...
+        return string(truncated_str, "]")
+    end
+end
+
+function main(exercise_name)
+    # Get the exercise the user is working on from command line argument
+    current_exercise = exercise_name
+
+    # Download the test data for the exercise the user is working on and parse the JSON.
+    get_json = HTTP.get("https://raw.githubusercontent.com/exercism/problem-specifications/master/exercises/$current_exercise/canonical-data.json")
```
SaschaMann: Not all exercises have canonical data. Perhaps a custom error message for those cases will look nicer than whatever the default error would look like.

generators/JSONtoTests.jl
```diff
@@ -0,0 +1,138 @@
...
+
+        # Make it a string and remove the last extra comma and space from array
+        truncated_str = join(str)[1:end-2]
+
+        # Return final array string by closing with bracket
+        return string(truncated_str, "]")
+    end
+end
+
+function main(exercise_name)
```
SaschaMann: In Exercism lingo, this would be the exercise slug, not name. Some exercises have a "prettified" name that is used on the website. For example, the RNA transcription exercise has the slug `rna-transcription` and the title/name `RNA Transcription`: https://github.com/exercism/problem-specifications/blob/master/exercises/rna-transcription/metadata.yml

generators/JSONtoTestsCoverage.jl
```diff
@@ -0,0 +1,33 @@
+# NOTE: This code is purely to test the coverage of the JSON to test converter.
```
SaschaMann: could you rename the file to `test_JSONtoTests.jl` to make it more clear? When I read coverage, I think of coverage reporting like on codecov and not the tests themselves.

generators/JSONtoTests.jl
soumitradev: This might be causing some errors! Thanks for pointing this out.

generators/JSONtoTests.jl
soumitradev: I'll refactor this in the next commit.

generators/JSONtoTestsCoverage.jl
soumitradev: Sure, I'll do that in the next commit.

generators/JSONtoTests.jl
soumitradev: That's for the non-empty arrays. The return on L16 is a special case handle for an empty array.

SaschaMann: Yes but unless I'm misunderstanding the code, there are two cases here: 

1. The condition of the `if` statement is true, then it will return immediately and everything below that won't be executed.
2. It's false, the entire `else` block will be executed.

Everything below is wrapped in the else block, so you could leave out the else block. Keep the if and return above it, but don't wrap the entire rest inside the else block. This saves an indentation level and is a bit more readable. This is often referred to as "aligning the happy path to the left" or something similar.

```julia
function array_to_str(array)
    # Start array with opening bracket
    str = ["["]
    # If array is empty, return an empty array
    if length(array) == 0
        return "[]"
    end

    # Else, for each item in array, add it to array
    for item in array
        # If item is a string, enclose it in quotations and add it to our array
        if typeof(item) == String
            push!(str, "\"$item\", ")
        # If item is an array, call function recursively to add it to our array
        elseif typeof(item) == AbstractArray
            push!(str, array_to_str(item))
        # Else, if item is something else, just write it to array
        else                
            push!(str, "$item, ")
        end
    end

    # Make it a string and remove the last extra comma and space from array
    truncated_str = join(str)[1:end-2]

    # Return final array string by closing with bracket
    return string(truncated_str, "]")
end
```

It's also possible that I'm misunderstanding or overreading something, but I think this should work.

soumitradev: Aha! I see what you mean now. I’ll add this too in the next commit.

soumitradev:
Still WIP, haven't made the change that will massively increase exercises on which this works (I won't use the word coverage anymore). Sorry for the late reply, I have been really busy with GCI, last 7 days left :)

I'll make the changes ASAP. As you suggested, I ran the tester on all the exercises in problem-specifications, and I got this:
```
The converter works on the following exercises:
-   all-your-base
-   anagram
-   bob
-   change
-   circular-buffer
-   connect
-   crypto-square
-   diamond
-   flatten-array
-   go-counting
-   grade-school
-   hello-world
-   markdown
-   matrix
-   meetup
-   minesweeper
-   palindrome-products
-   poker
-   proverb
-   pythagorean-triplet
-   raindrops
-   react
-   resistor-color-trio
-   reverse-string
-   rna-transcription
-   roman-numerals
-   saddle-points
-   satellite
-   say
-   series
-   sgf-parsing
-   sieve
-   spiral-matrix
-   sublist
-   tournament
-   transpose
-   two-bucket
-   two-fer
-   word-count
-   word-search
-   zebra-puzzle
-   zipper
Pass: 42, Fail = 73, Neutral = 19
```

soumitradev:
I just made the huge commit I was talking about earlier. The generator now works on 34 exercises instead of the 13 earlier! (these exercises are the ones in the `julia` repo) Only 12 exercises fail on this one. 

Turns out, some exercises have a different JSON structure that needed to be handled. Also added output support for Dicts and Bools.

I think the generator should now print the exercise slugs on which it _doesn't_ work instead of the ones on which it works. Thoughts?

The generator now works on 84/134 of the exercises on `problem-specifications`. This means that the generator works on ~73% of the exercises that have a `canonical-data.json`

As of now, the generator works on the following exercises in `problem-specifications`:
```
The converter works on the following exercises:
-   acronym
-   all-your-base
-   alphametics
-   anagram
-   armstrong-numbers
-   binary
-   binary-search
-   bob
-   book-store
-   bowling
-   change
-   circular-buffer
-   collatz-conjecture
-   connect
-   crypto-square
-   darts
-   diamond
-   diffie-hellman
-   dominoes
-   etl
-   flatten-array
-   food-chain
-   gigasecond
-   go-counting
-   grade-school
-   hamming
-   hello-world
-   house
-   isbn-verifier
-   isogram
-   knapsack
-   largest-series-product
-   leap
-   luhn
-   markdown
-   matching-brackets
-   matrix
-   meetup
-   micro-blog
-   minesweeper
-   nth-prime
-   nucleotide-count
-   ocr-numbers
-   palindrome-products
-   pangram
-   pascals-triangle
-   phone-number
-   poker
-   prime-factors
-   protein-translation
-   proverb
-   pythagorean-triplet
-   raindrops
-   react
-   rectangles
-   resistor-color-duo
-   resistor-color-trio
-   reverse-string
-   rna-transcription
-   roman-numerals
-   rotational-cipher
-   saddle-points
-   satellite
-   say
-   scrabble-score
-   secret-handshake
-   series
-   sgf-parsing
-   sieve
-   space-age
-   spiral-matrix
-   sublist
-   sum-of-multiples
-   tournament
-   transpose
-   trinary
-   two-bucket
-   two-fer
-   word-count
-   word-search
-   wordy
-   yacht
-   zebra-puzzle
-   zipper
Pass: 84, Fail = 31, Neutral = 19
```

soumitradev:
I made some more changes to the generator, and it can now handle nested test cases. I also changed the test script to print only the slugs of the exercises on which the generator fails. This really helps since we can look into their `runtests.jl` files and improve the generator in handling such cases.

Now, in the `julia` implemented exercises, the generator only fails on 4/46 exercises with a `canonical-data.json`, and these need quite a bit of custom test code: 
```
The converter fails on the following exercises:
-   complex-numbers
-   dnd-character
-   grains
-   rational-numbers
Pass: 42, Fail = 4, Neutral = 2
```

Coming to the general `problem-specifications` repo, the generator fails on only 9 of them: 
```
The converter fails on the following exercises:
-   beer-song
-   binary-search-tree
-   complex-numbers
-   dnd-character
-   grains
-   high-scores
-   kindergarten-garden
-   rational-numbers
-   resistor-color
Pass: 106, Fail = 9, Neutral = 19
```
This looks pretty good, we have ~92% passing, but we'll have to look at the output and see if this gives correct output for **all** of them.

I'll wait for a review, and work on this further if needed.
@SaschaMann

SaschaMann:
That sounds awesome! I want to take a look at the output which will take some time, so I won't get to reivew it before the weekend unfortunately.

SaschaMann:
Just did some spot checking:

- `bob`: it creates a test case that fails due to an unescaped `$` and the test suite doesn't match the Julia one (this is expected)
- `acronym`: it uses the wrong function name compared to the Julia implementation

Some general notes:

- It always uses named arguments, but in many if not most cases we want positional arguments.
- it'd be quite useful if we could give it a map of property names to Julia function names (e.g. `abbreviate` -> `acronym`)
- In general it will fail if the Julia test suite has been changed a bit or has additional tests. This is obviously expected.

I'd recommend you generate test suites for all existing exercises and see if the example solutions passes.

If the named args are addressed, I'd be fine with merging the PR as a first version that can be used as starting point for manually implementing an exercise. For automating the creation entirely, I think there are too many special cases that would somehow need to be addressed.

soumitradev:
I generated test suites locally, and I realised I'll have to change the function names (as you pointed out, from `abbreviate` to `acronym`) before doing any testing. How do you think we should give it a map of names? For now, I think automatically picking up the name from the `slug.jl` file would be a good idea.

SaschaMann:
> For now, I think automatically picking up the name from the slug.jl file would be a good idea.

Not all exercises have, or should have, slugs. For existing exercises we could guess it from the tests, I guess. I think there's no way around defining them manually for new exercises. Having a dictionary `Dict{String,Symbol}` that looks like `"abbreviate" => :acronym` would work.

SaschaMann:
Due to the lack of responses to this PR and the upcoming v3 migration, I'm gonna close the PR. We can reuse the code from the branch if we want to implement this at some point but for now it would just break anyway.

PR #220, opened 2020-03-25
Add @testset macro to collatz conjecture exercise

guilhermebodin:


SaschaMann:
Thanks for catching this bug and the PR! :)

The notebooks still had to be generated, so I updated them and merged it manually in https://github.com/exercism/julia/commit/29b4ba884e2ce3373f9927d668ef6beb1b17eaf0

PR #221, opened 2020-04-16
Fix swapped function docstrings

bovine3dom:
Minor typo that confused me for longer than it should have when I did the exercise :)


SaschaMann:
How did this go unnoticed for three years?! :D 

Thanks a lot for reporting and fixing this bug!

I'll update the notebook and merge it in a bit when I'm back at my PC

PR #223, opened 2020-04-17
Add leeway for floating point rounding issues

bovine3dom:
Some perfectly reasonable solutions (i.e. [mine](https://exercism.io/tracks/julia/exercises/complex-numbers/solutions/8c0d5c34022b4670b03811b1d45da70f) :wink:) to this exercise end up with slightly different numbers here.


SaschaMann:
Thanks! That's quite an interesting solution to the exercise :)

PR #226, opened 2020-04-22
Use docstrings instead of type annotations

cmcaine:
Users often overdo it on the type annotations in a way that is
unidiomatic for Julia and contrary to the recommendations of experienced
Julia programmers like Lyndon White and the Julia core team.

To provide some guidance to users, I've added docstrings to all of the
exercise skeletons that I've removed annotations from (the first three
in the track).

### DIFF
--- a/exercises/nucleotide-count/nucleotide-count.jl
+++ b/exercises/nucleotide-count/nucleotide-count.jl
@@ -3,7 +3,7 @@
 
 The frequency of each nucleotide within `strand` as a dictionary.
 
-Invalid strands raise a DomainError.
+Invalid strands raise a `DomainError`.
 
 """
 function count_nucleotides(strand)
--- a/exercises/nucleotide-count/nucleotide-count.jl
+++ b/exercises/nucleotide-count/nucleotide-count.jl
@@ -1,3 +1,11 @@
-function count_nucleotides(strand::AbstractString)
+"""
+    count_nucleotides(strand)
+
+The frequency of each nucleotide within `strand` as a dictionary.
+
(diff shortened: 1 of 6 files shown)


cmcaine:
I think later exercises should have docstrings that say "Your description here" or something.

exercises/nucleotide-count/nucleotide-count.jl
```diff
@@ -1,3 +1,11 @@
-function count_nucleotides(strand::AbstractString)
+"""
+    count_nucleotides(strand)
+
+The frequency of each nucleotide within `strand` as a dictionary.
+
+Invalid strands raise a DomainError.
```
SaschaMann, line 6:
```suggestion
Invalid strands raise a `DomainError`.
```

SaschaMann:
> I think later exercises should have docstrings that say "Your description here" or something.

I think we should add mentoring notes if we do that say that docstrings are (or aren't) optional so that everyone's on the same page. Perhaps even adding an example docstring to them?

cmcaine:
At the moment, we don't have mentoring notes for many exercises, but I was thinking we should add one or two annotated example solutions to each. Those can all have suitable docstrings.

I think we should consistently suggest them and show them in example code, but I don't think I would block a student from progressing if their solution was otherwise good.

SaschaMann:
> but I don't think I would block a student from progressing if their solution was otherwise good.

Yea that sounds good to me

SaschaMann:
(Please don't fix the notebooks, I'm trying to do it via an action and need to test it on something :P)

PR #231, opened 2020-05-08
rot13: Autodetect bonus macros in tests

cmcaine:
This means mentors and students don't need to fiddle with the tests file.


cmcaine:
Looks like the Travis build errored/timed out, but I don't think that's meaningful.

@SaschaMann is it ok to ignore travis in general? Else, is it ok to ignore it this time?

cmcaine:
Can we get rid of the lying code coverage bot, too? Or fix it or whatever?

cmcaine:
@SaschaMann can you advise on whether it's okay to ignore the Travis CI error here?

I'm 90% sure it is okay and I'll just merge this tomorrow if you're too busy :)

PR #232, opened 2020-05-13
Perfect perfect-numbers

cmcaine:


cmcaine:
The tests are failing because I'm relying on a feature that's in Julia 1.4 so that I can sum over an empty collection.

exercises/perfect-numbers/example.jl
```diff
@@ -1,6 +1,10 @@
 function aliquot_sum(n)
     if !isinteger(n) || n < 0
         throw(DomainError(n,"n must be a non-negative integer"))
+    # In Julia <1.4, you need this special case because `sum` doesn't like
```
SaschaMann, line 4: A workaround would be `reduce(+, i for i in 1:div(n, 2) if n % i == 0; init=0)` but it's slightly slower.

cmcaine: I want to keep the workaround separate to emphasise that there is a nice option for modern Julias.

cmcaine:
Might change this to be defining predicates instead.

PR #236, opened 2020-05-25
Fail if  there are 26 letters case sensitive

Wikunia:
I received one solution where it was only checked whether there are exactly 26 letters (case sensitive)
This would result in that `a...m A...M` is a pangram for example.
Also fails if there are more than 26 letters when counting i.e o and O as two different ones.

### DIFF
--- a/exercises/pangram/runtests.jl
+++ b/exercises/pangram/runtests.jl
@@ -38,3 +38,10 @@ end
     @test !ispangram("the quick brown fox jumped over the lazy FOX")
 end
 
+@testset "a-m and A-M are 26 different letters but not a pangram" begin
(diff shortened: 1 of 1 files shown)


Wikunia:
Oh I think I should update it here:
https://github.com/exercism/problem-specifications
Correct?

cmcaine:
Thanks!

cmcaine:
Also merged this: https://github.com/exercism/problem-specifications/pull/1658

Wikunia:
Ah okay in that case it should be added to the notebook as well, right?

cmcaine:
Yup, also done in #237

cmcaine:
There's a script you can run for it:

julia bin/generate-notebooks.jl

If you speak the magic words, a github action will do it for you, too.

PR #239, opened 2020-05-28
robot-name: be better

SaschaMann:
Testsets have the same seed, so `rand()` will always return the same value.
By moving the loop inside the testset, this can be avoided.

@cmcaine I'd appreciate it if you could take a look if this makes sense


cmcaine:
So the motivation is that you want there to be a chance that the tests will fail?

I was wondering if we should just mandate that the students come up with a solution that only fails when it's completely out of names.

SaschaMann:
> So the motivation is that you want there to be a chance that the tests will fail?

Yes, indeed.

> I was wondering if we should just mandate that the students come up with a solution that only fails when it's completely out of names.

That sounds better. Is there a way to accomplish that better than running the test often enough that it will fail once all valid names have been used?

cmcaine:
We don't necessarily need to change the tests, we can just change the readme and tell the mentors what to look for, but, yeah, just asking for 26^2 * 10^3 names will do it. That's only 700k or so iterations, and it will fail before the end.

Downside is that naive solutions will not fail, but will take a long time to find the last names.

Are you aware that with the begin, the testset begins with the same seed so will always check the same 100 names? That's better than checking the same name 100 times, of course, but maybe not what you want?

You can seed the rng with the device random number generator before executing the loop, if you like:

```jl
@testset begin
    Random.seed!(rand(Random.RandomDevice(), UInt))
    for i in 1:10
        @show sum(rand(100))
    end
end
```

cmcaine:
If testing is taking too long, we could just make the robot names shorter.

cmcaine:
Oh, you can just do `Random.seed!()` without an argument and that seems to work fine, too.

cmcaine:
@SaschaMann I revised the PR, changing the readme and fixing the tests. Take a look and see if you like it.

/make-notebooks

cmcaine:
Oh, I changed the interface, too. Now it's `name(robot)` not `robot.name`, which is generally better, I reckon.

exercises/robot-name/README.md
```diff
@@ -2,36 +2,41 @@
...
 
 Every once in a while we need to reset a robot to its factory settings,
-which means that their name gets wiped. The next time you ask, it will
+which means that its name gets wiped. The next time you ask, that robot will
 respond with a new random name.
 
 The names must be random: they should not follow a predictable sequence.
-Random names means a risk of collisions. Your solution must ensure that
+Using random names means a risk of collisions. Your solution must ensure that
 every existing robot has a unique name.
```
SaschaMann, line 16:
```suggestion
Using random names means a risk of collisions. Your solution must ensure that
every existing robot has a unique name.

The test suite only generates ~100 names by default. There are ~700k valid names, so it will only give a small chance of collisions. Consider testing your solution for collisions in some other way in addition to the test suite.
```

exercises/robot-name/example.jl
```diff
@@ -1,24 +1,37 @@
...
 
 function reset!(instance::Robot)
-    instance.name = new_name()
+    instance.id = mint_name()
     instance
 end
+
+function name(instance::Robot)
+    id2name(robot.id)
+end
```
SaschaMann, line 37:
```suggestion
function name(robot::Robot)
    id2name(robot.id)
end
```

exercises/robot-name/README.md
```diff
@@ -2,36 +2,41 @@
...
-and random [numbers](http://docs.julialang.org/en/stable/stdlib/numbers/).
+This exercise continues our exploration of Julia's
+[type system](https://docs.julialang.org/en/v1/manual/types/),
+this time with mutable types,
+and introduces us to
+[random number generation](https://docs.julialang.org/en/v1/stdlib/Random/).
 
-Let's create a structure that will describe our robots
-and then use it to create new instances and to work with them.
+We will imagine that resetting the robot to the factory settings is like a surgery or a really good book: it changes its subject.
```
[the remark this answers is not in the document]
cmcaine: A really good book changes the subject it's written about ;)

But that was a bit silly, I'll write something simpler.

cmcaine:
Changes made as requested. I'll approve it, and then you can merge if you think it's ready ;)

exercises/robot-name/example.jl
```diff
@@ -1,24 +1,37 @@
...
+    id, d2 = divrem(id, 10)
+    d3 = id
+    return join((Char('A' + c1),
+                 Char('A' + c2),
+                 string.((d1, d2, d3))...))
 end
 
 function reset!(instance::Robot)
-    instance.name = new_name()
+    instance.id = mint_name()
```
SaschaMann, line 31:
```suggestion
    instance.id = mint_id()
```

exercises/robot-name/runtests.jl
```diff
@@ -28,41 +36,45 @@ end
...
     @testset "names of robots are valid" begin
-        @test isname(r2.name)
-        @test isname(r3.name)
+        @test isname(name(r2))
+        @test isname(name(r3))
     end
     
     @testset "names of robots are not equal" begin
-        @test r2.name != r3.name
+        @test name(r2) != r3.name
```
SaschaMann, line 44:
```suggestion
        @test name(r2) != name(r3)
```

exercises/robot-name/example.jl
cmcaine: I think I just copied the wrong version of the solution over :(

cmcaine:
Soz, looked like I was testing the wrong version of example.jl somehow!

SaschaMann:
No worries. Is it the correct version now or are there any local changes that you still want to push? Otherwise this is goot to be merged once CI passes.

cmcaine:
It was just those remaining typos, I think. Anything else was lost when I did git checkout -f ;)

The tests pass on example.jl, so that's probably enough.

SaschaMann:
Please squash-merge next time ;)

PR #244, opened 2020-05-30
Make generate-notebooks less fiddly to call

cmcaine:
This will break workflows that rely on generate-notebooks working for
other directories, but I think that's pretty unlikely.

### DIFF
--- a/bin/generate-notebooks.jl
+++ b/bin/generate-notebooks.jl
@@ -1,12 +1,16 @@
+#! julia
+
+const BASEPATH = joinpath(@__DIR__, "..")
+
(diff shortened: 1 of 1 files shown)


SaschaMann:
What does this change in practice? Will I still be able to run `julia bin/generate-notebooks.jl` from the root and have it work? If not, the workflow needs to be updated as well.

cmcaine:
Yes, you can still call it from the root like that.

You can also call it from other directories by its path, so if my working dir is exercises, I can call it as ../bin/generate-notebooks.jl

The #! also means that prefixing it with `julia` is optional.

SaschaMann:
Great!

PR #245, opened 2020-05-31
Regenerate READMEs

cmcaine:
Replaces #164


cmcaine:
Ah, I've found the issue. I can improve this.

cmcaine:
dnd-character says that it's tested on versions of julia v1.1 and up, does it actually not work on Julia v1.0?

cmcaine:
This is ready to go, imo :)

SaschaMann:
> dnd-character says that it's tested on versions of julia v1.1 and up, does it actually not work on Julia v1.0?

It does, or CI wouldn't pass. The exercise was a GCI submission, maybe they hand-wrote the file, and it wasn't noticed during review.

cmcaine:
Cool, that's what I thought. Fixed now.

PR #250, opened 2020-06-23
Escape $ in notebooks

### LINKED ISSUE #249: generate-notebooks doesn't work properly
The notebooks for Bob and Robot Name apparently don't work due to "bad escaped character"s. I think it's because the $ are being escaped. That's a bug in generate-notebooks.jl. Constructing the JSON with a lib will fix that. Other simple fixes are available.

cmcaine:
Fix #249


runtests.jl
```diff
@@ -28,5 +28,10 @@ for exercise in readdir("exercises")
     @info "Testing $exercise"
     Base.include(m, joinpath(exercise_path, "runtests.jl"))
     
+    # Check that all ipynb's are valid JSON
+    # open("$exercise_path/$exercise.ipynb") do f
+    #     string(JSON3.read(f))
+    # end
+
```
SaschaMann, line 35: What do you think about moving this check to the end of the notebook generator script and enabling it?

cmcaine: JSON3 isn't used by any of the other packages and I don't know what packages are available in the images, otherwise, yeah, some kind of validation at time of generation would make sense.

cmcaine:
I think this is good to go unless you want the CI to install external packages.

I don't think validation is very important, and even if we do want it, it can be punted to another issue/PR :)

runtests.jl
SaschaMann: You can load all packages that you want in CI. Only need to add a link that runs `Pkg.add` or environment files and `Pkg.instantiate`

SaschaMann:
I think validation would be nice to have to catch these errors. Thanks for fixing the bug :)

PR #252, opened 2020-06-26
Update fetch-configlet scripts from exercism/configlet

vyu:
It's borked and breaking the ["Progression diff" workflow process](https://github.com/exercism/julia/actions?query=workflow%3A%22Progression+diff%22). An upstream update fixed it: https://github.com/exercism/configlet/commit/c70be6269e71512ef0c041644f2814ad1a0e2981

This just copies over that update.


cmcaine:
Thanks!

PR #253, opened 2020-06-30
rotational-cipher: Ask to not submit bonus tasks

SaschaMann:


exercises/rotational-cipher/.meta/hints.md
```diff
@@ -1,3 +1,10 @@
+## Bonus Tasks
+
+This exercise is quite large and potentially complicated, and that makes it more challenging
+and time-consuming to mentor. To help your mentor out, please don't submit code for the
+bonus exercises until your mentor has reviewed your solution for the first part of the
+exercise.
```
iHiD, line 6:
```suggestion
This exercise is quite large and potentially complicated, and that makes it more challenging
and time-consuming to mentor. If you are in Mentor Mode or are intending to request a mentor in Practice model, to help your mentor out, please don't submit code for the
bonus exercises until your mentor has reviewed your solution for the first part of the
exercise.
```

SaschaMann: Thanks! I'll also change that hint in circular-buffer

iHiD: Sweet :)

SaschaMann:
> I don't think we need to add a warning to the rot cipher exercise. Or at least not this warning.

I saw a lot of submission that contained a solution to the bonus exercise. Since it's quite an advanced topic, I felt it may prevent some mentors from picking the solutions up, even if they could mentor the main part of the exercise well.

You do a lot more mentoring than me, so if you say it's not needed, let's go with that

PR #257, opened 2020-09-24
Remove accumulate exercise

### LINKED ISSUE #256: Accumulate exercise confusing and flawed
The Accumulate exercise says to implement the `accumulate` operation, but the instructions are akin to the `map` operation.

Accumulate: https://docs.julialang.org/en/v1/base/arrays/#Base.accumulate

Map: https://docs.julialang.org/en/v1/base/collections/#Base.map

Additionally, the test cases make no sense.

```
@testset "multiply" begin
    @test accumulate(collect(1:10), *) == [1, 4, 9, 16, 25, 36, 49, 64, 81, 100]
end
```

If it were the `accumulate` operation, the result would be `[1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800]`
And for `map` it would be `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]` but mapping a binary operation to a single argument is... strange.

From what I can tell, the rest of the tests cases are similarly nonsensical. They seem to be either testing the binary operation performed on the element and itself again, or the element and its index in the collection.

cmcaine:
Fix #256.

This is a very confusing exercise and I don't think we can save it.


SaschaMann:
I don't know if removing it entirely will break things for people that have already submitted solutions to it. I think setting it to deprecated is how exercises should be removed: https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md#track-configuration-file

cmcaine:
Hmm. Maybe, but those keys aren't present in our config.json.

@iHiD How do we stop serving an exercise / remove it entirely?

iHiD:
Yeah - it needs to have `deprecated: true`. Removing the exercise from your config basically just breaks the connection between your config and the data in the database, but doesn't remove it from the db.

cmcaine:
Cool. Let's try that.

iHiD:
Let me know if it doesn't work. Note that it will still appear for you if you have started the exercise.

cmcaine:
It's disappeared from the list, so that worked 👍

iHiD:
Brilliant. Thanks @cmcaine and sorry the docs weren't great!

PR #213, opened 2020-01-06
Add high-scores exercise

josobar:
---

Fixes #212

PR #260, opened 2020-09-28
Add high scores exercises, attempt 2

cmcaine:
This is #213, but on a new branch so that we can more easily edit, etc.


cmcaine:
You couldn't implement this without reading the tests, so I think this needs some work on the description. That might then feed into some work on the API of the function.

Here's my attempt:

> Implement a function `scores` with two keyword arguments, `latest` and `top`.
>
> If `latest` is given, return the most recent score (the final one in the array). If `top` is given, return the best scores in descending order. Else return `scores` unmodified.

I think it's strange to have a function that is just `identity` if you don't provide keyword arguments. Another code smell is that the keyword arguments are mutually incompatible.

I think we should drop this exercise and PR or turn it into a "simple custom structs" exercise, which is what it is on the Ruby track. https://github.com/exercism/ruby/blob/master/exercises/high-scores/.meta/solutions/high_scores.rb

That would look something like this:

```julia
struct Scores
    x::Vector{Int}
end

latest(scores::Scores) = last(scores.x)
personal_best(scores::Scores) = maximum(scores.x)
top(scores::Scores, n) = sort(scores.x; rev=true)[1:n]
```

SaschaMann:
Due to the lack of responses to this PR and the upcoming v3 migration, I'm gonna close the PR. We can reuse the code from the branch if we want to implement this at some point but for now it would just break anyway.

PR #261, opened 2020-09-28
reverse-string: encourage grapheme awareness

### LINKED ISSUE #238: Add tests to reverse-string to teach correct string handling
I have brought this up centrally in https://github.com/exercism/problem-specifications/issues/1661 but it will hit each language differently, so it was suggested it might be changed just for Julia.

Since Julia is made to be able to handle strings properly, we shouldn't be encouraging solutions that treat strings as if they were arrays of bytes (as many of the elegant but too simple community solutions do). Incorrect handling of non-ASCII characters (or surrogate pairs if you use something like Java) is one of the biggest sources of bugs in the industry.

I propose adding at least a test with a multibyte UTF-8 character as is common in most European languages, e.g. "skåp" which becomes "påks" (this would make a difference in the Julia solutions, for example)

For even better benefit (for UTF-16 languages), it could be good to have a string containing a surrogate pair, e.g. "\uD834\uDD1E", a G-clef, which should come out as it was.

These two changes are easy to handle in Julia if you just treat strings with the proper respect.

I think that is enough, and it would probably be too difficult to also handle combining marks correctly, e.g. "as⃝df̅" should become "f̅ds⃝a", not "̅fd⃝sa" (from rosettacode.org)

cmcaine:
Closes #238. I've put in a compulsory unicode test with an emoji and
left the grapheme tests optional for now because they're a bit fiddlier
to get right.

### DIFF
--- a/exercises/reverse-string/README.md
+++ b/exercises/reverse-string/README.md
@@ -6,6 +6,19 @@ For example:
 input: "cool"
 output: "looc"
 
+## Bonus
+
+Preserve grapheme clusters, i.e.
+
+```julia
+myreverse("hi 👋🏾") == "👋🏾 ih"
+myreverse("as⃝df̅") == "f̅ds⃝a"
+```
+
+You will probably find the `Unicode` stdlib useful for this bonus task.
+
+To enable the graphemes test, add `const TEST_GRAPHEMES = true` to the global scope of your file.
+
 ## Source
 
 Introductory challenge to reverse an input string [https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb](https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb)
--- a/exercises/reverse-string/example.jl
+++ b/exercises/reverse-string/example.jl
@@ -1,7 +1,7 @@
+using Unicode: graphemes
+
+const TEST_GRAPHEMES = true
+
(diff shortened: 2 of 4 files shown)


exercises/reverse-string/README.md
```diff
@@ -6,6 +6,19 @@ For example:
...
+Preserve grapheme clusters, i.e.
+
+```julia
+myreverse("hi 👋🏾") == "👋🏾 ih"
+myreverse("as⃝df̅") == "f̅ds⃝a"
+```
+
+You will probably find the `Unicode` stdlib useful for this bonus task.
+
+To enable the graphemes test, add `const TEST_GRAPHEMES = true` to the global scope of your file.
```
SaschaMann, line 20: I wish Julia had a nicer way of handling this.

SaschaMann (approved):

exercises/reverse-string/README.md
cmcaine: I don't think there's much that can be done. Test or build suites in other languages would read a DEFINE or config files or environment variables. We want the student to only have to modify the one file and we want mentors to run the same tests the students are running, so we probably don't want a config file or env variables.

We could define a function in runtests and give the students an API like

```julia
enable_bonus_test(:graphemes)
```

or something?

That would require some changes to our repo-level runtests because that includes the files in a different order.

SaschaMann: I think how you did it works fine, it would just be nice if Julia had a nice way to run specific testsets. (there's an issue open for that I think)

PR #262, opened 2020-09-28
Faster tests with a fast pythagorean-triplet example

cmcaine:
This shaves 25% off the test time for me.


SaschaMann:
Neat, thanks!

Issue #258, opened 2020-09-28
Deprecate transpose exercise?


cmcaine:
It's really weird. At a minimum I think we should emphasise the strangeness in the instructions, IMO.

Here's my solution to it, to explain the weirdness:

https://exercism.io/tracks/julia/exercises/transpose/solutions/0882224dd6864e448cc89762665ed7ee

```jl
"""
    transpose_strings(input)

Given an iterable of strings, interpret them as a matrix of characters and
transpose them returning a new iterable of strings.

The tricky business is in padding: we are expected to pad the output on the
left with spaces but strip spaces on the right while also preserving all
spaces in all columns.

So an input of:

["a c",
 "A",
 "12345"]


Yields:

["1Aa",
 "2  ",
 "3 c",
 "4",
 "5"]

"""
function transpose_strings(input)
    isempty(input) && return input
    maxlen = maximum(length, input)
    # Pad each line with null characters to the same length, then rotate
    map(eachrow(hcat(collect.(rpad.(input, maxlen, '\0'))...))) do row
        # Strip null bytes from the right,
        # then replace any remaining nulls with spaces
        replace(rstrip(join(row), '\0'), '\0' => ' ')
    end
end
```

SaschaMann:
I agree it's a bit strange. Otoh while the solution is strange, it requires thinking about various edge cases. There might something else to be learned from this exercise if the padding is cleared up so that it can be solved with `transpose(s::String) = s` and `input'`.

cmcaine:
It's tricky enough that a bunch of the community exercises pass the tests but don't obey this property (the tests happen to only require the last line to be stripped because the first line of the input is only ever at most one character shorter than the longest).

I think practicing how to do weird and silly things is worthwhile, because that is what a lot of programming is, but I think this problem is underspecified in the description and a lot of answers just end up bodging it.

glennj:
I just noticed this got deprecated. The key is to ensure that each output row is _at least as long as the **next** one_.

cmcaine:
Sure, constructing the rows of the transposed matrix in reverse order is a bit simpler, I hadn't thought of that. It doesn't change my point that this is a deeply weird exercise that most people get wrong, though :)

PR #269, opened 2020-10-09
Bringing your track in line with the latest changes to Problem Specifications

ErikSchierboom:
We're in the process of re-opening the [Problem Specifications repo](https://github.com/exercism/problem-specifications), as discussed in [this issue](https://github.com/exercism/problem-specifications/issues/1674).
This PR adds `.meta/tests.toml` files for all exercises for which canonical data is defined in the Problem Specifications repo.
We'll now discuss _why_ we're making this change.
## Keeping track of implemented tests
While most of the changes to the Problem Specifications repo are specific to that repo, there is one track-specific change:
> If a track implements an exercise for which test data exists, the exercise _must_ contain a `.meta/tests.toml` file.
The goal of the `tests.toml` file is to keep track of which tests are implemented by the exercise.
Tests in this file are identified by their UUID and each test has a boolean value that indicates if it is implemented by that exercise.
A `tests.toml` file for a track's `two-fer` exercise looks like this:
```toml
[canonical-tests]
# no name given
"19709124-b82e-4e86-a722-9e5c5ebf3952" = true
# a name given
"3451eebd-123f-4256-b667-7b109affce32" = true
# another name given
"653611c6-be9f-4935-ab42-978e25fe9a10" = false
```
In this case, the track has chosen to implement two of the three available tests.
If a track uses a _test generator_ to generate an exercise's test suite, it _must_ use the contents of the `tests.toml` file to determine which tests to include in the generated test suite.
## Tooling
To make it easy to keep the `tests.toml` up to date, tracks can use the [`canonical_data_syncer` application](https://github.com/exercism/canonical-data-syncer).
This application is a small, standalone binary that will compare the tests specified in the `tests.toml` files against the tests that are defined in the exercise's canonical data.
It then interactively gives the maintainer the option to include or exclude test cases that are currently missing, updating the `tests.toml` file accordingly.
To use the canonical data syncer tool, tracks should copying the [`fetch-canonical_data_syncer`](https://github.com/exercism/canonical-data-syncer/blob/master/scripts/fetch-canonical_data_syncer) and/or [`fetch-canonical_data_syncer.ps1`](https://github.com/exercism/canonical-data-syncer/blob/master/scripts/fetch-canonical_data_syncer.ps1) scripts into their repository.
Then, running either of these scripts will download the latest version of the tool to the track's `bin` directory.
The tool can be run using `./bin/canonical_data_syncer` or `.\bin\canonical_data_syncer.exe`, depending on your operating system.
## Changes
In this PR, we're adding `meta/tests.toml` files for all the exercises for which canonical data is defined.
We've initially marked all tests as _included_, which means that we'll assume that your track has implemented those tests.
If there isn't anything obviously wrong with the PR, I would suggest to merge this PR first, and then later on update the `meta/tests.toml` files according to your track's actual implementation.


SaschaMann:
Checked/updated:

- [ ] accumulate
- [ ] acronym
- [ ] allergies
- [ ] anagram
- [ ] armstrong-numbers
- [ ] atbash-cipher
- [ ] binary-search
- [ ] bob
- [ ] circular-buffer
- [ ] clock
- [ ] collatz-conjecture
- [ ] complex-numbers
- [ ] custom-set
- [ ] darts
- [ ] difference-of-squares
- [ ] dnd-character
- [ ] etl
- [ ] gigasecond
- [ ] grains
- [ ] hamming
- [ ] hello-world
- [ ] isbn-verifier
- [ ] isogram
- [ ] leap
- [ ] luhn
- [ ] matching-brackets
- [ ] minesweeper
- [ ] nucleotide-count
- [ ] pangram
- [ ] pascals-triangle
- [ ] perfect-numbers
- [ ] phone-number
- [ ] prime-factors
- [ ] pythagorean-triplet
- [ ] raindrops
- [ ] rational-numbers
- [ ] resistor-color-trio
- [ ] reverse-string
- [ ] rna-transcription
- [ ] robot-name
- [ ] robot-simulator
- [ ] roman-numerals
- [ ] rotational-cipher
- [ ] run-length-encoding
- [ ] scrabble-score
- [ ] secret-handshake
- [ ] sieve
- [ ] spiral-matrix
- [ ] transpose
- [ ] triangle
- [ ] trinary
- [ ] word-count

SaschaMann:
This was obnoxious to review but it's finally done. All tests are updated to the latest canonical data, all tests.toml files reflect the current state etc.

@cmcaine Some example solutions in mentoring notes may need to be checked against new tests. Some of the example solutions in this repo were wrong, and a few test cases may break assumptions that were valid before. If that's not wanted, we can revert those changes individually in follow-up PRs.

SaschaMann (approved):

cmcaine:
Looks like it's only the luhn and rna-transcription exercises whose examples were wrong and we don't have mentoring notes for either, so we're probably okay? Do you think any specific exercise changed in a way that the mentoring notes are more likely to be invalid?

It would be nice to automatically test the example solutions in the mentoring notes, though. They should probably be in this repo, too. They don't make sense without the context of the exercise and it would be more convenient for us. Maybe that gripe is obsoleted by v3

SaschaMann:
> It would be nice to automatically test the example solutions in the mentoring notes, though. They should probably be in this repo, too. They don't make sense without the context of the exercise and it would be more convenient for us. 

Yeah, I was thinking that they should be tested too.

They're in separate repos because there may be differences between the copy and maintainer teams. I don't think that's particularly relevant for the Julia track but that was the original reason iirc.

Where they are doesn't matter too much for testing but we'd probably have to standardise the format a bit so that we can automatically insert them into the markdown file.

> Maybe that gripe is obsoleted by v3

🤷‍♂️

>  Do you think any specific exercise changed in a way that the mentoring notes are more likely to be invalid?

`luhn` is the main one, also because it contains invalid characters now. `transpose` has an additional test case that may break some solutions, at least by the look of it. Presumably most new cases were added because they cover one previously uncovered edge case, that's why I brought it up.

SaschaMann:
From Slack:

> Will example solutions in mentoring notes still be a thing in v3?

> I've not thought about adding any formal structure to V3 mentoring notes, so for now at least, yes.

So it seems like that gripe will remain and we need to fix it ourselves.

cmcaine:
The new `transpose` test still doesn't hit the edge case I identified in #258. I still think we should remove the `transpose` exercise from the Julia track.

Regarding testing the mentoring notes, I was thinking we might write them in a format like weave.jl and then just generate a plain markdown file and copy that over or template it in?

Also, lol, our example solution for `luhn` is kind of bad. I'll push some changes and mentor notes for that.

SaschaMann:
> Regarding testing the mentoring notes, I was thinking we might write them in a format like weave.jl and then just generate a plain markdown file and copy that over or template it in?

That would work well but it would abolish the copy-maintainer separation entirely. I'm fine with that but @iHiD may not be.

We could also have some kind of `${{ example1 }}` placeholders in the notes and then generate the actual note where those are replaced with the examples.

> our example solution for luhn is kind of bad

A lot of example solutions are. I wrote them while I was fairly new to Julia myself :D

PR #275, opened 2020-10-18
Minor updates to isogram

cmcaine:
`isisogram` -> `is_isogram`, @ScottPJones pointed out that the original
sounds a little like a letter to ISIS.

Clarified that isograms are case insensitive.


SaschaMann (commented):
I don't like the rename. There are many `is...` methods in germaniccase. This would make it inconsistent with that convention.

> @ScottPJones pointed out that the original sounds a little like a letter to ISIS.

The word `isogram` has the same 'issue' give that IS is also commonly used as name for ISIS.

cmcaine:
I don't think anyone would mistake isogram for IS-o-gram because you don't say IS as "is", but at least one person briefly thought `isisogram` was `isis-o-gram`.

I don't care that much, but I don't think this is a spurious suggestion either.

SaschaMann:
> I don't think anyone would mistake isogram for IS-o-gram because you don't say IS as "is"

You don't say isisogram `/ˈaɪsɪs/`ogram either, it's `/ɪzaɪs`ogram, though. Mispronouncing things can sometimes lead to things having different meanings 🤷‍♂️

cmcaine:
So the intended pronunciation is `/ɪzaɪs`ogram, but if I saw the word `isisogram` outside of the context of an exercise on isograms then I'd probably interpret it as `/ˈaɪsɪs/`ogram.

I think this is a harmless change, but I also don't care about leaving it as is. You can decide :)

SaschaMann:
> I think this is a harmless change, but I also don't care about leaving it as is. You can decide :)

I'd rather leave it as it is to stay consistent. If more people report this, we can change it but for now I don't think it's an issue.

The clarification on case-sensitivity should be merged though

SaschaMann (approved):
Thanks! :)

PR #276, opened 2020-10-18
CI: Fail if step fails

SaschaMann:


SaschaMann:
(should be rebased, not squased)

cmcaine (approved):

PR #277, opened 2020-10-18
Add Julia maintainers as default codeowners

SaschaMann:
This will automatically request a review from Julia maintainers.

Later rules in the CODEOWNERS file take precedence, so the maintainers.json and CODEOWNERS rules are unaffected: https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners#example-of-a-codeowners-file

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,6 @@
+# Default fallback
+*                       @exercism/julia
+
 # Maintainers 
 config/maintainers.json @exercism/maintainers-admin
 


cmcaine (approved):

SaschaMann:
(@ErikSchierboom this might a useful change for other tracks too)

iHiD:
Entirely happy for this to be merged.

However, could you explain why it is desirable?  As in, presuming you're subscribed to the repo you get a notification when there's a PR. What extra value does this give you?

iHiD (approved):

SaschaMann:
Makes it easier to filter the notifications. I watch some repos where I almost never want to review a PR.

ErikSchierboom:
Ah, I just saw the notification filtering options:

image

Will this change mean that if someone is in a certain team and they (accidentally) don't watch the repo, they'll still get notified?

SaschaMann:
Yes, they will. There might be a way to disable notifications for team mentions though.

PR #280, opened 2020-10-19
Fix #278

### LINKED ISSUE #278: CI: Add check that no files are missing
https://github.com/exercism/julia/commit/6313e702d82f4fee10efdf29e943df50857cd7b5 accidentally deleted `reverse-string.jl` which caused CI to fail (see #275 and #276) and potentially caused issues to students who couldn't find the file. There should be a check that all required files exist for all exercises:

- `example.jl`
- `$slug.jl`
- `$slug.ipynb`
- `runtests.jl`

cmcaine:


cmcaine:
We could combine some of these checks so that a single GHA has multiple jobs? Seems a bit wasteful to spin up a whole container to do almost nothing with it.

We could also reduce runs by triggering only on `push` and not on `pull_request`? I don't think there's any `pull_request` event that we care about that isn't covered by `push`?

cmcaine:
I think we could combine configlet.yml, notebooks.yml and this new one and that we could remove the pull_request trigger from all of these. What do you think?

.github/workflows/check-files-exist.yml
```diff
@@ -0,0 +1,13 @@
+name: 'Check exercise files exist'
+
+on: [push, pull_request]
+
+jobs:
+  check-files-exist:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
SaschaMann, line 10:
```suggestion
    - uses: actions/checkout@v2
```

v2 is a bit faster

SaschaMann:
> I don't think there's any pull_request event that we care about that isn't covered by push?

PRs from contributors that aren't maintainers aren't covered by push. I'd rather get rid of `push` for anything but the main branch then. Maybe add workflow_dispatch so that one can manually trigger a build on a branch if needed.

> We could combine some of these checks so that a single GHA has multiple jobs? 

Sure, but let's do it in a way that all checks are run even if they fail and at the end the success state is set. Should be doable with outputs or `if`.

cmcaine:
>  I'd rather get rid of push for anything but the main branch then. Maybe add workflow_dispatch so that one can manually trigger a build on a branch if needed.

This seems like it might be worth it?

> Sure, but let's do it in a way that all checks are run even if they fail and at the end the success state is set. Should be doable with outputs or if.

Hmm. That sounds like more work than I cba with right now :)

SaschaMann:
> This seems like it might be worth it?

yea sounds good. then the tests wont run twice at least which takes up the most resources anyway.

> That sounds like more work than I cba with right now :)

could you open an issue for it then? :)

cmcaine:
Done and done.

cmcaine:
I don't know if the workflow_dispatch bit is working. Maybe you could check that?

SaschaMann (approved):
LGTM, thanks!

SaschaMann:
> Maybe you could check that?

It looks correct, can't test it before it's merged to the main branch.

SaschaMann:
Works: https://github.com/exercism/julia/actions?query=workflow%3A%22Run+tests%22

PR #283, opened 2020-10-19
pangram: Add case with letters that aren't part of ASCII

SaschaMann:
This will break solutions relying on `isletter` rather than `isascii`.


exercises/pangram/runtests.jl
```diff
@@ -45,3 +45,7 @@ end
 @testset "pangram with more than 26 letters (if case sensitive)" begin
     @test ispangram("the 1 quick brown fox jumps Over the 2 lazy dogs")
 end
+
+@testset "letters that aren't part of ASCII" begin
+    @test !ispangram("the 1 qüick bröwn föx jümps över the 2 läzy dögs")
```
SaschaMann, line 50: The sentence is obviously nonsense but it demonstrates that a solution using `isletter` would work here because it's a pangram with aou instead of äöü

SaschaMann:
Not sure if this has unintended side effects but given that the input is a string, not some ascii-string type, and we explicitely define the alphabet, a good solution should pass that test regardless.

exercises/pangram/README.md
```diff
@@ -6,7 +6,7 @@ The best known English pangram is:
 > The quick brown fox jumps over the lazy dog.
 
 The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
-insensitive. Input will not contain non-ASCII symbols.
+insensitive.
```
cmcaine, line 9: This change should maybe go in .meta/description.md

cmcaine (approved):

exercises/pangram/README.md
cmcaine: Otherwise configlet generate will overwrite it.

cmcaine (approved):

PR #286, opened 2020-10-28
Replace inappropriate ArgumentErrors with DomainErrors

SaschaMann:


cmcaine:
I don't think the ArgumentErrors are inappropriate, and I kinda think our use of DomainError is dodgy because we often don't actually return the argument that caused the error, we return some property of it.

Historically, I think Julia has both DomainError and ArgumentError because an error without a string was more efficient, so `DomainError(-1)` or whatever was more efficient for using in `sqrt` and friends. Now that that isn't really the case, they're basically the same.

I'm not sure what the usual use is, so I've asked on Zulip: https://julialang.zulipchat.com/#narrow/stream/225542-helpdesk/topic/.60DomainError.60.20or.20.60ArgumentError.60

SaschaMann:
For reference, this is why I think these should be DomainErrors:

> The argument val to a function or constructor is outside the valid domain.

https://docs.julialang.org/en/v1/base/base/#Core.DomainError

> The parameters to a function call do not match a valid signature. Argument msg is a descriptive error string.

https://docs.julialang.org/en/v1/base/base/#Core.ArgumentError

---

In these cases it does match a signature

cmcaine:
Yeah, but what does "valid signature" mean? In most cases an invalid signature gives a `MethodError`.

exercises/hamming/runtests.jl
```diff
@@ -37,6 +37,6 @@ end
 end
 
 @testset "different strand lengths" begin
-    @test_throws ArgumentError distance("AATG", "AAA")
-    @test_throws ArgumentError distance("ATA", "AGTG")
+    @test_throws DomainError distance("AATG", "AAA")
+    @test_throws DomainError distance("ATA", "AGTG")
```
SaschaMann, line 41:
```suggestion
    @test_throws ArgumentError distance("AATG", "AAA")
    @test_throws ArgumentError distance("ATA", "AGTG")
```

exercises/hamming/example.jl
```diff
@@ -1,4 +1,4 @@
 function distance(s1, s2)
-    length(s1) != length(s2) && throw(ArgumentError("Sequences must have the same length"))
+    length(s1) != length(s2) && throw(DomainError((length(s1), length(s2)), "Sequences must have the same length"))
```
SaschaMann, line 2:
```suggestion
    length(s1) != length(s2) && throw(ArgumentError("Sequences must have the same length"))
```

SaschaMann:
I removed it from hamming but I think it still makes sense for the ISBN exercise. What do you think?

cmcaine (approved):
Yeah, I agree with these

cmcaine:
I still don't think this is necessary, but it is not wrong, and I don't mind if these are changed, so your call :)

PR #290, opened 2020-12-06
Add check for denominator not being 0 in testsets

miguelraz:
This blew up in my face.
Here's the fix. 🦾

### DIFF
--- a/exercises/rational-numbers/runtests.jl
+++ b/exercises/rational-numbers/runtests.jl
@@ -103,7 +103,7 @@ end
 @testset "Ordering" begin
     for a in -5:5, b in -5:5, c in -5:5
         a == b == 0 && continue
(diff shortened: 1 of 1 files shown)


miguelraz:
Note that we could a test that checks if `RationalNumber(1, 0) == Inf && RationalNumber(-1, 0) == -Inf` or something like that.

SaschaMann:
> Note that we could a test that checks if `RationalNumber(1, 0) == Inf && RationalNumber(-1, 0) == -Inf` or something like that.

That's not Base's behaviour. `Rational(1, 0)` is a valid number:

```
julia> 1//0
1//0
```

SaschaMann:
To stay consistent with Base, I think we should not skip these tests.

PR #291, opened 2020-12-19
Improve example for luhn exercise

cmcaine:
This is a lot more readable, imo. It's more efficient, too, though we
can fairly easily go 25x faster than this one.


SaschaMann (approved):
Way better indeed. Thanks!

PR #294, opened 2020-12-19
Remove canon. data version comments

SaschaMann:
problem-specs specs are no longer versioned.


exercises/darts/runtests.jl
```diff
@@ -1,6 +1,3 @@
-# canonical data version: 2.2.0
-# This file was generated by a script.
```
SaschaMann, line 2: The script likely doesn't work anymore because it doesn't respect tests.toml, so the comment is irrelevant for now.

cmcaine (approved):

PR #323, opened 2021-01-29
Tooling: Add .meta/config.json generator

SaschaMann:


SaschaMann:
I'm not sure if this is a good idea in the long run, but for now it should make things easier.

PR #324, opened 2021-01-29
WIP: add prerequisites to practice exercises

### LINKED ISSUE #314: Add prerequisites to Practice Exercises
_This issue is part of the migration to v3. You can read full details about the various changes [here](https://github.com/exercism/v3-launch/)._

Exercism v3 introduces a new type of exercise: [Concept Exercises](https://github.com/exercism/v3-docs/blob/main/product/concept-exercises.md). All existing (V2) exercises will become [Practice Exercises](https://github.com/exercism/v3-docs/blob/main/product/practice-exercises.md).

Concept Exercises and Practice Exercises are linked to each other via [Concepts](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/concepts.md). Concepts are taught by Concept Exercises and practiced in Practice Exercises. Each Exercise (Concept or Practice) has prerequisites, which must be met to unlock an Exercise - once all the prerequisite Concepts have been "taught" by a Concept Exercise, the exercise itself becomes unlocked.

For example, in some languages completing the Concept Exercises that teach the "String Interpolation" and "Optional Parameters" concepts might then unlock the `two-fer` Practice Exercise.

Each Practice Exercise has two fields containing concepts: a `practices` field and a `prerequisites` field.

## Practices

The `practices` key should list the slugs of Concepts that this Practice Exercise actively allows a student to practice. 

- These show up in the UI as "Practice this Concept in: TwoFer, Leap, etc"
- Try and choose 3 - 8 Exercises that practice each Concept.
- Try and choose at least two Exercises that allow someone to practice the basics of a Concept.
- Some Concepts are very common (for example `strings`). In those cases we recommend choosing a few good exercises that make people think about those Concepts in interesting ways. For example, exercises that require UTF-8, string concatenation, char enumeration, etc, would all be good examples.
- There should be one or more Concepts to practice per exercise.

## Prerequisites

The `prerequisites` key lists the Concept Exercises that a student must have completed in order to access this Practice Exercise. 

- These show up in the UI as "Learn Strings to unlock TwoFer"
- It should include all Concepts that a student needs to have covered to be able to complete the exercise in at least one idiomatic way. For example, for the TwoFer exercise in Ruby, prerequisites might include `strings`, `optional-params`, `implicit-return`. 
- For Exercises that can be completed using alternative Concepts (e.g. an Exercise solvable by `loops` or `recursion`), the maintainer should choose the one approach that they would like to unlock the Exercise, considering the student's journey through the track. For example, the loops/recursion example, they might think this exercise is a good early practice of `loops` or that they might like to leave it later to teach recursion. They can also make use of an analyzer to prompt the student to try an alternative approach: "Nice work on solving this via loops. You might also like to try solving this using Recursion."
- There should be one or more prerequisites Concepts per exercise.

Although ideally all Concepts should be taught by Concept Exercises, we recognise that it will take time for tracks to achieve that. Any Practice Exercises that have prerequisites which are not taught by Concept Exercises, will become unlocked once the final Concept Exercise has been completed.

### Practices

The `"practices"` field of each element in the `"exercises.practice"` field in the `config.json` file should be updated to contain the practice concepts. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#practice-exercises).

To help with identifying the practice concepts, the `"topics"` field can be used (if it has any contents). Once prerequisites have been defined for a Practice Exercise, the `"topics"` field should be removed.

Each practice concept should have its own entry in the top-level `"concepts"` array. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#concepts).

### Prerequisites

The `"prerequisites"` field of each element in the `"exercises.practice"` field in the `config.json` file should be updated to contain the prerequisite concepts. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#practice-exercises).

To help with identifying the prerequisites, the `"topics"` field can be used (if it has any contents). Once prerequisites have been defined for a Practice Exercise, the `"topics"` field should be removed.

Each prerequisite concept should have its own entry in the top-level `"concepts"` array. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#concepts).

## Example

```json
{
  "exercises": {
    "practice": [
      {
        "uuid": "8ba15933-29a2-49b1-a9ce-70474bad3007",
        "slug": "leap",
        "name": "Leap",
        "practices": ["if-statements", "numbers", "operator-precedence"],
        "prerequisites": ["if-statements", "numbers"],
        "difficulty": 1
      }
    ]
  }
}
```

## Tracking

https://github.com/exercism/v3-launch/issues/6

cmcaine:
Closes #314.

I've done the first few and welcome feedback on whether they're okay.

As I'm sure you know, we don't have enough concepts in config.json to do this at the moment.

Even including all the ones in the concepts dir, we're still missing a concepts for:

- loops
- exceptions
- common data structures (dicts, vectors, sets, tuples) (I wonder if there should be a parent concept of "common data structures" to avoid revealing which ones to use?)
- structs
- bitwise operators

I've just picked likely names and put them in `prerequisites` without adding them elsewhere.

There's also the issue of multiple solutions to problems. Should I add a concept as a prereq if it is commonly used in correct solutions, or only if it is required?

And I'm assuming I should avoid adding concepts that are parents of concepts that are required.


config.json
```diff
@@ -158,7 +158,7 @@
         "slug": "leap",
         "name": "Leap",
         "uuid": "8740c914-f008-4b6b-8e7d-7470e78e6a8e",
-        "prerequisites": [],
+        "prerequisites": ["boolean-logic"],
         "difficulty": 1,
         "topics": [
           "math",
```
SaschaMann, line 164: This is a concept exercise now, needs to be removed in a different PR. Forgot about this when merging #310.

config.json
```diff
@@ -240,7 +255,9 @@
         "slug": "clock",
         "name": "Clock",
         "uuid": "39bc1f1e-68e3-4d35-880b-d6b6fa174605",
-        "prerequisites": [],
+        "prerequisites": [
+          "arithmetic-operators", "structs", "loops"
```
SaschaMann, line 259:
```suggestion
          "arithmetic-operators", "structs", "loops", "multiple-dispatch"
```

config.json
```diff
@@ -252,7 +269,10 @@
         "slug": "robot-name",
         "name": "Robot Name",
         "uuid": "7cb2d96a-4707-4cec-9ead-313fcb5fbb94",
-        "prerequisites": [],
+        "prerequisites": [
+          "arithmetic-operators", "structs", "methods", "randomness",
```
SaschaMann, line 273: What concept does `methods` refer to? The same as `multiple-dispatch` or do you have something else in mind for an exercise explicitely about methods?

config.json
```diff
@@ -182,7 +185,10 @@
         "slug": "pangram",
         "name": "Pangram",
         "uuid": "4bc9a4f8-fc4d-4e6b-a93c-07069bbc5bc9",
-        "prerequisites": [],
+        "prerequisites": [
+          "boolean-logic", "loops",
```
SaschaMann, line 189:
```suggestion
          "boolean-logic", "loops", "sets"
```

This gives away the solution but I guess it may still make sense to have it as pre-req?

SaschaMann (commented):
Thanks!

> I've just picked likely names and put them in prerequisites without adding them elsewhere.

Yep, we can create those later. Knowing which ones we need is quite helpful.

> (I wonder if there should be a parent concept of "common data structures" to avoid revealing which ones to use?)

I think if there's one clearly preferable structure to use, it's fine to add that as a pre-req. But it might be useful to have a less-specific container for more advanced exercises so that students learn when to use which structure. @iHiD do you have thoughts on this?

> There's also the issue of multiple solutions to problems. Should I add a concept as a prereq if it is commonly used in correct solutions, or only if it is required?

This was discussed on the call today but unfortunately I forgot what the conclusion was (there were a few suggestions)... I'll have to check the VoD once it's up :D 

I think any concept that's required to understand the "top" exemplar solution, i.e. the one we currently recommended first in mentoring notes and not one of the performance-optimised solutions, should definitely be a pre-req.

> And I'm assuming I should avoid adding concepts that are parents of concepts that are required.

Yes. I don't think it would break anything but there's no need to have them.

config.json
iHiD: As a reminder, we can't move practice to concepts, so:
- This Practice Exercise needs to be marked as deprecated
- The Concept Exercise needs to have a different uuid, etc.
- The exercises need to have different names and slugs to each other (naming the new one differently would be kinder for students)

cmcaine: Done.

config.json
cmcaine: I was just picking from the list in the concepts directory. multiple-dispatch is fine, I guess I just picked it because the functions here will operate on the student-defined struct, but will probably only have the one method.

The concept matches the OOP understanding of what a method is: you don't need to know anything about multiple-dispatch or open classes to do this exercise.

But multiple-dispatch is better, wouldn't want students to think that OOP-like methods are special.

cmcaine:
```suggestion
          "arithmetic-operators", "structs", "multiple-dispatch", "randomness",
```

config.json
cmcaine: That's the question, I guess. You don't need to use sets to solve this (and many students solve it without sets). Sets are just the most elegant solution.

But this one is perfectly fine, so I don't think I'd want to discourage solutions like this:

```jl
ispangram2(input) = all(in(lowercase(input)), 'a':'z')
```

This is why I think a data-structures or iterables concept might be a good idea?

cmcaine:
I've resolved the merge conflicts.

I don't know how all of this interacts with us not having concept exercises turned on. Maybe we should move all of these prerequisites to `practices` or something instead?

colinleach:
Several years later, this has been overtaken by events.

PR #327, opened 2021-02-09
Add concept exercise: regex / stage-heralding

SaschaMann:


config.json
```diff
@@ -124,6 +124,22 @@
...
+        "slug": "stage-heralding",
+        "concepts": [
+          "regex"
+        ],
+        "prerequisites": [
+          "strings",
+          "conditionals",
+          "nothingness",
+          "symbols",
+          "getindex"
```
SaschaMann, line 139: This is something this exercise could also introduce. The idea of "you can use `foo[bat]` to access more objects than just dicts/vectors" probably doesn't require its own concept exercise.

exercises/concept/stage-heralding/.meta/exemplar.jl
```diff
@@ -0,0 +1,18 @@
...
+    re = r"^»(?P<title>.+)« – (?P<speaker>[\p{L} ]+)(?:\s+\(?(?P<pronouns>[a-z\/]+)?\))?(?:, from (?P<org>[^.]+))?\. Start: (?P<start>[\d:]+)(?:, Q&A: (?P<qanda>[\d:]+))?, End: (?P<end>[\d:]*)$"
+    m = match(re, line)
+
+    """
+    - Our next speaker is $(m[:speaker])$(isnothing(m[:org]) ? "" : ", from $(m[:org])")
+    - $(isnothing(m[:pronouns]) ? first(split(m[:speaker])) * "'s" : titlecase(possessive_pronouns[m[:pronouns]])) talk is called »$(m[:title])«
+    - $(isnothing(m[:qanda]) ? "There will not be a Q&A session." : "$((isnothing(m[:pronouns]) ? first(split(m[:speaker])) : titlecase(first(split(m[:pronouns], "/"))))) will answer your questions in the Q&A session at the end of the talk, starting at $(m[:qanda])")
+
+    $(m[:start]) - $(isnothing(m[:qanda]) ? "NO Q&A" : m[:qanda]) - $(m[:end])
+    """
```
SaschaMann, line 17: I'm so sorry.

SaschaMann (commented):
As usual, still needs an introduction because I'm really bad at writing those

cmcaine (commented):
I think this is mostly okay? The regex is a bit tricky, though. Both of us have used quite complicated features of it.

Here's an exemplar that I think is decent enough to show to students.

Also includes a regex without named groups, which I personally find easier to read, but you might not!

```jl
function generate_card(line)
    re = r"»(.*)« – (.*?)(?: \((.*?)\)|)(?:, from (.*?)|). Start: (.*?), (?:Q&A: (.*?), |)End: (.*)"
    title, name, pronouns, org, start_time, qa_time, end_time = match(re, line).captures
    they = isnothing(pronouns) ? "$name" : titlecase(first(split(pronouns, '/')))
    qa_line = (isnothing(qa_time) ?
               "There will not be a Q&A session" :
               "$they will answer your questions in the Q&A session at the end of the talk, starting at $qa_time")

    """
    - Our next speaker is $(name * (isnothing(org) ? "" : ", from $org"))
    - Their talk is called »$title«
    - $qa_line

    $start_time $(isnothing(qa_time) ? "-" : "- $qa_time -") $end_time
    """
end
```

It's not a very compelling example, but we might want to show them how you can break a regex up with interpolation?

```jl
    rtime = raw"\d\d:\d\d"
    re = Regex("»(.*)« – (.*?)(?: \\((.*?)\\)|)(?:, from (.*?)|). Start: ($rtime), (?:Q&A: ($rtime), |)End: ($rtime)")
```

It's not the best (see automata.jl or https://bakpakin.com/writing/how-janets-peg-works.html (this is a really nice blog post, IMO) for better ways to compose patterns), but maybe it's useful enough to include? idk.

I guess I've never done it "in the wild", and I have used regex a few times, so maybe it's not worth showing.

SaschaMann:
Thanks for the review!

(I'll apply the suggestions tomorrow)

> I think this is mostly okay? The regex is a bit tricky, though. Both of us have used quite complicated features of it.

I've asked in [#maintainers on Slack](https://exercism-team.slack.com/archives/GC3K95MRR/p1612803025043100) and a few people tried it and managed to write one in ~8-10 min on average. The whole exercise will take a bit more than the usual 5-10 min for someone experienced in the language but I think it's within reason. I might add a few more hints that explain the actual regex syntax step by step.

> Oh `informal_name = first(split(name))` will be wrong for lots of cultures.

Perhaps adding the informal name inside the parentheses would work? I would rather not complicate the regex itself even more but doing some post processing seems fine. The capture group would still be the same but it would require some handling afterwards (e.g. checking `m[:pronouns] in keys(possessive_pronouns)` or if it contains a slash or sth).

I've tried to a address a lot of common misconceptions about names and people through the examples but covering all may be impossible in the scope of a concept exercise.

> Also includes a regex without named groups, which I personally find easier to read, but you might not!

It makes the regex itself less scary, so let's use yours :) It has the added advantage that we don't need to use `getindex` as pre-req. 

I feel like this is largely personal preference, so we should probably mention both ways _somewhere_. Maybe the analyzer could eventually detect which approach the student uses and suggest the alternative.

> It's not a very compelling example, but we might want to show them how you can break a regex up with interpolation?

We could mention this in the full-length concept `about.md`. I'm not sure if it really adds anything to this particular example.

SaschaMann:
Note for myself: `string-literals` should either be a pre-req or a `concept`.

cmcaine:
Maybe we could just use the full name? And just use `their` unconditionally somewhere else (as I did)?

SaschaMann:
> Maybe we could just use the full name?

Sure, let's do that.

> And just use `their` unconditionally somewhere else (as I did)?

I know people who don't want `their` to be used either but instead want their (heh) name to be used all the time. That's the sole motivation behind adding that particular edge case.

Obviously these edge cases aren't technically necessary to understand how regex work. But people tend to assume a lot of things about names etc. that simply won't hold true in reality, in particular in regexes or input parsing, so I felt this would be a good way to nudge students towards being more careful. If you feel I went too far at the expense of the exercise, I'm happy to remove that last case and just stick to `their`.

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -0,0 +1,108 @@
...
+    Make sure to set the regex engine/flavour to PCRE.
+
+If you're new to regex or struggle with creating an expression that captures all required information at once, you can also define several smaller expressions that capture parts of the information, e.g.
+
+```julia
+julia> get_org("»Speedrunning 101« – Sasha Duda Krall (they/them), from GDQU. Start: 13:00, Q&A: 13:20, End: 13:30")
+"GDQU"
+```
+
+## 2. Write a function that takes a line and returns the generated moderation card
```
ErikSchierboom, line 101: Maybe just do:

```suggestion
## 2. Generate the moderation card
```

Per https://github.com/exercism/docs/blob/main/anatomy/tracks/concept-exercises.md#file-docshintsmd:

> Task headings should describe the what of the task, not the how.

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -0,0 +1,108 @@
...
+14:50 - 15:05 - 15:10
+```
+
+A few things to note here:
+
+- If no organisation is given for a speaker, leave it out in the moderation card.
+- If no time for the Q&A session is given, the speaker opted against answering questions.
+- If the speaker does not specify pronouns, use the speaker's first name instead.
+
+## 1. Write a regex that captures the information you need to generate the cards
```
ErikSchierboom, line 87: Maybe just do:

```suggestion
## 1. Capture the information needed to generate the cards
```

Per https://github.com/exercism/docs/blob/main/anatomy/tracks/concept-exercises.md#file-docshintsmd:

> Task headings should describe the what of the task, not the how.

ErikSchierboom (approved):
I like the theme. It works well enough. Good idea to have people test the exercise, writing regexes can be hard :)

exercises/concept/stage-heralding/.docs/instructions.md
cmcaine: Probably want to specify that we want a regex, otherwise people will use other tools.

ErikSchierboom: But that would be explained in the task's text, right?

cmcaine:
Oh PCRE has a mode where you can split the match across lines and have comments. Maybe we should show that rather than interpolation.

cmcaine:
I think the named groups version will be a lot more readable with line breaks.

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -0,0 +1,108 @@
...
+13:00 - 13:20 - 13:30
+```
+
+The entire list may look like this:
+
+```text
+»Speedrunning 101« – Sasha Duda Krall (they/them), from GDQU. Start: 13:00, Q&A: 13:20, End: 13:30
+»How I learned to say Farewell« – Madeline (she/her). Start: 13:40, Q&A: 14:00, End: 14:05
+»How To Find a Good Title For Your Conference Talk« – Vítězslav Kruse (he/him). Start: 14:15, End: 14:40
+»Why Emoji Matter❣« – Ash van Amelsvoort, from the University of 🧬🧪⚛. Start: 14:50, Q&A: 15:05, End: 15:10
```
cmcaine, line 52:
```suggestion
»Why Emoji Matter❣« – Ash van Amelsvoort, from the University of 🧬🧪⚛. Start: 14:50, Q&A: 15:05, End: 15:10
»Can dogs look up?« – Kira "k1ralli" Sørensen. Start: 20:50, Q&A: 21:05, End: 21:10
»SQL Injection 201« – Robert'); DROP TABLE Students; --. Start: 20:50, Q&A: 21:05, End: 21:10
```

cmcaine:
Here's what a regex with comments would look like:

```jl
    re = r"""»(.*)«                     # title
             \ –\ (.*?)                 # name
             (?:\ \((.*?)\)|)           # optional pronouns
             (?:,\ from\ (.*?)|).       # optional organisation
             \ Start:\ (.*?),\          # start time
             (?:Q&A:\ (.*?),\ |)        # optional Q&A time
             End:\ (.*)                 # end time
         """x
```

exercises/concept/stage-heralding/.docs/instructions.md
cmcaine: We can have an example with a comma in it, but then the student has to solve it with non-greedy matches (which I think is the right way, but opinions may vary).

SaschaMann: Imo that complicates the regex a bit too much for people that are newish to regex. I'd prefer the structural punctuation to be unique.

SaschaMann:
I think those 3 commits address all the feedback but please let me know if I missed anything.

I'm not a huge fan of what the lines look like with the full name but it's probably better to not guess the first name based on false assumptions.

exercises/concept/stage-heralding/.meta/exemplar.jl
```diff
@@ -0,0 +1,31 @@
...
+               "$they will answer your questions in the Q&A session at the end of the talk, starting at $qa_time")
+
+    """
+    - Our next speaker is $(name * (isnothing(org) ? "" : ", from $org"))
+    - $their talk is called »$title«
+    - $qa_line
+
+    $start_time - $(isnothing(qa_time) ? "NO Q&A -" : "$qa_time -") $end_time
+    """
+end
```
SaschaMann, line 31: This doesn't technically meet the requirements for a small solution space, but I think an aggressive representer could still cut it down sufficiently.

There are so many ways to build and format strings that I don't think this is an avoidable issue for anything that involves non-trivial strings.

cmcaine:
We can have the code pick the first name and have some explanation that for many people, their personal name is their last name, but none of them happen to be speaking at this conference.

Also, I personally think the single line regex is clearer (and is easier to copy into online regex tools), I just added the multiline one as an example of another approach and a technique for splitting up the much bigger named groups regex.

I think we should probably present both in the exemplar (can we have more than one exemplar?)

SaschaMann:
> I think we should probably present both in the exemplar (can we have more than one exemplar?)

I don't think so. I guess we could add a comment with the multiline version?

cmcaine:
Sure. Sounds fine. When we find out what's happening re: showing material to students after they finish an exercise we can maybe revisit?

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -100,7 +100,9 @@ A few things to note here:
 
 - If no organisation is given for a speaker, leave it out in the moderation card.
 - If no time for the Q&A session is given, the speaker opted against answering questions.
-- If the speaker does not specify pronouns, use the speaker's name instead.
+- If the speaker does not specify pronouns, use the speaker's first name instead.
+  Note that in general, it is not safe to assume that the name of a person used in informal contexts is also their first name.
+  However, among the speakers on this stage, this happens to be the case.
```
SaschaMann, line 105: I find it really hard to write this in a way that makes sense since "first name" could refer to both the order and the informally used name. [Wikipedia suggests "given name"](https://en.wikipedia.org/wiki/Given_name) but that doesn't cover all cases either.

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -0,0 +1,128 @@
...
+21:15 - 21:40 - 21:45
+```
+
+A few things to note here:
+
+- If no organisation is given for a speaker, leave it out in the moderation card.
+- If no time for the Q&A session is given, the speaker opted against answering questions.
+- If the speaker does not specify pronouns, use the speaker's first name instead.
+  Note that in general, it is not safe to assume that the name of a person used in informal contexts is also their first name.
+  However, among the speakers on this stage, this happens to be the case.
```
cmcaine, line 105:
```suggestion
- If the speaker does not specify pronouns, use the speaker's first name instead.
  Note that it is not always appropriate to abbreviate someone's name like this, e.g. many Chinese names are better abbreviated by keeping the last name.
  However, the abbreviate-to-first-name rule happens to work fine for everyone speaking on this stage.
```

How about this?

exercises/concept/stage-heralding/runtests.jl
```diff
@@ -0,0 +1,63 @@
...
+
+@testset "SQL Injection 201" begin
+    @test generate_card("»SQL Injection 201« – Robert'); DROP TABLE Students; --. Start: 21:15, Q&A: 21:40, End: 21:45") == """
+        - Our next speaker is Robert'); DROP TABLE Students; --
+        - Robert');'s talk is called »SQL Injection 201«
+        - Robert'); will answer your questions in the Q&A session at the end of the talk, starting at 21:40
+        
+        21:15 - 21:40 - 21:45
+        """
+end
```
cmcaine, line 63: I know I suggested this, but maybe it is too silly and we should drop it?

```suggestion
```

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -0,0 +1,128 @@
...
+20:50 - 21:05 - 21:10 
+```
+
+```text
+- Our next speaker is Robert'); DROP TABLE Students; --
+- Robert');'s talk is called »SQL Injection 201«
+- Robert'); will answer your questions in the Q&A session at the end of the talk, starting at 21:40
+
+21:15 - 21:40 - 21:45
+```
```
cmcaine, line 97:
```suggestion
```

exercises/concept/stage-heralding/.docs/instructions.md
```diff
@@ -0,0 +1,128 @@
...
+
+The entire list may look like this:
+
+```text
+»Speedrunning 101« – Sasha Duda Krall (they/them), from GDQU. Start: 13:00, Q&A: 13:20, End: 13:30
+»How I learned to say Farewell« – Madeline (she/her). Start: 13:40, Q&A: 14:00, End: 14:05
+»How To Find a Good Title For Your Conference Talk« – Vítězslav Kruse (he/him). Start: 14:15, End: 14:40
+»Why Emoji Matter❣« – Ash van Amelsvoort, from the University of 🧬🧪⚛. Start: 14:50, Q&A: 15:05, End: 15:10
+»Can dogs look up?« – Kira "k1ralli" Sørensen. Start: 20:50, Q&A: 21:05, End: 21:10
+»SQL Injection 201« – Robert'); DROP TABLE Students; --. Start: 21:15, Q&A: 21:40, End: 21:45
```
cmcaine, line 54:
```suggestion
```

cmcaine (approved):

exercises/concept/stage-heralding/.docs/instructions.md
SaschaMann: Way better, thanks! :)

exercises/concept/stage-heralding/runtests.jl
SaschaMann: I don't mind either way. Your call if you want to remove it again or not

cmcaine (approved):

PR #332, opened 2021-02-25
Add exercise: elyses-enchantments / vectors1

SaschaMann:
Previous discussion: https://github.com/exercism/v3/pull/3080/


exercises/concept/elyses-enchantments/.docs/hints.md
```diff
@@ -0,0 +1,35 @@
+# Hints
+
+## 1. Retrieve a card from a stack
+
+- `Vector` indices start at `1`.
+
+## 2. Exchange a card in the stack
+
+- `Vector`s are mutable, you can change their contents at any time.
```
SaschaMann, line 9: ### Previous discussion

> Maybe it would be better to do this without mutation first?

_Originally posted by @cmcaine in https://github.com/exercism/v3/pull/3080#r563385512_

> Do you mean we should have a copying method first?
> 
> I can't think of many things to do with arrays other than indexing (and maths) that don't require mutation.

_Originally posted by @SaschaMann in https://github.com/exercism/v3/pull/3080#r563393209_

> I mean we could show how to do all these things with copying (possibly even with tuples instead of arrays?).

_Originally posted by @cmcaine in https://github.com/exercism/v3/pull/3080#r563459072_

SaschaMann: I don't think this exercise works well with tuples, since it relies on mutation quite a bit. Maybe we could have a similar one with tuples?

A lot of these basic methods to manipulate arrays don't even have copying equivalents, e.g. there's no `push` or `deleteat!`. Perhaps we could mention in the text that, if copying is required, one can use `f!(copy(x))` instead?

concepts/pairs/introduction.md
```diff
@@ -0,0 +1,57 @@
+A `Pair` is a data structure that contains exactly two elements accessible through the fields `first` and `second`.
+
+They can be constructed using `x => y`:
```
SaschaMann, line 3:
```suggestion
They can be constructed using `x => y` or `Pair(x, y)`:
```

concepts/pairs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+
+```julia
+julia> p = 3 => true
+3 => true
```
SaschaMann, line 7:
```suggestion
3 => true

julia> p == Pair(3, true)
true
```

concepts/pairs/about.md
```diff
@@ -0,0 +1,60 @@
...
+  3 => -2
+  5 => 10
+  1 => 4
+```
+
+<!-- TODO: Add widget to dictionary exercise -->
+
+## Replacements
+
+Many replacement methods take a `Pair` as argument to make it clear which element is being replaced.
```
cmcaine, line 44:
```suggestion
Many replacement methods take a `Pair` as an argument to make it clear which element is being replaced.
```

concepts/pairs/about.md
```diff
@@ -0,0 +1,60 @@
+A `Pair` is a data structure that contains exactly two elements accessible through the fields `first` and `second`.
```
cmcaine, line 1:
```suggestion
A `Pair` is a data structure that contains exactly two elements accessible through the fields `first` and `second` or the indices `1` and `2`.
```

cmcaine (approved):
There's some duplication between files, which I guess is due to auto-generation or whatever?

Looks good!

exercises/concept/elyses-enchantments/.docs/hints.md
```diff
@@ -0,0 +1,35 @@
+# Hints
+
+## 1. Retrieve a card from a stack
+
+- `Vector` indices start at `1`.
```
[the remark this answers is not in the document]
SaschaMann: Yes, in the not-yet-written intro, but let's have it as a hint anyway.

SaschaMann:
> There's some duplication between files, which I guess is due to auto-generation or whatever?

Most of it is due to concept explanations and exercises not being coupled anymore. The files in `concepts/` will likely grow in the future, while the exercise docs won't.

PR #338, opened 2021-03-05
Improve runtests.jl

SaschaMann:
This cleans up `runtests.jl` a bit so that it runs the same script for practice & concept exercises

### DIFF
--- a/exercises/practice/accumulate/.meta/config.json
+++ b/exercises/practice/accumulate/.meta/config.json
@@ -3,8 +3,8 @@
   "authors": [],
   "files": {
     "solution": [],
-    "test": [],
-    "example": []
+    "test": ["runtests.jl"],
+    "example": [".meta/example.jl"]
   },
   "source": "Conversation with James Edward Gray II",
   "source_url": "https://twitter.com/jeg2"
--- a/runtests.jl
+++ b/runtests.jl
@@ -26,7 +26,7 @@ for exercise_type in ("concept", "practice")
         # runtests.jl includes the solution by calling `include("slug.jl")`
         # Our anonymous module doesn't have `include(s::String)` defined,
         # so we define our own.
-        Core.eval(m, :(include(s) = $(Base.include)($m, $(joinpath(exercise_path, ".meta", example_or_exemplar * ".jl")))
+        Core.eval(m, :(include(s) = $(Base.include)($m, $(joinpath(exercise_path, ".meta", example_or_exemplar * ".jl")))))
         @info "[$(uppercase(exercise_type))] Testing $exercise"
         Base.include(m, joinpath(exercise_path, "runtests.jl"))
         
--- a/runtests.jl
+++ b/runtests.jl
@@ -1,63 +1,37 @@
 using Test
 
-for exercise in readdir(joinpath("exercises", "concept"))
-    # Allow only testing specified exercises
-    if !isempty(ARGS) && !(exercise in ARGS)
-        continue
+for exercise_type in ("concept", "practice")
+    for exercise in readdir(joinpath("exercises", exercise_type))
+        # Allow only testing specified exercises
+        if !isempty(ARGS) && !(exercise in ARGS)
+            continue
+        end
+
+        exercise_path = joinpath("exercises", exercise_type, exercise)
+        isdir(exercise_path) || continue
+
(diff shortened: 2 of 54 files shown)


exercises/practice/accumulate/.meta/config.json
```diff
@@ -3,8 +3,8 @@
   "authors": [],
   "files": {
     "solution": [],
-    "test": [],
-    "example": []
+    "test": ["runtests.jl"],
+    "example": [".meta/example.jl"]
```
cmcaine, line 7: Is this just part of the spec now?

runtests.jl
```diff
@@ -1,63 +1,37 @@
...
+        @eval m $(Symbol("@test_broken")) = $(Symbol("@test"))
+
+        # runtests.jl includes the solution by calling `include("slug.jl")`
+        # Our anonymous module doesn't have `include(s::String)` defined,
+        # so we define our own. We manually include the example solution in our
+        # anonymous module, so we can define `m.include(s::String)` to do nothing.
+        Core.eval(m, :(include(s) = nothing))
+        Base.include(m, joinpath(exercise_path, ".meta", example_or_exemplar * ".jl"))
+        @info "[$(uppercase(exercise_type))] Testing $exercise"
+        Base.include(m, joinpath(exercise_path, "runtests.jl"))
```
cmcaine, line 33: Doing this does make the code evaluate at different times when tested here and when tested by the student, which caused some minor issues when I was adding some optional tests.

This is untested, but would make the `include` happen at the right time if it works:

```suggestion
        # runtests.jl includes the solution by calling `include("slug.jl")`
        # Our anonymous module doesn't have `include(s::String)` defined,
        # so we define our own.
        Core.eval(m, :(include(s) = $(Base.include)($m, $(joinpath(exercise_path, ".meta", example_or_exemplar * ".jl")))
        @info "[$(uppercase(exercise_type))] Testing $exercise"
        Base.include(m, joinpath(exercise_path, "runtests.jl"))
```

cmcaine (approved):

exercises/practice/accumulate/.meta/config.json
SaschaMann: Yes, see https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md#file-metaconfigjson

The reason it's in this PR is that I originally wanted `runtests.jl` to use the file paths from the config, but then afterwards figured instead of introducing a dependency, I might as well hardcode them since they're identical across exercises.

runtests.jl
```diff
@@ -1,63 +1,35 @@
...
+        # The student will not be affected by this.
+        # Overwrite @test_skip and @test_broken with @test
+        @eval m using Test
+        @eval m $(Symbol("@test_skip")) = $(Symbol("@test"))
+        @eval m $(Symbol("@test_broken")) = $(Symbol("@test"))
+
+        # runtests.jl includes the solution by calling `include("slug.jl")`
+        # Our anonymous module doesn't have `include(s::String)` defined,
+        # so we define our own.
+        Core.eval(m, :(include(s) = $(Base.include)($m, $(joinpath(exercise_path, ".meta", example_or_exemplar * ".jl")))
```
SaschaMann, line 29:
```suggestion
        Core.eval(m, :(include(s) = $(Base.include)($m, $(joinpath(exercise_path, ".meta", example_or_exemplar * ".jl")))))
```

runtests.jl
```diff
@@ -26,7 +26,7 @@ for exercise_type in ("concept", "practice")
         # runtests.jl includes the solution by calling `include("slug.jl")`
         # Our anonymous module doesn't have `include(s::String)` defined,
         # so we define our own.
-        Core.eval(m, :(include(s) = $(Base.include)($m, $(joinpath(exercise_path, ".meta", example_or_exemplar * ".jl")))))
+        @eval m include(s) = Base.include($m, joinpath(".meta", $example_or_exemplar * ".jl"))
```
SaschaMann, line 29: Does this still do what you want it to do?

I fixed the path and changed it to use `@eval` like the other defs in this file

cmcaine: Looks fine to me :)

SaschaMann:
Thanks for the review & fix :)

PR #340, opened 2021-03-13
Concepts: Add docs on booleans

SaschaMann:


concepts/booleans/about.md
```diff
@@ -1,3 +1,41 @@
 # About
 
-<!-- TODO: add information on booleans concept -->
+True or false values are represented by the `Bool` type.
+It contains only two values: `true` and `false`.
+
+`Bool` is a subtype of `Integer` and acts as one and zero numerically:
```
SaschaMann, line 6:
```suggestion
`Bool` is a subtype of `Integer` and acts as `1` and `0` numerically:
```

concepts/booleans/about.md
```diff
@@ -1,3 +1,41 @@
...
+
+  julia> true == 1
+  true
+
+  julia> 0 * NaN
+  NaN
+
+  julia> false * NaN
+  0.0
+```
```
cmcaine, line 31: Maybe it would be nice to retain the formatting of the docstring? I don't know.

You can get the markdown of a docstring with: `(@doc Bool) |> clipboard`

````suggestion
```julia
help?> Bool
```

```
Bool <: Integer
```

Boolean type, containing the values `true` and `false`.

`Bool` is a kind of number: `false` is numerically equal to `0` and `true` is numerically equal to `1`. Moreover, `false` acts as a multiplicative "strong zero":

```jldoctest
julia> false == 0
true

julia> true == 1
true

julia> 0 * NaN
NaN

julia> false * NaN
0.0
```
````

cmcaine (approved):

concepts/booleans/about.md
SaschaMann: Sounds good. I looked for a function to get the docstring but didn't find one, so I copied the REPL response instead. Thanks :)

PR #359, opened 2021-04-01
Add missing contributors and authors to concept exercises

SaschaMann:
I've tried to apply the same policy as in #354 where all reviews count as contribution or more.

### DIFF
--- a/exercises/concept/fibonacci-iterator/.meta/config.json
+++ b/exercises/concept/fibonacci-iterator/.meta/config.json
@@ -12,7 +12,11 @@
       "fibonacci.jl"
     ]
   },
-  "contributors": [],
+  "contributors": [
+    "cmcaine",
+    "miguelraz",
+    "iHiD"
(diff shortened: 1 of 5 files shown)


exercises/concept/fibonacci-iterator/.meta/config.json
```diff
@@ -12,7 +12,11 @@
       "fibonacci.jl"
     ]
   },
-  "contributors": [],
+  "contributors": [
+    "cmcaine",
+    "miguelraz",
```
SaschaMann, line 17: This contribution is non-obvious: He solved this on stream which resulted in various improvements to the exercise.

SaschaMann:
@cmcaine Let me know if I still missed adding you as contributor or author somewhere, or if you think you should be author and not contributor for an exercise. I went through all exercises and PRs but when doing that for 60+ exercises it's likely I've missed something.

PR #365, opened 2021-04-04
Leap is dead, long live leap!

SaschaMann:
@iHiD I'd appreciate it if you could check that I did everything correctly


ee7:
Just FYI: `configlet lint` will soon say something like:

> The slug `leap` is used for both a Concept Exercise and a Practice Exercise, but must only appear once on the track:
> ./config.json

until there's only one `leap`. But it looks like this PR is handling it.

(I haven't reviewed this PR).

iHiD:
@ErikSchierboom Could you check this PR locally pls?

ErikSchierboom:
@SaschaMann There is a merge conflict, could you fix that?

SaschaMann:
I'll leave it until we enable concept exercises. Otherwise it deprecates leap for no good reason.

colinleach:
No longer relevant.

PR #371, opened 2021-04-06
Clarify what a docstring is

SaschaMann:
I didn't add it to the `instructions.md` of the concept exercise because it's not necessary to know why docstrings are useful in order to solve the exercise.

### DIFF
--- a/concepts/docstrings/about.md
+++ b/concepts/docstrings/about.md
@@ -2,7 +2,7 @@
 
 A docstring is a piece of text that describes the functionality of the subsequent code.
 Docstrings allow for the automatic generation of documentation.
-They differ from comments in that they're part of the source tree and accessible to the program itself at runtime.[1]
+They differ from comments in that they are integrated with the REPL and documentation help systems. (More details on those later.)
 
 ~~~~exercism/note
 You are not required to write docstrings for your functions while solving exercises on the Julia track if you don't want to.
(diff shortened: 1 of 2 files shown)


concepts/docstrings/about.md
```diff
@@ -1,12 +1,16 @@
 # About
 
+A docstring is a piece of text that describes the functionality of the subsequent code.
+Docstrings allow for the automatic generation of documentation.
+They differ from comments in that they're part of the source tree and accessible to the program itself at runtime.[1]
```
cmcaine, line 5: I think this is confusing in the Julia context. We can also easily look up comments and implementation with `@edit`.

I think we should just say that they differ because they're integrated with the repl and documenter.jl help systems.

cmcaine: Only commenting on the last line.

SaschaMann:
```suggestion
They differ from comments in that they are integrated with the REPL and documentation help systems. (More details on those later.)
```

concepts/docstrings/about.md
```diff
@@ -107,6 +111,9 @@ Base.Docs.apropos
 REPL.stripmd
 ```
 
-## Source
+## Sources
 
 This entire document is derived from the [Documentation chapter](https://docs.julialang.org/en/v1/manual/documentation/) of the Julia Manual.
+
+[1]Docstring, Wikipedia. (2020). https://en.wikipedia.org/w/index.php?title=Docstring&oldid=991140087 (accessed April 6, 2021).
+
```
SaschaMann, line 119:
```suggestion
```

cmcaine (approved):
I think this is good to go if the two proposed changes are merged!

SaschaMann:
Thanks!

PR #373, opened 2021-04-07
Update runtests.jl so to match instruction

Rigante-pl:
Hi,
The project description says:

>Your task is to clean up differently formatted telephone numbers by removing punctuation and the country code (1) if present.
>
>For example, the inputs
>
>+1 (613)-995-0253
>613-995-0253
>1 613 995 0253
>613.995.0253
>should all produce the output
>6139950253

And that's basically it. There is nothing in the project instruction that would suggest that the challenge include changing if the number is valid
(wrong number of digits, '0' or '1' on index 1 or 4). 

Either test or description should be corrected.

### DIFF
--- a/exercises/practice/phone-number/runtests.jl
+++ b/exercises/practice/phone-number/runtests.jl
@@ -8,9 +8,7 @@ if VERSION < v"1.1"
     @eval isnothing(::Nothing) = true
 end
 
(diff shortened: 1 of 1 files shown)


SaschaMann:
See https://github.com/exercism/julia/pull/374#issuecomment-815660868

Rigante-pl:
Thank you, @SaschaMann

PR #374, opened 2021-04-07
phone-number: Update instructions to match tests

Rigante-pl:
The instruction don't contain information about the conditions that "_detect an invalid number_" test is testing. Either this should be included, or the test should be removed (pr #373).
Note: Those first pull request ever... I have no idea what I'm doing. Please assume good will, if I'm doing it wrong.

### DIFF
--- a/exercises/practice/phone-number/.meta/example.jl
+++ b/exercises/practice/phone-number/.meta/example.jl
@@ -11,7 +11,7 @@ function clean(phone_number)
     result = match(r, phone_number)
 
     if result === nothing
-        throw(ArgumentError("\"$phone_number\" is not a valid US phone number."))
+        throw(ArgumentError("\"$phone_number\" is not a valid NANP phone number."))
     else
         return Base.string(result.captures...)
     end
--- a/exercises/practice/phone-number/.meta/example.jl
+++ b/exercises/practice/phone-number/.meta/example.jl
@@ -10,9 +10,9 @@ function clean(phone_number)
 
     result = match(r, phone_number)
 
-    if result != nothing
-        result = Base.string(result.captures...)
+    if result === nothing
+        throw(ArgumentError("\"$phone_number\" is not a valid US phone number."))
+    else
+        return Base.string(result.captures...)
     end
-
-    return result
 end
--- a/exercises/practice/phone-number/.docs/instructions.md
+++ b/exercises/practice/phone-number/.docs/instructions.md
@@ -26,6 +26,11 @@ should all produce the output
 
 `6139950253`
 
-**Note** Valid output falows the _"NXXNXXXXXX"_ format, where `N` is any digit from 2 through 9 and `X` is any digit from 0 through 9. **In other case please return `nothing`**
-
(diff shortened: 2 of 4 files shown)


SaschaMann:
Hi and welcome to Exercism! 👋 

The description should be adjusted, not the tests, so I'm gonna close #373.

I wonder if it may be better if the tests are changed to throw an error when the input is invalid instead of returning `nothing`. What do you think? (cc @cmcaine)

---

Please add yourself as a contributor to https://github.com/Rigante-pl/julia/blob/patch-2/exercises/practice/phone-number/.meta/config.json#L6-L9 so that the change will be properly attributed in the upcoming v3 of Exercism. You should be able to edit that file through the web editor as well. I can do this on merge as well if you want.

Rigante-pl:
Hi, 
Thank you. You're making contributions, so easy to people like me. I appreciate it. 

>I wonder if it may be better if the tests are changed to throw an error when the input is invalid instead of returning nothing. What do you think? (cc @cmcaine)

Hmm, I may be wrong, but `nothing` kind of make sense to me.  You could use the function for looping through the large set of numbers without error that would break out of it.

cmcaine:
Either an error value (nothing) or an exception is okay by me.

If forced to choose, I think exceptions are used more often for signalling invalid input, and the "real" way you'd do this (parse(PhoneNumber, str)) would throw an error, so my vote is for an error.

cmcaine:
I've pushed commits to this branch that adjust the exercise to now require argument errors. Feel free to overwrite them if you strongly disagree.

cmcaine (approved):

Rigante-pl:
Hmm, I wonder If this change the scope of the challenge? It's seems harder then other beginner projects that unluck right after "Hello, World!" is passed.

cmcaine:
Should we unlock this on Hello World in v3? I'd expect it to be gated behind a concept on regex or iteration or similar.

We get users to throw exceptions fairly early in the v2 track (e.g. DomainError in count-nucleotides) and users don't seem to struggle with that, so I'd say it's probably fine?

If there's a dispute we should just stick with `nothing` so that the existing solutions stay more relevant.

SaschaMann (approved):
In v2 it's one of the floating exercises that unlock right away but aren't necessarily suitable towards newbies.

In v3 the unlocks will all change anyway, so it's not an issue

Thanks for the contribution! :)

exercises/practice/phone-number/.meta/example.jl
```diff
@@ -10,9 +10,9 @@ function clean(phone_number)
 
     result = match(r, phone_number)
 
-    if result != nothing
-        result = Base.string(result.captures...)
+    if result === nothing
+        throw(ArgumentError("\"$phone_number\" is not a valid US phone number."))
```
cmcaine, line 14:
```suggestion
        throw(ArgumentError("\"$phone_number\" is not a valid NANP phone number."))
```

cmcaine (approved):

SaschaMann:
Thank you both!

Just a heads-up: This change won't make its way onto the website until v3 launches due to #296.

PR #377, opened 2021-04-16
Add concept exercise: dicts / high-score

SaschaMann:


exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -85,7 +85,11 @@ Base.iterate(S::Squares) = iterate(S, 1)
 
 You may find it useful to combine these two definitions into one using optional arguments:
 
-## optional-arguments
+## Making use of optional arguments
```
SaschaMann, line 88: I feel like it makes sense to quickly go over optional arguments in the introduction for `fibonacci-iterator` again, even if it's technically already introduced in the new `high-score` exercise.

By the time the student gets to this, they may have forgotten and it's commonly used for defining iterators.

exercises/concept/high-score/.docs/instructions.md
```diff
@@ -0,0 +1,127 @@
...
+
+In this exercise, you're implementing a way to keep track of the high scores for the most popular game in your local arcade hall.
+
+<!-- TODO
+    Decide if "edge cases" should be listed in the instructions or not.
+
+    By explicitely mentioning them, students may consider them ahead of time when they don't really need to, because the edge cases match the behaviour of the naive implementation anyway.
+
+    However, if students deviate from the naive implementation, they may find it frustrating that tests fail even if their implementation conforms to the instructions.
+-->
```
SaschaMann, line 11: This refers to edge cases such as "What happens if the score of a player that isn't on the board is reset?".

I'd prefer not mentioning them in the instructions for the reasons in the comment but I don't have a strong opinion on it and am happy to change it if you think it's better to cover every case in the instructions.

exercises/concept/high-score/.meta/design.md
```diff
@@ -0,0 +1,14 @@
...
+## Considered but dropped
+
+Implement a `high_score` function that returns the leader, e.g.
+
+```julia
+function high_score(scores)
+    best = findmax(scores)
+    best[2] => best[1]
+end
+```
```
SaschaMann, line 12: This doesn't actually work because it doesn't handle the edge cases, I merely left it in as an example for future reference.

SaschaMann:
One thing I'm kinda missing is a task about _defining_ a dictionary rather than manipulating it. That said, I assume for testing they'll like do so in the REPL and since there's not much to it, I think it's fine to not have an explicit task about it.

That said, do you have any ideas for a task that requires explicitely defining a Dict?

exercises/concept/high-score/.docs/introduction.md
```diff
@@ -0,0 +1,26 @@
+# Introduction
+
+<!-- TODO: As you may recall from Elyse's enchantments, arrays also use push!() -->
```
cmcaine, line 3: Where will students use `push!` in this exercise?

exercises/concept/high-score/.meta/exemplar.jl
```diff
@@ -0,0 +1,34 @@
+"""
+    add_player!(scores, name, score=0)
+
+Add a player to the score board.
+"""
+add_player!(scores, name, score=0) = push!(scores, name => score)
```
cmcaine, line 6: No action required, but your earlier comment about `push!` confused me because I always do this as `scores[name] = score`. Didn't even know this was possible!

exercises/concept/high-score/.meta/exemplar.jl
```diff
@@ -0,0 +1,34 @@
...
+Update the score of a player on the score board.
+"""
+update_score!(scores, name, score) = scores[name] = score
+
+"""
+    reset_score!(scores, name)
+
+Reset the score of a player on the score board.
+"""
+reset_score!(scores, name) = update_score!(scores, name, 0)
```
cmcaine, line 27: Can do three of these at once with this slightly confusing line ;)

```julia
const add_player! = reset_score! = update_score!(scores, name, score=0) = scores[name] = score
```

exercises/concept/high-score/.docs/instructions.md
```diff
@@ -0,0 +1,119 @@
...
+In this exercise, you're implementing a way to keep track of the high scores for the most popular game in your local arcade hall.
+
+## 1. Add players to the high score board
+
+To add a player to the high score board, define two `add_player!` methods that take the following arguments:
+
+- The first argument is the dictionary of scores.
+- The second argument is the name of a player as a string.
+- The third argument is the score as an integer.
+  The argument is optional, when no value is given, the score should be set to 0.
```
cmcaine, line 12:
```suggestion
  This argument is optional, when no value is given, the score should be set to 0.
```

cmcaine (approved):
Looks good to me! I think it's a smart move to avoid specifying the return types.

exercises/concept/high-score/.meta/exemplar.jl
SaschaMann: That's pretty neat :D

SaschaMann:
Thanks for the review! :)

exercises/concept/high-score/.meta/exemplar.jl
SaschaMann: I'd usually also do it as `scores[name] = score` but I think there's some benefit to introduce people to the idea that `push!` works on a large number of collections, even if in this particular case it may not be the most common use. We still show off `scores[name] = score` in the update method below.

Technically `push!(scores, name => score)` can even _update_ the value of `scores[name]` if `name` is already in `scores` but I find that use of `push!` slightly misleading.

PR #381, opened 2021-04-24
Add concept exercise: dnd-char / kwdef, using-macros 

SaschaMann:


SaschaMann:
The testsuite is still missing a check that `Base.@kwdef` was actually used but since that doesn't really change the exercise as a whole anymore, I'd already like to get some feedback on it and add that check afterwards.

exercises/concept/dnd-char/.docs/instructions.md
```diff
@@ -0,0 +1,105 @@
...
+
+$$
+
+and the hitpoints are defined as
+
+$$
+
+\operatorname{HP}\left(\text{constitution}\right) = 10 + \operatorname{modifier}\left(\text{constitution}\right)
+
+$$
```
SaschaMann, line 36: grafik

concepts/kwdef/about.md
```diff
@@ -0,0 +1,131 @@
...
+only defines the following two
+
+```julia
+julia> methods(HumanStats)
+# 2 methods for type constructor:
+[1] HumanStats(birthyear::Int64, height::Int64, mass::Int64, bmi::Float64) [...]
+[2] HumanStats(birthyear, height, mass, bmi) [...]
+```
+
+and would require manually defining a keyword-based constructor.
```
cmcaine, line 58:
```suggestion
and would require manually defining a keyword-based constructor if you wanted one.
```

concepts/kwdef/about.md
```diff
@@ -0,0 +1,131 @@
...
+    number::Int
+    drafted::Int
+    championships::Vector{Int}
+    points::Int
+    rebounds::Int
+    assists::Int
+end
+```
+-->
+
```
cmcaine, line 108:
```suggestion

<!-- TODO Maybe mention named tuples as an alternative? -->

```

concepts/kwdef/introduction.md
```diff
@@ -0,0 +1,132 @@
...
+## Sources
+
+1. Dirk Nowitzki, Wikipedia. (2021). https://en.wikipedia.org/w/index.php?title=Dirk_Nowitzki&oldid=1019583359 (accessed April 24, 2021).
+
+
+[struct]: https://exercism.io/tracks/julia/concepts/structs
+[discourse-kwdef]: https://discourse.julialang.org/t/what-does-kwdef-do/51973/2
+[public-api]: https://github.com/JuliaLang/julia/pull/35715/files#diff-3591da1fafe13d9fcae96e5b32a9492c6fb68b7e33ac0810b106b09b45d59534
+[add-kwdef-to-manual]: https://github.com/JuliaLang/julia/issues/32659
+[export-kwdef]: https://github.com/JuliaLang/julia/issues/33192
```
cmcaine, line 132: Gripe: I still don't like that `about` and `introduction` are different identical files. Can we not just symlink them if we're going to keep them the same?

exercises/concept/dnd-char/.docs/instructions.md
```diff
@@ -0,0 +1,105 @@
...
+```julia
+julia> modifier(17)
+3
+``` 
+
+## 3. Write a random character generator that follows the rules above
+
+Due to house rules, some characters have pre-determined ability values.
+Those values still follow the same rules, but they're not generated by rolling dice.
+
```
cmcaine, line 77:
```suggestion
```

cmcaine: I think it is less confusing to just go straight to the constructors. Also means we don't confuse anyone unfamiliar with "house rules".

exercises/concept/dnd-char/.docs/introduction.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Sources
+
+1. Dirk Nowitzki, Wikipedia. (2021). https://en.wikipedia.org/w/index.php?title=Dirk_Nowitzki&oldid=1019583359 (accessed April 24, 2021).
+
+[struct]: https://exercism.io/tracks/julia/concepts/structs
+[discourse-kwdef]: https://discourse.julialang.org/t/what-does-kwdef-do/51973/2
+[public-api]: https://github.com/JuliaLang/julia/pull/35715/files#diff-3591da1fafe13d9fcae96e5b32a9492c6fb68b7e33ac0810b106b09b45d59534
+[add-kwdef-to-manual]: https://github.com/JuliaLang/julia/issues/32659
+[export-kwdef]: https://github.com/JuliaLang/julia/issues/33192
```
cmcaine, line 133: I think this should also be a symlink, if possible.

exercises/concept/dnd-char/.meta/exemplar.jl
```diff
@@ -0,0 +1,33 @@
...
+    sort!(dice)
+    sum(dice[2:end])
+end
+
+"""
+    modifier(ability)
+
+Return the modifier based on an ability value.
+"""
+modifier(ability) = floor(Int, (ability - 10) / 2)
```
cmcaine, line 23: Could use `fld` here.

`fld(ability - 10, 2)`

It's a bit faster, but maybe obscure to students.

exercises/concept/dnd-char/runtests.jl
```diff
@@ -0,0 +1,46 @@
...
+include("dnd-character.jl")
+
+@testset "Randomly generated ability is within range" begin
+    for i in 1:1000
+        @test 3 <= ability() <= 18
+    end
+end
+
+@testset "Ability modifiers" begin
+    @test modifier(3)::Int == -4
```
cmcaine, line 12: Do we need to coerce to `Int` here? `-4.0 == -4`.

concepts/kwdef/about.md
```diff
@@ -0,0 +1,131 @@
+# About
+
+The macro `Base.@kwdef` can be used to automatically define constructors that take keyword arguments for the definition of a [struct][struct].
+It also allows one to set default values for fields.
```
cmcaine, line 4:
```suggestion
These keyword argument constructors can also set default values for fields.
```

Otherwise I'd think that the positional constructors have defaults too.

cmcaine (approved):
Looks good to me! Some minor comments.

concepts/kwdef/introduction.md
SaschaMann: > Can we not just symlink them if we're going to keep them the same?

I think in the long run they may diverge because `about.md` may contain more info (e.g. the note on named tuples would only go into `about`, not `introduction`).

But I agree, there will be a lot of shared information and I'd rather avoid the triple duplication.

I've asked about adding symlinks before but was told it's not easy to make symlinks work consistently across several operating systems. I wanted to symlink the `concepts/` documents into the exercise directory that teach those concepts for easier editing.

exercises/concept/dnd-char/runtests.jl
SaschaMann: I wanted to "force" people to use an integer solution because in the context of the exercise, an Int is the natural type of the modifier. Happy to remove it if you disagree with enforcing it, though.

exercises/concept/dnd-char/.meta/exemplar.jl
SaschaMann:
```suggestion
modifier(ability) = fld(ability - 10, 2)
```

> It's a bit faster, but maybe obscure to students.

I think this late into the track progression, we can expect people to look up functions they don't recognise in the REPL. `floor(T, val)` isn't explained either.

exercises/concept/dnd-char/runtests.jl
cmcaine: I think this is the kind of thing that mentoring should pick up?

But I'm also fine with this so long as we specify in the instructions that `modifier` must return an `Int`, and then maybe just have one test case for it?

```julia
@test typeof(modifier(7)) <: Integer
```

SaschaMann: > I think this is the kind of thing that mentoring should pick up?

Concept exercise are mostly not meant to be mentored, so I'd rather have this be part of the tests directly.

> But I'm also fine with this so long as we specify in the instructions that modifier must return an Int, and then maybe just have one test case for it?

I like that idea.

SaschaMann:
Thanks for the suggestions :)

PR #387, opened 2021-05-04
Remove Tasks heading from `instructions.md`

ErikSchierboom:
The `## Tasks` heading should not be in `instructions.md` files.


exercises/practice/circular-buffer/.docs/instructions.append.md
```diff
@@ -1,7 +1,5 @@
 # Instructions append
 
-## Tasks
```
SaschaMann, line 3: This removes it from a practice exercise. Is that intended?

ErikSchierboom: It wasn't. Sorry about that.

ErikSchierboom: Fixed

SaschaMann: No worries, I was just wondering if it might have been a change to the spec too. Thanks for fixing it :)

PR #417, opened 2021-07-03
annalyn2: Add parts of introduction 

SaschaMann:


exercises/concept/annalyns-infiltration2/.docs/introduction.md
```diff
@@ -1,3 +1,81 @@
 # Introduction
 
+This exercise will introduce you to Julia's key paradigm: Multiple Dispatch.
+You will learn how Julia determines which function to call by generating random values.
+
```
cmcaine, line 5: It may be better to explain a little more earlier. I think the manual does this quite well by focusing on the notion of methods rather than on multiple dispatch. This may be a better approach especially for this exercise where we only use single dispatch.

> Recall from Functions that a function is an object that maps a tuple of arguments to a return value, or throws an exception if no appropriate value can be returned. It is common for the same conceptual function or operation to be implemented quite differently for different types of arguments: adding two integers is very different from adding two floating-point numbers, both of which are distinct from adding an integer to a floating-point number. Despite their implementation differences, these operations all fall under the general concept of "addition". Accordingly, in Julia, these behaviors all belong to a single object: the `+` function.

I think that at the top of the intro we should clearly explain the relationship between functions and methods (as is done in the manual).

And I think we may want to say explicitly:

> In this exercise you will call different _methods_ of the `rand` _function_ in order to pick random elements in a few different ways

colinleach:
I'm inclined to leave this old PR open for now. It covers things that the new syllabus still needs to deal with.

PR #429, opened 2021-07-17
Minor update to julia resistor practice instruction

inokawazu:


exercises/practice/resistor-color-trio/.docs/instructions.md
```diff
@@ -39,7 +39,7 @@ So an input of `"orange", "orange", "black"` should return:
 
 > "33 ohms"
 
-When we get more than a thousand ohms, we say "kiloohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams.
+When we get three or more trailing zeros in terms of "ohms", we say "kiloohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams.
```
SaschaMann, line 42: Hi, welcome to Exercism!

I'm not sure this is correct. `1300 ohms` has only two trailing zeros, yet we'd say `1.3 kiloohms`. Could you elaborate why the current text should be changed or clarified?

inokawazu: The main problem are these two tests from runtests.jl.
```julia
@testset "Orange and orange and red" begin
    @test label(["orange", "orange", "red"]) == "3300 ohms"
end
```

```julia
@testset "Red and black and red" begin
    @test label(["red", "black", "red"]) == "2 kiloohms"
end
```

The current tests go off of the rule of "three trailing zeros" then convert to kiloohms.

SaschaMann: Thanks, I didn't know about those tests. I think it might be better to change the tests instead of the description. Counting zeros to determine whether something has a kilo-prefix seems a bit unnatural.

```julia
@testset "Orange and orange and red" begin
    @test label(["orange", "orange", "red"]) == "3.3 kiloohms"
end
```

What do you think?

inokawazu: This makes sense to me.

PR #432, opened 2021-08-07
Add hints to documented-lasagna

angelikatyborska:
It's basically just a repetition of the introduction, but hints are required and what else can you write there? 🤷‍♀️


SaschaMann (approved):
> but hints are required and what else can you write there?

I think a summary of the intro like this works, even if it's technically just another repetition of the same material.

PR #436, opened 2021-08-11
Add debug.md

SaschaMann:
Based on @angelikatyborska's/Elixir's debug.md: https://github.com/exercism/elixir/blob/main/exercises/shared/.docs/debug.md

See also https://github.com/exercism/julia/pull/435

### DIFF
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -6,9 +6,21 @@ You can output any value by passing it to `@show`.
 `@show` shows an expression and its result, while returning its result.
 
 ```julia
-julia> @show uppercase("hello")
-uppercase("hello") = "HELLO"
-"HELLO"
+@show uppercase("hello") # HELLO will be shown as debug output
+```
+
+```julia
+julia> function f(a, b)
+           @show a b
+           return (@show a^b) / 1000
+       end
+f (generic function with 1 method)
+
+julia> f(3, 5)
+a = 3
+b = 5
+a ^ b = 243
+0.243
 ```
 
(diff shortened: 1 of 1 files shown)


exercises/shared/.docs/debug.md
```diff
@@ -0,0 +1,14 @@
...
+To help with debugging, you can use the fact that all console output will be captured and shown in the test results window.
+You can output any value by passing it to `@show`.
+
+`@show` shows an expression and its result, while returning its result.
+
+```julia
+julia> @show uppercase("hello")
+uppercase("hello") = "HELLO"
+"HELLO"
+```
```
SaschaMann, line 12: I wonder if this should be `@show uppercase("hello");` to show what it prints rather than what it returns.

cmcaine: I think a semicolon would be a bit confusing. How about a bigger example?

```julia
julia> function f(a, b)
           @show a b
           return (@show a^b) / 1000
       end
f (generic function with 1 method)

julia> f(3, 5)
a = 3
b = 5
a ^ b = 243
0.243

**Review by cmcaine — APPROVED** [review]

File: exercises/shared/.docs/debug.md (hunk 1 above)
Comment (cmcaine) [reply exercises/shared/.docs/debug.md:12]:
````suggestion
```julia
@show uppercase("hello")
```
````

Comment (SaschaMann) [reply exercises/shared/.docs/debug.md:12]: > How about a bigger example?

Let's have both maybe? The bigger one assumes a bit more knowledge that might be too much earlier on in the track.

Comment (SaschaMann) [reply exercises/shared/.docs/debug.md:12]:
````suggestion
```julia
@show uppercase("hello")
```

```julia
julia> function f(a, b)
           @show a b
           return (@show a^b) / 1000
       end
f (generic function with 1 method)

julia> f(3, 5)
a = 3
b = 5
a ^ b = 243
0.243
```
````

angelikatyborska:
It's time to decide on something and merge, people will need this in 3 days 🙂

PR #448, opened 2021-09-28
Fix runtests for rational-numbers

niczky12:
Only the denominators cannot be 0. The test case of having the numerator 0 is valid. If we only skip the test when `a == b` we get a domainerror as required by the exercises before.


SaschaMann:
Hi!

> Only the denominators cannot be 0. The test case of having the numerator 0 is valid. If we only skip the test when a == b we get a domainerror as required by the exercises before.

The denominator can be 0 as well, see Julia's `Rational` type:

```
julia> 1//0
1//0

julia> 5//0
1//0
```

They can't _both_ be 0 at the same time: 

```
julia> 0//0
ERROR: ArgumentError: invalid rational: zero(Int64)//zero(Int64)
```

Which is what we skip:

```
julia> a, b = 1, 0
(1, 0)

julia> a == b == 0
false

julia> a, b = 0, 0
(0, 0)

julia> a == b == 0
true
```

PR #456, opened 2021-10-09
CI: Run tests using julia-test-runner

SaschaMann:


SaschaMann:
I am not entirely sure why `isbn-verifier` passes tests in our existing CI setup but not in the test-runner setup, but fixing that will be done in a follow-up PR once this is merged

PR #471, opened 2021-11-11
[WIP] Add practice exercise: alphametics

SaschaMann:
I'm struggling to come up with a good enough solution that passes tests in time without relying on packages. If you want to give it a go, help would be appreciated :)


exercises/practice/alphametics/runtests.jl
```diff
@@ -0,0 +1,102 @@
+using Test
+
+include("alphametics.jl")
+
+@testset "puzzle with three letters" begin
+    @test solve("I + BB == ILL") == Dict(
+        'I' => 1,
+        'B' => 9,
+        'L' => 0
```
SaschaMann, line 9: Might be better to change these to `Symbol`s but I'm not sure

cmcaine: Chars are better here, imo

cmcaine:
Here's a brute-force solution that's probably fast enough to pass the CI.

https://gist.github.com/cmcaine/2a056d6b8926e0ae033fe041c5d5aec7

I should probably have timed myself, but to give an idea of the difficulty, I read the prompt on Thursday evening (2 days ago) and thought about the problem for probably 1-3 hours while walking the dog, etc. Then I wrote this solution over about 2-2.5h today.

The solution uses some fairly advanced Julia knowledge: metaprogramming (but you could do string manipulation and Meta.parse to simplify it a bit) and `Base.invokelatest` with a function barrier.

cmcaine:
Oh, I used `Combinatorics.jl` for `permutations`, but that bit could be replaced fairly straightforwardly.

cmcaine:
The "Efficient Brute Force" approach described in the British informatics olympiad solutions would probably perform well in Julia, too.

That doesn't require any metaprogramming. Instead you represent the problem as one or two matrices. SEND MORE MONEY would look like:

```
#    S     E     N     D     M     O     R     Y
A = [1000  100   10    1     0     0     0     0; # SEND
     0     1     0     0     1000  100   10    0] # MORE

B = [0     10    100   0     10000 1000  0     1] # MONEY

# Solution
X = [9 5 6 7 1 0 8 2]

is_valid(A, B, X) == sum(A .* X) == sum(B .* X)
```

SaschaMann:
> Oh, I used Combinatorics.jl for permutations, but that bit could be replaced fairly straightforwardly.

That's not an issue, the definition is short enough we can just copy it and add the copyright notice in the file itself.

PR #472, opened 2021-11-14
Prevent ArgumentError in Ordering testset

gmlewis:
When running the "Ordering" testset, both `b` and `d` can become zero (causing an ArgumentError) when `a` or `c` are not zero.
This PR fixes the logic for this test.


SaschaMann:
Hi!

The implementation is supposed to mirror Julia's default implementation of rational numbers. There should only be an argument error when both the numerator and the denominator are zero. If only the denominator is zero, it should behave similarly to an infinity value for the purposes of ordering.

See https://github.com/exercism/problem-specifications/pull/1856 (which I haven't yet merged downstream)

PR #489, opened 2022-03-13
Update instructions to reflect tests and mentor notes

SirZayers:
Contrary to what the current instructions say the tests do input non-ASCII characters, specifically test 14. 
In my personal opinion this only complicates the exercise without providing any sort of benefit to beginners, which are those most likely to attempt such an easy exercise.
Instead of removing the offending test I have decided to change the instructions instead, as I am unsure how to change the mentor notes, and the code solutions they contain.

### DIFF
--- a/exercises/practice/pangram/.docs/instructions.md
+++ b/exercises/practice/pangram/.docs/instructions.md
@@ -5,5 +5,5 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 The best known English pangram is:
 > The quick brown fox jumps over the lazy dog.
 
-The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
-insensitive. Input will not contain non-ASCII symbols.
+The alphabet used consists of UTF-8 letters `a` to `z`, inclusive, and is case
+insensitive. Input can contain non-ASCII symbols.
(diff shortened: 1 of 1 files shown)


exercises/practice/pangram/.docs/instructions.md
```diff
@@ -5,5 +5,5 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 The best known English pangram is:
 > The quick brown fox jumps over the lazy dog.
 
-The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
-insensitive. Input will not contain non-ASCII symbols.
+The alphabet used consists of UTF-8 letters `a` to `z`, inclusive, and is case
+insensitive. Input can contain non-ASCII symbols.
```
SaschaMann, line 9:
```suggestion
The alphabet used consists of letters `a` to `z`, inclusive, and is case
insensitive.
```

SaschaMann (approved):
Hi, thanks for the PR and pointing out that the description is wrong!

>  Instead of removing the offending test I have decided to change the instructions instead, as I am unsure how to change the mentor notes, and the code solutions they contain.

The mentoring notes can be changed here: https://github.com/exercism/website-copy/blob/main/tracks/julia/exercises/pangram/mentoring.md

>  In my personal opinion this only complicates the exercise without providing any sort of benefit to beginners, which are those most likely to attempt such an easy exercise.

There are endless discussions of this topic over in [prob-specs](https://github.com/exercism/problem-specifications/) where the shared data for practice exercises lives. I think it's a terrible habit to teach people that they can safely assume that unicode strings are limited to ascii characters. In this case I'm not sure how it would complicate the exercise since the vast majority of solutions I've seen don't even need to consider the encoding.

PR #493, opened 2022-04-14
Update some exercises to latest spec

SaschaMann:
More will come but for now I'm done syncing, so this is ready for review & merging


cmcaine (approved):
There's nothing obviously fucked, lmk if you want a more careful review :)

SaschaMann:
> There's nothing obviously fucked, lmk if you want a more careful review :)

That's good enough for me :D 

I'm not too concerned with reviews when tests are only updated to match the upstream implementation unless the exercise is heavily modified or the example solution breaks.

Issue #503, opened 2022-05-05
ISBN Verifier


gpucce:
Hello,

In the ISBN verifier exercise, in the way the tests are designed I think there is no way of getting around having two occurrences of `replace(s, "-"=>"")` or equivalent without a `try ... catch` statement. 

If this is the case and I am not missing something, maybe it is worth writing an hint in the description or changing the tests so it is possible?

I can make the PR if it is something that would make sense.

cmcaine:
Hello!

Why do you particularly want to avoid calling replace in two places?

If you do want to call it just the once you can do so without a try/catch. I think what's tripping you up is that you can't reuse ISBN(s) in isvalid(ISBN, s) without a try/catch and yet you don't want to duplicate the code in the ISBN function. But what if you had a third function?

Does that help?

Regarding a hint in the description, I don't think we are pushing students to use replace only once or avoid try/catch, so I'm not sure we need one?

Anyone wanting to improve their solution can always request mentoring :)

Looking at the description, I think our main issue is that we don't specify the problem very completely (you have to read the tests for that).

We also suggest a truly horrendous bonus task: "Implement show(io::IO, isbn::ISBN) to print the ISBN in the standard format with dashes." Where to place the dashes depends on pages and pages of data about different worldwide publishers, etc cos each part of the ISBN is variable length.

We should probably remove that, tho I think that bit is only visible in the instructions.md that doesn't get used? I forget how that works.

gpucce:
Hi thanks for the answer! I had not thought about that indeed, I really wanted to use those two together :) 

However, just so I don't miss the point again with a third function I would need to use that function twice or can I avoid that?

The strange bonus track I have not noticed honestly so it should be hidden. 

For sure I had to read the tests while doing the exercises otherwise the `DomainError` would have never crossed my mind.

cmcaine:
ISBN(s) and isvalid(ISBN, s) will each need to call your third function at least once each. Further details (not too spoilery) below.

<details>

The removal of the try/catch is simply by replacing a function that will *throw* an error with one that *returns* an error type.

I would implement this by defining `Base.tryparse(::Type{ISBN}, s)`, but of course you don't need to add a method to that existing function you could write something like `tryparse_isbn(s)` or `isbn_or_nothing(s)` or whatever you'd like to call it.

</details>

gpucce:
Thanks so much! This is the solution I didn't know existed :)

SaschaMann:
```
[...]

PR #504, opened 2022-05-06
ISBN biznis

cmcaine:
As I wrote in https://github.com/exercism/julia/issues/503

> Thinking about it some more, I think that's because `isvalid` isn't really the right function for us. `isvalid` checks if a string or character contains only valid UTF-8 codepoints. I think it only makes sense for types where the representation of the type (the bits in memory) can encode something invalid in normal use. This is the case for strings because Julia does not validate that strings or characters are correctly encoded when they are created, it just treats them as opaque bytes.

> For most Julia types, however, you cannot create an invalid value of the type, except through silly use of pointers.

> I think in real code the advice would be to parse the string into an ISBN type, and if that is not possible (e.g. because the string does not encode a valid ISBN), the ISBN constructor (or `parse(ISBN, s)`, whichever makes more sense) should throw an error. If a non-throwing option is required, `tryparse(ISBN, s)` is the right function to define.

> Personally, I would define only a constructor and have it throw errors.

I haven't updated the description yet. If this looks okay to you Sascha then I'll do that, too.

The `Base.show` part in the example isn't tested, I just did that for fun.

`value(x::ISBN)` is inspired by `Dates.value` in the standard library.


exercises/practice/isbn-verifier/.meta/example.jl
```diff
@@ -1,28 +1,39 @@
...
-function verify(s::AbstractString)
-    s = replace(s, "-" => "")
-    chars = split(s, "")
-
-    length(chars) == 10 || return false
-
-    if last(chars) == "X"
-        chars[end] = "10"
+struct ISBN
+    s::String
```
cmcaine, line 2: I'm using a `String` here rather than an `AbstractString` or a generic `T` because I want `ISBN("3-598-21508-8") == ISBN(SubString("3-598-21508-8"))`. I'm not sure if that's too subtle a point to include in the exercise, though.

Maybe a good thing for the mentoring notes.

cmcaine: Might have been better to keep ISBN generic and define `==(a::ISBN, b::ISBN) = value(a) == value(b)`, tho.

cmcaine:
I'm still not sure if `ISBN(s)` or `parse(ISBN, s)` is the better choice here.

Some types, like `Base.UUID`, do both and define `UUID(s::AbstractString) = parse(UUID, s)`, which is probably what would make most sense here, but I'm not sure if it's a good idea to overcomplicate it, y'know?

SaschaMann:
> Some types, like Base.UUID, do both and define UUID(s::AbstractString) = parse(UUID, s), which is probably what would make most sense here (especially for implementations which store the ISBN internally as something other than a string), but I'm not sure if it's a good idea to overcomplicate it, y'know?

Maybe that's something for the mentoring notes as well?

cmcaine:
I'll merge this tomorrow or Wednesday unless you have any objections.

exercises/practice/isbn-verifier/.docs/instructions.md
```diff
@@ -37,6 +43,12 @@ Now, it's even trickier since the check digit of an ISBN-10 may be 'X' (represen
 
 ## Bonus tasks
 
-* Generate a valid ISBN-13 from the input ISBN-10 (and maybe verify it again with a derived verifier).
+If you attempt these tasks, write your own tests for them and include those in your solution too!
```
SaschaMann, line 46: I'm not sure this will work with the web editor. Perhaps we can specify they should be in a separate file or in the main solution?

SaschaMann (approved):

exercises/practice/isbn-verifier/.docs/instructions.md
cmcaine: Works in the web editor for me. Try adding `@test false` at the end of an exercise and then running the tests in the web editor.

cmcaine: We include the student's solution file within a custom testset so it will all work. It'll even work if the student adds `using Test` because they think they need it or a linter tells them to.

SaschaMann: > Works in the web editor for me. Try adding @test false at the end of an exercise and then running the tests in the web editor.

Ah yea, that works. I was thinking about adding it to the test file itself.

cmcaine:
```suggestion
If you attempt these tasks please write your own tests for them in your solution file!
```

Clearer?

SaschaMann: Yes, definitely!

cmcaine: Thanks :)

PR #509, opened 2022-05-14
Remove unused .meta/description.md's

cmcaine:
I compared each with our current instructions.md's and kept the bits that I thought were important. I think the only disputable changes are that I haven't reinstated the comment about broadcasting on accumulate.


SaschaMann (approved):
Accumulate is deprecated so that's whatever.

Thanks for fixing these!

PR #513, opened 2022-08-06
Add extra contraction test for word-count.jl

icweaver:
I noticed this edge case (contractions with more than one letter after the apostrophe) while mentoring. Happy to use different/more examples or incorporate them into the original test if that is preferred

### DIFF
--- a/exercises/practice/word-count/runtests.jl
+++ b/exercises/practice/word-count/runtests.jl
@@ -40,6 +40,7 @@ end
 
 @testset "with apostrophes" begin
     @test wordcount("First: don't laugh. Then: don't cry.") == Dict("first" => 1, "don't" => 2, "laugh" => 1, "then" => 1, "cry" => 1)
+    @test wordcount("Should've could've would've") == Dict("should've" => 1, "could've" => 1, "would've" => 1)
 end
 
 @testset "with quotations" begin


cmcaine (approved):
Good stuff, thanks! We sometimes leave intentional gaps in the automated tests, but I think this is fine.

cmcaine:
@iHiD, we don't require configlet to pass for PRs on this track. The rationale is that the messages about missing files, etc, is useful to anyone wanting to find work to do, but that work is hidden if we e.g. generate a bunch of empty files. Can we have that requirement turned off?

If we ever fix all of the configlet issues we can turn it back on again.

iHiD:
@cmcaine We can't really disable configlet as doing so would risk breaking the track at the website end. 

But it seems that rename the `concepts` folder to `concepts.wip` (or something) and removing the `concepts` block from the config.json (maybe moved into `concepts.wip/config.json` or similar) might fix the errors?

iHiD:
@cmcaine I'm happy to force-merge PRs like this if you tell me you want me to btw.

cmcaine:
A WIP directory sounds like a good idea, but for now could you force-merge this one? Thanks!

iHiD:
Done! :)

iHiD:
I'm off next week btw - but please let me know if you want anything else doing this week or the week after from me :)

PR #522, opened 2022-11-11
Dig deeper

cmcaine:
Copy (and lightly edit) our mentoring notes from exercism/website-copy so that they're visible in the new "Dig Deeper" tab (also known as "Approaches").

This PR copies 13/18 mentoring notes adapted to the new format. More PRs may follow for the other mentoring notes.

Configlet will cry about this PR, but this should be force merged anyway.

### DIFF
--- a/exercises/concept/leap/.approaches/introduction.md
+++ b/exercises/concept/leap/.approaches/introduction.md
@@ -1,0 +1,45 @@
+# Introduction
+
+## General remarks
+
+- Use the remainder operator `%` to find if a number is evenly divisible by another
+- Prefer `&&` and `||` over `&` and `|` to benefit from [short-circuiting](https://docs.julialang.org/en/v1/manual/control-flow/#Short-Circuit-Evaluation). Julia will not evaluate the whole expression but only the expressions that are necessary to determine the result of the entire chain
+- You might be interested to see the Julia standard library's solution: `using Dates; @edit isleapyear(4)`
+
+## Approach: Short-circuiting
+
+```julia
+function is_leap_year(year)
+    year % 400 == 0 || year % 4 == 0 && year % 100 != 0
+end
+```
+
+If you're struggling to follow how this works, or trying to solve a similar problem, consider writing a truth table and seeing if you can work out what the simplest predicates are and how they can be combined with `&&` and `||` to give the full answer.
+
+## Mentoring notes
+
+- Whenever you have an expression like `if condition; return true; else; return false`, you can rewrite it as `return condition`.
+- Consider writing a truth table for this exercise and see if you can work out what the simplest predicates are and how they could be combined with `&&` and `||`
+
+<details>
+<summary>Example truth table</summary>
+
+```
+Let:
+
+p = year % 4 == 0
+q = year % 100 == 0
+r = year % 400 == 0
+
+o = is the desired output
+
+p q r | o
+0 0 0 | 0
+1 0 0 | 1
+...
+1 1 1 | 1
+
+```
+
+The idea is that you 1) work out what predicates you need and 2) notice that you can construct the output with p && !q || r.
+</details>
--- a/exercises/concept/leap/.approaches/introduction.md
+++ b/exercises/concept/leap/.approaches/introduction.md
@@ -3,6 +3,7 @@
 ## General remarks
 
 - Use the remainder operator `%` to find if a number is evenly divisible by another
+- Whenever you have an expression like `if condition; return true; else; return false`, you can rewrite it as `return condition`
 - Prefer `&&` and `||` over `&` and `|` to benefit from [short-circuiting](https://docs.julialang.org/en/v1/manual/control-flow/#Short-Circuit-Evaluation). Julia will not evaluate the whole expression but only the expressions that are necessary to determine the result of the entire chain
 - You might be interested to see the Julia standard library's solution: `using Dates; @edit isleapyear(4)`
 
@@ -15,11 +16,6 @@ end
 ```
 
 If you're struggling to follow how this works, or trying to solve a similar problem, consider writing a truth table and seeing if you can work out what the simplest predicates are and how they can be combined with `&&` and `||` to give the full answer.
-
-## Mentoring notes
-
-- Whenever you have an expression like `if condition; return true; else; return false`, you can rewrite it as `return condition`.
-- Consider writing a truth table for this exercise and see if you can work out what the simplest predicates are and how they could be combined with `&&` and `||`
 
 <details>
 <summary>Example truth table</summary>
--- a/exercises/practice/rotational-cipher/.approaches/introduction.md
+++ b/exercises/practice/rotational-cipher/.approaches/introduction.md
@@ -1,6 +1,6 @@
 # Introduction
 
-## Guidance
+## General guidance
 
 - The key lesson from this exercise is that you can define two methods for the function `rotate`
 - The macro/metaprogramming is quite challenging for lots of students. Feel free to read the answer below!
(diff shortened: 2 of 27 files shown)


ErikSchierboom:
> Configlet will cry about this PR, but this should be force merged anyway.

Great work, but why should this be force merged and the configlet issues not be addressed?

exercises/concept/leap/.approaches/introduction.md
```diff
@@ -0,0 +1,45 @@
...
+
+```julia
+function is_leap_year(year)
+    year % 400 == 0 || year % 4 == 0 && year % 100 != 0
+end
+```
+
+If you're struggling to follow how this works, or trying to solve a similar problem, consider writing a truth table and seeing if you can work out what the simplest predicates are and how they can be combined with `&&` and `||` to give the full answer.
+
+## Mentoring notes
```
ErikSchierboom, line 19: The mentoring notes are a separate thing and should remain in the website-copy repo. The dig deeper parts are there for students, not mentors.

Could you remove this section (and also in other files, if there are mentoring notes there).

cmcaine:
Here's what I said previously: https://github.com/exercism/julia/pull/513#issuecomment-1236393650

> iHiD, we don't require configlet to pass for PRs on this track. The rationale is that the messages about missing files, etc, is useful to anyone wanting to find work to do, but that work is hidden if we e.g. generate a bunch of empty files. Can we have that requirement turned off?
>
> If we ever fix all of the configlet issues we can turn it back on again.

I have just pushed a commit that adds config.jsons to the .approaches directories, tho, which I didn't realise I needed to do.

I've also opened a new PR #523 which can be merged after this one to silence the configlet warnings.

ErikSchierboom (approved):
This is fantastic content! I've left some comments, most of which also apply to the other files.

I'm happy for this to be merged and to tweaks things in follow-up PRs.
As a short summary, what I would propose follow-up PRs would do is:

- Create separate approach documents for each approach listed in the introductions. The introductions themselves would show the approach code (possible slightly trimmed) along with a short (one or two sentence) description of the approach and a link to the (detailed) approach document.
- Create articles for any content that isn't an approach but is still quite "meaty" (as in: it has quite some lines of text). Good examples are "technique" sections.
- Convert the headings to be more consistent

Let me know if any of my questions

exercises/practice/rotational-cipher/.approaches/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+for n in 0:26
+    @eval macro $(Symbol(:R, n, :_str))(s)
+        rotate($n, s)
+    end
+end
+```
+
+## Performance hints
+
+Iterating a UTF-8 string is slow because it's a variable length encoding. Iterating something with elements of a fixed size (a vector of UInt8s or an ASCIIStr (Strs.jl), perhaps), this can be faster.
```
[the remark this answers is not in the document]
cmcaine: Thanks, now using one :)

cmcaine:
Thanks for the review! I've made a few changes and I think my capacity for this PR has been reached.

I appreciate that there are some inconsistencies with how information is presented and how approaches are named both within this PR and when contrasting this PR with already merged dig-deeper pages in other tracks (which are also inconsistent with each other). I'd rather this PR be merged as is because making things consistent would be a fair amount of work with little obvious payoff (unless analytics or user studies show that students benefit from one layout strategy versus another?).

ErikSchierboom (approved):
> Thanks for the review! I've made a few changes and I think my capacity for this PR has been reached.

Makes sense. Smaller follow-up PRs are a better option than iterating on this PR.

> I appreciate that there are some inconsistencies with how information is presented and how approaches are named both within this PR and when contrasting this PR with already merged dig-deeper pages in other tracks (which are also inconsistent with each other). I'd rather this PR be merged as is because making things consistent would be a fair amount of work with little obvious payoff (unless analytics or user studies show that students benefit from one layout strategy versus another?).

Valid points. I would like to see all the approach heading have an "Approach: " prefix, but it's fine to do that in a future update.

Are you interested in also extracting the approaches into separate approach documents?
If you need help scripting that, I would be happy to help.

iHiD:
> It feels a bit disrespectful to say after the staff have put the work in to implement it this way, but I just don't see how splitting approaches and articles out into separate documents will help students. 

@cmcaine The logic here is that approaches have a wider future in their potential. One example would be that the analyzer can tell a student which approach they used to solve it and suggest other contrasting approaches. Another would be grouping community solutions by approach automatically etc. So the aim of having multiple approaches is that they can be used in more fine-grained ways.

The articles are then intended to contrast different ways of doing things. For example, the C# reverse-string exercise has a nice article on performance tradefoffs with benchmarking etc, which digs really deep into that, without overloading the wider pages around things.

As a starting point, having a single `introduction.md` that covers things is a totally pragmatic launch-point and a good way to get the content into the repo and so I'm :+1: on merging this if you're done for now (and grateful for the great work!), but I also definitely also see the value in splitting those things into individual files etc over time as Erik suggests, but someone else can take that work on - it shouldn't necessarily be your job just because you're doing this first step 🙂

ErikSchierboom:
Thanks Jeremy for helping explain things. I'll post my thoughts too.

> It feels a bit disrespectful to say after the staff have put the work in to implement it this way, but I just don't see how splitting approaches and articles out into separate documents will help students.

No disrespect felt! The dig deeper section has the following (optional) components:

- Approaches: these describe how one could solve an exercise in a particular way. They should show lots of code and explain _in detail_ how the approach solves the exercise.
- An approach introduction document: this document can contain general guidance about the exercise and should list the idiomatic approaches, including code. It can link to non-idiomatic approaches, but shouldn't show their code. You can think of this document as an easy way for a student to check what idiomatic approaches there are and then to decide which approach they want to learn more about
- Articles: they can be anything that is not an approach. An example being a performance analysis that compares multiple approaches (possibly also including non-idiomatic solutions), but could also be something else.
- Community videos: these are videos related to the exercises, e.g. walkthroughs or a discussion of an exercise

> Certainly it will hinder maintainers a little because we'll have to do the initial break-out and then edit two files instead of one (either by making a little tool or manually) - not loads of work, but not nothing.

This is only true if you take mentoring notes as a starting point, which doesn't necessarily have to be the case. All the approaches I've written, I've done so from scratch. Also, the two files (approach introduction and invididual approach document) are _not_ meant be copies of each other. The introduction should just list an outline of the code (with non-key parts stripped away), describe the approach and then link to the individual approach where the student can learn about the approach in detail.

> Maybe there's a design doc for the dig deeper feature that I can read that will convince me that I should do the work? The info on the docs site is quite sparse.

That's true. It is still quite a new feature and we have some, but not much documentation at https://exercism.org/docs/building/tracks/approaches and https://exercism.org/docs/building/tracks/articles. Sorry about that.

> convince me that I should do the work

You don't have to do the work if you don't feel like it. Someone else could do the work, or I could write a script to help extract the contents into approach documents, as that aligns with how we've designed the dig deeper feature.

Just to be clear, I think this PR is fantastic! It adds a ton of fantastic content to the Julia track, it might just need some iterating to have it integrate the best way it can on the website.

cmcaine:
Okay, I think everyone agrees that this is okay to merge now, so I suggest that we do this now (a staff member needs to do the merge), and tomorrow I'll rebase the PR that fixes CI for this track and I'll see what the material looks like when it's deployed and maybe take you up on the offer of writing a script, Eric.

> This is only true if you take mentoring notes as a starting point...

That's mostly fair.

exercises/concept/leap/.approaches/introduction.md
```diff
@@ -0,0 +1,41 @@
+# Introduction
```
iHiD, line 1: I've noticed that this is now a concept exercise. Should it actually be so as Julia doesn't have a Syllabus yet?

iHiD:
Looking great though: https://exercism.org/tracks/julia/exercises/atbash-cipher/dig_deeper

PR #523, opened 2022-11-14
Silence configlet warnings

cmcaine:
To be merged after #522.

I'm still not sure this is a good idea, but this PR moves the files so that configlet stops complaining about our partially done concepts and concept exercises.


cmcaine:
This is ready to merge now, pls review and approve.

SaschaMann (approved):
It's annoying that configlet still handles this case poorly but this seems like a good workaround.

PR #525, opened 2022-11-17
Approaches: tweaks

ErikSchierboom:
- approaches: use consistent general guidance headings
- approaches: use consistent "## Approach: " prefix

This is to prepare for extracting the approaches to individual approach documents.

### DIFF
--- a/exercises/practice/nucleotide-count/.approaches/introduction.md
+++ b/exercises/practice/nucleotide-count/.approaches/introduction.md
@@ -6,7 +6,7 @@
 - You can initialise a `Dict` with a generator, if you like: `Dict(base => 0 for base in "ACGT")`
 - In Julia, the properties of objects are generally considered private unless documented otherwise, so we generally prefer `value in keys(dict)` or `haskey(dict, value)` over `value in dict.keys`
 
-## Using a `Dict`
+## Approach: using a `Dict`
 
 Straightforward solution. Easily understood, but a bit slow because it will compute the hash of a character twice on each iteration (though any of these solutions is plenty fast enough for even pretty large inputs!).
 
@@ -24,7 +24,7 @@ function count_nucleotides(strand)
 end
 ```
 
-## Using four variables
+## Approach: using if/elseif
 
 We can get some speedup with a simple stack of if/elseifs like this:
 
@@ -48,7 +48,7 @@ function count_nucleotides2(strand)
 end
 ```
 
-## Using `count`
+## Approach: using `count`
 
 ```julia
 function count_nucleotides3(strand)
@@ -66,7 +66,7 @@ It can be surprising to some people that this solution using `count(==(base), st
 This is because the complexity of your loop body matters.
 In this case, we swap one iteration of a moderately complex loop body for four iterations of a very simple loop body.
 
-## Going real fast by counting bytes
+## Approach: counting bytes
 
 This solution, though perhaps a little harder to understand, is much faster at the price of just a little memory use:
 
--- a/exercises/practice/nucleotide-count/.approaches/introduction.md
+++ b/exercises/practice/nucleotide-count/.approaches/introduction.md
@@ -24,7 +24,7 @@ function count_nucleotides(strand)
 end
 ```
 
-## Approach: using if/elseif
+## Approach: using a variable for each type of nucleotide
 
 We can get some speedup with a simple stack of if/elseifs like this:
 
--- a/exercises/practice/nucleotide-count/.approaches/introduction.md
+++ b/exercises/practice/nucleotide-count/.approaches/introduction.md
@@ -1,11 +1,10 @@
 # Introduction
 
-## Guidance
+## General guidance
 
 - If your solution didn't use a `Dict` in its main loop and you don't know how it could have, then it's worth making sure you understand the first example solution before you move on!
 - You can initialise a `Dict` with a generator, if you like: `Dict(base => 0 for base in "ACGT")`
 - In Julia, the properties of objects are generally considered private unless documented otherwise, so we generally prefer `value in keys(dict)` or `haskey(dict, value)` over `value in dict.keys`
-
 
 ## Using a `Dict`
 
@@ -90,7 +89,7 @@ Its speed comes from three sources:
 
 1. Iterating a vector of bytes instead of characters (though this algorithm is still unicode-safe).
    This is faster because iterating characters from a `String` involves scanning the same vector of UTF-8 bytes but also doing a little work every iteration to establish if the current byte is the start of a multibyte sequence (and, if so to reassemble that sequence into its proper 32 bit character).
-2. The loop body is extremely simple and fast on many processors (load a value, increment a value at an offset*) and branchless (there are no conditionals within the loop body).
+2. The loop body is extremely simple and fast on many processors (load a value, increment a value at an offset\*) and branchless (there are no conditionals within the loop body).
    Branchless code is often faster because branching hinders CPU instruction level parallelism and stops compilers using "Single Instruction Multiple Data" (SIMD) instructions (though this particular loop body can't benefit from SIMD anyway).
 3. Bounds checks are turned off with `@inbounds`, this removes some hidden branches (normally every array access in Julia has a little check that the index is valid first).
    We can trivially prove that all indexes will be valid because a UInt8 has values ranging from 0:255 and the array we're indexing has indexes 1:256. The `+ 1` is there to make those two ranges line up.
--- a/exercises/practice/pangram/.approaches/introduction.md
+++ b/exercises/practice/pangram/.approaches/introduction.md
@@ -1,10 +1,9 @@
 # Introduction
 
-## Common suggestions
+## General guidance
 
 - You can express this whole problem as lowercasing the input and a single set relationship. Can you work out which relation it is? (answer below)
 - Instead of writing out the whole alphabet, you can use a character range `'a':'z'`
-
 
 ## Using set relations
 
--- a/exercises/practice/pangram/.approaches/introduction.md
+++ b/exercises/practice/pangram/.approaches/introduction.md
@@ -31,7 +31,7 @@ This approach uses an anonymous function and uppercases the character rather tha
 ispangram2(input) = all(c -> in(c, input) || in(uppercase(c), input), 'a':'z')
 ```
 
-## Approach: bitwise tracking
+## Approach: marking bytes
 
 And here's a faster, more complicated solution,
 just in case pangram-identification is on the critical path for your new startup.
--- a/exercises/practice/pangram/.approaches/introduction.md
+++ b/exercises/practice/pangram/.approaches/introduction.md
@@ -25,7 +25,7 @@ ispangram1(input) = all(in(lowercase(input)), 'a':'z')
 
 ## Approach: use an anonymous function
 
-The second uses an anonymous function and avoids taking a copy of the input to reduce memory usage.
+This approach uses an anonymous function and uppercases the character rather than the string to avoid making a copy of the input.
 
 ```julia
 ispangram2(input) = all(c -> in(c, input) || in(uppercase(c), input), 'a':'z')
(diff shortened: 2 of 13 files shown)


exercises/practice/nucleotide-count/.approaches/introduction.md
```diff
@@ -25,7 +24,7 @@ function count_nucleotides(strand)
 end
 ```
 
-## Using four variables
+## Approach: using if/elseif
```
cmcaine, line 27:
```suggestion
## Approach: using a variable for each nucleotide
```

I think this is the distinctive part of this approach, rather than the if/elseif

exercises/practice/pangram/.approaches/introduction.md
```diff
@@ -16,18 +15,23 @@ ispangram(input) = 'a':'z' ⊆ lowercase(input)
...
-The second uses an anonymous function and avoids taking a copy of the input to reduce memory usage.
+This approach uses function currying, which may look strange at first, but is very convenient for predicates like `in`.
 
 ```julia
 ispangram1(input) = all(in(lowercase(input)), 'a':'z')
+```
+
+## Approach: use an anonymous function
 
+The second uses an anonymous function and avoids taking a copy of the input to reduce memory usage.
```
cmcaine, line 28:
```suggestion
This approach uses an anonymous function and uppercases the character rather than the string to avoid making a copy of the input.
```

exercises/practice/pythagorean-triplet/.approaches/introduction.md
```diff
@@ -35,8 +35,7 @@ function pythagorean_triplets(n)
 end
 ```
 
-
-### Quadratic time algorithm
+### Approach: quadratic time algorithm
```
cmcaine, line 38:
```suggestion
## Approach: quadratic time algorithm
```

exercises/practice/pythagorean-triplet/.approaches/introduction.md
```diff
@@ -84,11 +83,10 @@ because `b` and `c` must both be larger.
 Similarly, the biggest `b` can be is `n/2 - ε` because `c` must be larger than it.
 
 When we move that to the integers, we observe that the maximum remainder of `n/3` is 2.
-We already know that the biggest `a` can be is  when `a`, `b`, and `c` are as close as possible, so that gives us something like this: `fld(n, 3) - 1 < fld(n, 3) + x < fld(n, 3) + y` where `x < y && -1 + x + y < 2`.
+We already know that the biggest `a` can be is when `a`, `b`, and `c` are as close as possible, so that gives us something like this: `fld(n, 3) - 1 < fld(n, 3) + x < fld(n, 3) + y` where `x < y && -1 + x + y < 2`.
 Similarly, the biggest remainder of `n/2` is 1, so the largest `b` can be is when our variables are `1 < fld(n, 2) - 1 < fld(n, 2) + 1`, so an upper bound on `b` is `fld(n, 2) - 1`.
 
-
-### Linear time algorithm
+### Approach: linear time algorithm
```
cmcaine, line 89:
```suggestion
## Approach: linear time algorithm
```

exercises/practice/robot-name/.approaches/introduction.md
```diff
@@ -116,9 +115,9 @@ it is often useful to consider the tradeoffs of different ways of representing i
 
 Converting names to integers and back again might seem tricky at first.
 A decent intuition is that the names are a number, but in an inconsistent base.
-Just as 153 is 3 + 5 * 10 + 1 * 100,
+Just as 153 is 3 + 5 _ 10 + 1 _ 100,
```
cmcaine, line 118:
```suggestion
Just as 153 is 3 + 5 * 10 + 1 * 100,
```

The *s here are for multiplication, not emphasis. They don't need to be escaped if they're surrounded by whitespace.

exercises/practice/robot-name/.approaches/introduction.md
```diff
@@ -116,9 +115,9 @@ it is often useful to consider the tradeoffs of different ways of representing i
...
 "AC5" could be interpreted as a number with two digits in base 26 and one in base 10:
-0 + 2 * 26 + 5 * 26 * 10
+0 + 2 _ 26 + 5 _ 26 \* 10
```
cmcaine, line 120:
```suggestion
0 + 2 * 26 + 5 * 26 * 10
```

exercises/practice/rotational-cipher/.approaches/introduction.md
```diff
@@ -5,7 +5,7 @@
 - The key lesson from this exercise is that you can define two methods for the function `rotate`
 - The macro/metaprogramming is quite challenging for lots of students. Feel free to read the answer below!
 
-## Example solution
+## Approach: modulo arithmetic
```
cmcaine, line 8: Almost any exercise with rotation will use modular arithmetic at some point, so I don't think this is a great name, but it's hard to think of a good one!

```suggestion
## Approach: using `Char` arithmetic and `map`
```

This is also generic, because most solutions will use `map`, but it's a bit more specific, imo.

exercises/practice/robot-name/.approaches/introduction.md
```diff
@@ -31,8 +31,7 @@ There's some basic stats to help you grasp the issue:
     Probability of finding the last name with N guesses =
       1 - ((676000 - 1) / 676000)^N
 
-
-## A fancy solution
+## Approach: shuffle all robot names in advance
```
cmcaine, line 34:
```suggestion
## Approach: encode names as numbers and shuffle them all in advance
```

Encoding names as numbers is the trickiest part of this, so let's lede (geddit) with that

exercises/practice/pythagorean-triplet/.approaches/introduction.md
```diff
@@ -1,15 +1,15 @@
 # Introduction
 
-## Hints and tips
+## General guidance
 
 - A O(n^3) solution is trivial to find, but too slow for the tests. A O(n^2) solution is pretty easy to find. A O(n) solution requires solving some simultaneous equations.
 
-## Example solutions
+## Approaches
```
cmcaine, line 7:
```suggestion
```

exercises/practice/pythagorean-triplet/.approaches/introduction.md
```diff
@@ -1,15 +1,15 @@
 # Introduction
 
-## Hints and tips
+## General guidance
 
 - A O(n^3) solution is trivial to find, but too slow for the tests. A O(n^2) solution is pretty easy to find. A O(n) solution requires solving some simultaneous equations.
```
cmcaine, line 6:
```suggestion
```

exercises/practice/pythagorean-triplet/.approaches/introduction.md
```diff
@@ -1,15 +1,15 @@
...
 
-## Example solutions
+## Approaches
 
 There are three reasonably common solutions to this problem, a cubic time solution, a quadratic time solution and a linear time solution
 (if those terms are unclear to you, you might like to read about [time complexity](https://en.wikipedia.org/wiki/Time_complexity)).
```
cmcaine, line 10:
```suggestion
There are three reasonably common solutions to this problem, a cubic time solution, a quadratic time solution and a linear time solution
(if those terms are unclear to you, you might like to read about [time complexity](https://en.wikipedia.org/wiki/Time_complexity)).
A cubic time solution is easy to find, but too slow for the tests or for many practical uses.
A quadratic time solution is pretty easy to find and will be good enough for the tests.
Finding a linear time solution will require some mathematics. The solution below was derived by solving some simultaneous equations.
```

The three suggestions above combine the "General guidance" and "Approaches" sections.

exercises/practice/pangram/.approaches/introduction.md
```diff
@@ -16,18 +15,23 @@ ispangram(input) = 'a':'z' ⊆ lowercase(input)
...
+
+```julia
 ispangram2(input) = all(c -> in(c, input) || in(uppercase(c), input), 'a':'z')
 ```
 
-## Going faster with bytes
+## Approach: bitwise tracking
```
cmcaine, line 34:
```suggestion
## Approach: marking bytes
```

Kinda bikesheddy, but I like that this and "counting bytes" have similar names because they're related approaches.

cmcaine (approved):
Thanks!

cmcaine:
I'll add all my suggestions above and merge this evening if there are no comments :)

ErikSchierboom:
Thanks for the review! I've applied all suggestions.

cmcaine:
Thanks again for the interest and the PR :)

I'm still not sure that splitting them out will actually be better. Is it okay for this track to opt out of that for now and maybe Sascha and I can think about it?

iHiD:
@cmcaine I'm happy for you personally to opt out, but not the track. We want to keep things consistent across the Exercism ecosystem as every time we don't it just makes life more confusing for students and means we have to build exceptions into our thinking which slows us down lots. So it's absolutely fine for you to choose not to split them out and keep adding things these wider documents, but if someone does split them out, we'd want to merge that PR 🙂 

Btw, if you want to read some of the discussion that went into this, a lot of it is documented at https://github.com/exercism/exercism/issues/6520 (lots of this conversation occurred on the weekly community calls, but I tried to put all the salient bits in that issue)

cmcaine:
Well, I'm not going to pointlessly make work for people by doing it my own way if someone is then going to come along and reformat it.

I'm not going to tell you what to do, nor am I demanding that you explain yourself, but this is why I would prefer the Julia track opt out. Unless you have a good reason not to, I ask that you take it into consideration as you develop these feature and interact with maintainers/authors regarding it.

Splitting the approaches out into their own pages has costs as well as benefits. I have asked "why" a few times, and here are the benefits I have understood from your messages and links:

1. from Jeremy: that future tooling might be able to match student's solutions with approaches and do some clever stuff with that
2. from inspecting the C# dig-deeper pages myself: that it is easier for students to discover community videos if the approaches introduction page is shorter
3. guessing at future features from the discussion thread: if we ever have many many approaches, or user interaction on them, then a page per approach is better

Here are the main things I think a single page strategy does well, that a multipage strategy loses:

1. visibility of content: don't have to click a link to discover that there's a lot written about a topic & easier to scan or search for content
2. easier to write and especially easier to write a narrative if you know what content the reader has already been exposed to (they might not have read it, but they will have usually at least scrolled past it)
4. ease of use (no user interaction required beyond scrolling to access all information)

I feel like, at least on the julia track, the "representer" motivated strategy is unproven both in its means:

- when we prototyped a representer it was hard to categorise solutions well and it seemed harder for more complex exercises
- development of the future tooling is speculative

and its supposed gains: 

- no decision yet (as far as I can tell) on how the representer and approaches will be used
- no trials to test effectiveness of more advanced strategies versus simpler strategies that are well proven in other fields (text explaining expert solutions)

cmcaine:
All that said, the bigger issue is about how exercism makes the dig deeper page visible to the student at all. When I finish an exercise it isn't advertised to me, it's just one more tab on quite a busy page.

iHiD:
> I have asked "why" a few times, and here are the benefits I have understood from your messages and links:

Ah, I've not really answered why approaches are individual things. I only answered why approaches and articles are different. 

The main reason they're individual things is because it makes discovery much easier for students.

Take for example in the C# example, students can very quickly grok all the approaches and explore ones that are unfamiliar to them.

Screenshot 2022-11-17 at 22 17 06
Screenshot 2022-11-17 at 22 17 55
Screenshot 2022-11-17 at 22 17 00

With a long-post as the main format, we're asking them to scroll through pages of text to be able to quickly get an idea of the different approaches and detect what's interesting to them.

(Compare Julia's atbash cipher for example):
Screenshot 2022-11-17 at 22 17 38

That ability for a student to quickly decide where to put their attention is the most important reason for these being split out. In my experience, most people faced with a wall of text immediately just opt out, whereas when people have quickly scannable information, they're able to determine what they want and act on it.

> All that said, the bigger issue is about how exercism makes the dig deeper page visible to the student at all. When I finish an exercise it isn't advertised to me, it's just one more tab on quite a busy page.

Yep - this page needs work in general. We have various quick ideas for this, not least using the blue-dot paradigm we have within Exercism so that exercise is shown as needing you to look at something until it's read, but also want to reconsider this whole page layout in the future.

PR #531, opened 2022-12-04
Reformat exercise config files

kytrinyx:
This runs the `bin/configlet fmt --update` command to
reformat the exercise configs.

### DIFF
--- a/exercises/practice/acronym/.meta/config.json
+++ b/exercises/practice/acronym/.meta/config.json
@@ -1,5 +1,4 @@
 {
-  "blurb": "Convert a long phrase to its acronym",
   "authors": [
     "PseudoCodeNerd"
   ],
@@ -17,6 +16,7 @@
       ".meta/example.jl"
     ]
   },
+  "blurb": "Convert a long phrase to its acronym",
   "source": "Julien Vanier",
   "source_url": "https://github.com/monkbroc"
 }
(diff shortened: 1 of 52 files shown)


SaschaMann (changes requested):
This moves the most important info in that file down for no apparent reason. It should not be merged as it is imo.

kytrinyx:
This is the behavior of Configlet. 
The order of keys is alphabetical. This has been reviewed, approved, and released for over a year.

I am in favor of leaving fmt the way it is, as I value the consistency and predictability of the choices that we're made.

Please reconsider merging this PR.

If you care so strongly about this that you do not wish to merge the PR, then please do what it takes to reach consensus with staff and maintainers so that all tracks can use the automation consistently.

SaschaMann:
That's an arbitrary order for a file that's meant to be edited and updated by humans and there's no "why" that explains why this change is useful or should be made.

kytrinyx:
@SaschaMann I'm not saying that you're wrong. I'm asking you to help move things forward.

I see two solutions here:

1. Configlet remains as is.
2. Configlet changes.

You can decide that you don't actually care that much about this particular change, or you can decide that you do, in fact, care a lot about this particular change, in which case it is up to you to convince all the other maintainers and the Exercism staff that your way is better.

Whatever the decision ends up being, that's a decision that I expect all track maintainers to follow.

> This isn't the first time weird workarounds are needed due to configlet behaviour that isn't well thought out

Did you suggest an improvement to Configlet for that case? For example, if the track's `config.json` had exercises and concepts that are in a `wip` state, perhaps Configlet could refrain from linting or fmting them.

PR #541, opened 2023-01-24
Arithmetic operators update

SJ-Nosrat:
### Updated - `concepts.wip/arithmetic-operators`
- blurb for arithmetic-operators,
- links for arithmetic-operators,
- introduction for arithmetic-operators.

### DIFF
--- a/concepts.wip/arithmetic-operators/.meta/config.json
+++ b/concepts.wip/arithmetic-operators/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for arithmetic-operators concept",
-  "authors": [],
-  "contributors": []
+  "blurb": "Arithmetic operations are supported on all (numeric) primitive types such as: integers and floating-point numbers.",
+  "authors": ["SaschaMann"],
+  "contributors": ["SJ-Nosrat"]
 }
--- a/concepts.wip/arithmetic-operators/.meta/config.json
+++ b/concepts.wip/arithmetic-operators/.meta/config.json
@@ -1,5 +1,5 @@
 {
   "blurb": "Arithmetic operations are supported on all (numeric) primitive types such as: integers and floating-point numbers.",
-  "authors": ["SaschaMann"],
-  "contributors": ["SJ-Nosrat", "cmcaine"]
+  "authors": ["SJ-Nosrat"],
+  "contributors": ["SaschaMann", "cmcaine"]
 }
--- a/concepts.wip/arithmetic-operators/.meta/config.json
+++ b/concepts.wip/arithmetic-operators/.meta/config.json
@@ -1,5 +1,5 @@
 {
   "blurb": "Arithmetic operations are supported on all (numeric) primitive types such as: integers and floating-point numbers.",
   "authors": ["SaschaMann"],
-  "contributors": ["SJ-Nosrat"]
+  "contributors": ["SJ-Nosrat", "cmcaine"]
 }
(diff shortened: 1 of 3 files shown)


concepts.wip/arithmetic-operators/.meta/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for arithmetic-operators concept",
-  "authors": [],
-  "contributors": []
+  "blurb": "Arithmetic operations are supported on all (numeric) primitive types such as: integers and floating-point numbers.",
+  "authors": ["SaschaMann"],
+  "contributors": ["SJ-Nosrat", "cmcaine"]
```
SaschaMann, line 4: I didn't write this, your name should be in the author field :)

concepts.wip/arithmetic-operators/introduction.md
```diff
@@ -3,3 +3,26 @@
 Julia provides a number of [arithmetic operators](https://en.wikipedia.org/wiki/Arithmetic#Arithmetic_operations) for numeric types[^1].
 
 [^1]: You can find a list of them in the [Julia Manual](https://docs.julialang.org/en/v1/manual/mathematical-operations/#Arithmetic-Operators).
+
+Many of the arithmetic operations are straightforward. Below are examples of some operations in Julia that may have different **syntax** in other languages, ex: Python.
```
SaschaMann, line 7: Two notes:

- It's often better to avoid language that calls something "straightforward", "easy" etc. to not discourage people who may not find it as straightforward. (even if in this particular case that seems unlikely)
- A comparison with a specific language will likely only be useful for a small subset of people. Perhaps we can keep the spirit of this sentence but change it to something like "Below are examples of some arithmetic operations in Julia. You may find that Julia embraces unicode operators more than many other languages."

concepts.wip/arithmetic-operators/introduction.md
```diff
@@ -3,3 +3,26 @@
...
+
+```julia
+# Exponentiation
+julia> 3 ^ 2
+9
+```
+
+```julia
+# Integer Division
+julia> div(5, 2)
```
SaschaMann, line 17: This is about operators, so let's use the operator form even if they're equivalent:

```suggestion
# Integer Division
# ÷ can be typed as \div<tab>
julia> 5 ÷ 2
```

concepts.wip/arithmetic-operators/introduction.md
```diff
@@ -3,3 +3,26 @@
...
+2
+```
+
+The latter operation after performing division is **truncated** to an integer.
+
+```julia
+# Modulo
+# 7 % 2 syntax is also supported
+julia> rem(7, 2)
```
SaschaMann, line 26: Same as above

SaschaMann (commented):
Thanks for the PR!

Left a few minor comments.

concepts.wip/arithmetic-operators/.meta/config.json
SJ-Nosrat: Done.

concepts.wip/arithmetic-operators/introduction.md
SJ-Nosrat: Applied change as per suggestion.

concepts.wip/arithmetic-operators/introduction.md
SJ-Nosrat: Applied as per suggestion.

concepts.wip/arithmetic-operators/introduction.md
SJ-Nosrat: Done.

SaschaMann (approved):
Thanks for the PR!

PR #542, opened 2023-01-25
Arithmetic operators update for about page

SJ-Nosrat:
### Updated - concepts.wip/arithmetic-operators/about.md

- Added detailed content for about page related to arithmetic operators in Julia.

Looking forward to your review and suggestions.

Thanks.

### DIFF
--- a/concepts.wip/arithmetic-operators/.meta/config.json
+++ b/concepts.wip/arithmetic-operators/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for arithmetic-operators concept",
-  "authors": [],
-  "contributors": []
+  "blurb": "Arithmetic operations are supported on all (numeric) primitive types such as: integers and floating-point numbers.",
+  "authors": ["SaschaMann"],
+  "contributors": ["SJ-Nosrat"]
 }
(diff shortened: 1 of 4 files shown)


SaschaMann (approved):
Thanks for the PR!

I've not been able to do a thorough review but in order to stick to optimistic merge, I'll go ahead and merge it already.

cmcaine:
Thanks for the contribution! Do the Julia docs talk about "primitive numeric types"? I don't remember them doing so and unless they do I don't think we should use the word "primitive" because it will imply something untrue (e.g. the Java-style distinction between primitive and other types).

SJ-Nosrat:
Hi @cmcaine,
Thanks for your feedback. From my understanding on reading the Julia Docs - link [here](https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/) as per their quote: "Built-in representations of such values are called numeric primitives".

However, if my understanding is incorrect, I'll be happy to make the change and remove it.

Looking forward to your reply.

SaschaMann:
I was equally surprised before seeing that the manual uses the term. Especially since being told there's basically never a reason to define primitive types other than bootstrapping Julia when we considered creating an exercise thst teaches them.

But given the manual uses it, I don't object.

cmcaine:
Sorry for the noise, then. No objection here. I should have checked but I was on my phone and lazy.

PR #550, opened 2023-02-06
If expressions update

SJ-Nosrat:
## Update - `julia/concepts.wip/if-expressions`

- Add blurb to `config.json`
- Add intro to `introduction.md`
- Add about to `about.md`

### DIFF
--- a/concepts.wip/if-expressions/.meta/config.json
+++ b/concepts.wip/if-expressions/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for if-expressions concept",
+  "blurb": "if-expressions control the flow of a code block by allowing to execute certain statements, given that the conditional expression is satisfied.",
   "authors": [],
   "contributors": []
 }
--- a/concepts.wip/if-expressions/.meta/config.json
+++ b/concepts.wip/if-expressions/.meta/config.json
@@ -1,5 +1,5 @@
 {
   "blurb": "if-expressions control the flow of a code block by allowing to execute certain statements, given that the conditional expression is satisfied.",
-  "authors": [],
+  "authors": ["SJ-Nosrat"],
   "contributors": []
 }
--- a/concepts.wip/if-expressions/about.md
+++ b/concepts.wip/if-expressions/about.md
@@ -18,7 +18,7 @@ The `function` defined below is not very realistic, but it shows the anatomy of 
 ```julia
 julia> function password_checker(password)
            if password == system_password
-               println("Welcome back!")
+               return "Welcome back!"
             end
         end
 
--- a/concepts.wip/if-expressions/about.md
+++ b/concepts.wip/if-expressions/about.md
@@ -1,52 +1,67 @@
 # About
 
-`if`-expressions in Julia are similar to those seen in other languages:
+In Julia, like many other languages, allow for the following with `if`-expressions:
+
+- Chained if-expressions
+- Nested if-expressions
+
+~~~~exercism/note
+`if`-expressions as defined here will encompass the following syntaxes:`if`, `if-else` and `if-elseif-else` statements.
+~~~~
+
+`if`-expressions allows for branching in our programs, given a conditional is satisfied. 
+
+We can do the following for effective branching:
+
+## Chained expressions example
 
 ```julia
-julia> function say_if_positive(n)
-           if n > 0
-               println("n is positive!")
-           else
-               println("n is not positive!")
-           end
-       end
-say_if_positive (generic function with 1 method)
-
-julia> say_if_positive(10)
-n is positive!
-
-julia> say_if_positive(-10)
-n is not positive!
+# We can test for multiple cases
+# with chained if expression.
+if 0 ≤ n ≤ 0.5
+    println("n is in the interval [0, 0.5], inclusive.")
+elseif -0.5 ≤ n < 0
+    println("n is in the interval [-0.5, 0).")
+elseif n > 0.5
+    println("n is greater than 0.5.")
+else
+    println("n is less than -0.5.")
+end
 ```
 
-<!-- TODO: Add that fancy concept highlight embed thing to boolean expression -->
+In the above chained expression, we can have mulitple `elseif` statements, with only one `else` statement at the end, but it's not necessary. 
 
-If the boolean expression following the `if` evaluates to `true`, the first block of code is run and the second block is skipped.
-If the boolean expression following the `if` evaluates to `false`, the first block of code is skipped and the second block is run.
-The program continues running at the first line of code after the `end` keyword.
-
-In cases where the second block of code would be just another `if`-expression, `elseif` allows us to avoid nesting `if`-expressions within the block:
+## Nested expressions example
 
 ```julia
-julia> function dessert(fruit)
-           if fruit == "apple"
-               return "Apple Crumble"
-           elseif fruit == "lemon"
-               return "Lemon Meringue Pie"
-           else
-               return "Fruit Salad"
-           end
-       end
-dessert (generic function with 1 method)
-
-julia> dessert("apple")
-"Apple Crumble"
-
-julia> dessert("lemon")
-"Lemon Meringue Pie"
-
-julia> dessert("peach")
-"Fruit Salad"
+# We can refine our coniditonal evaluation
+# with nested if expressions.
+if n ≥ 0
+    if 0.5 ≤ n ≤ 1
+        println("n is in the interval [0.5, 1], inclusive.")
+    else
+        println("n is in the interval [0, 0.5).")
+    end
+else
+    println("n is negative!")
+end
 ```
 
-If an `if`-expression only needs to perform code for one of the cases, there's no need to write out the `else` branch.
+## Order matters
+
+```julia
+# Suppose you guessed the number 2,
+# then the if block will be executed,
+# and the others skipped.
+if guess_number > 0
+    println("The number guessed is greater than 0.")
+elseif guess_number > 1
+    println("The number guessed is greater than 1.")
+else
+    println("n is negative!")
+end
+```
+In the above `if` expression, it is instructive to note that even though the conditional expression in the `elseif` clause would evaluate to `true`, it is skipped,
+since the conditional in the `if` clause is `true` first.
+
+As such sometimes you'll get unexpected results depending on how you structure your `if` expression.
--- a/concepts.wip/if-expressions/about.md
+++ b/concepts.wip/if-expressions/about.md
@@ -11,25 +11,25 @@ In Julia, like many other languages, allow for the following with `if`-expressio
 
 `if`-expressions allows for branching in our programs, given a condition is satisfied.
 
-## A prompt program example
+## A first example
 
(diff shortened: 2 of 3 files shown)


concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,67 @@
 # About
 
-`if`-expressions in Julia are similar to those seen in other languages:
+In Julia, like many other languages, allow for the following with `if`-expressions:
+
+- Chained if-expressions
+- Nested if-expressions
+
+~~~~exercism/note
+`if`-expressions as defined here will encompass the following syntaxes:`if`, `if-else` and `if-elseif-else` statements.
```
SaschaMann, line 9:
```suggestion
`if`-expressions as defined here will encompass the following syntaxes:`if`, `if-else` and `if-elseif-else` statements.
```

This alludes to other syntaxes existing, could you perhaps add a link to their concept pages? I'm guessing the conditionals page?

concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,67 @@
...
+~~~~
+
+`if`-expressions allows for branching in our programs, given a conditional is satisfied.
```
SaschaMann, line 12:
```suggestion
`if`-expressions allows for branching in our programs, given a condition is satisfied. 
```

SaschaMann (commented):
I like the examples! I'm thinking perhaps it makes sense to have a very simple example without `else` or `else-if` above the more complicated ones. What do you think?

SJ-Nosrat:
> I like the examples! I'm thinking perhaps it makes sense to have a very simple example without `else` or `else-if` above the more complicated ones. What do you think?

Sure, that's a good idea. Where do you want to put the examples in? Intro or about page? or Both?

concepts.wip/if-expressions/about.md
SJ-Nosrat: Done.

concepts.wip/if-expressions/about.md
SJ-Nosrat: Done. Thanks.

SJ-Nosrat:
> > I like the examples! I'm thinking perhaps it makes sense to have a very simple example without `else` or `else-if` above the more complicated ones. What do you think?
> 
> Sure, that's a good idea. Where do you want to put the examples in? Intro or about page? or Both?

I've added an example using only an `if` statement. Please let me know if that's okay? Thanks for your valued feedback.

concepts.wip/if-expressions/about.md
```diff
@@ -6,10 +6,30 @@ In Julia, like many other languages, allow for the following with `if`-expressio
...
+password = readline()
+
+# Our main logic
+if password_checker(password)
+    println("Welcome!")
+end
+```
+Although the above program is not very realistic, but it shows the anatomy of an `if` statement. 
+
+In this case the `println("Welcome!")` is run, if the `password_checker(password)` returns `true`, otherwise it is skipped.
```
SaschaMann, line 32: This is mostly a personal view, not an Exercism view, even though some other maintainers agree:

Imho any examples with reading input or printing anything should be avoided completely. So many people new to programming struggle with understanding the difference between printing and returning a value, often because their introductory classes or texts use printing everywhere. This is even more confusing in REPL settings where the difference is visually indistinguishable. To avoid this, I'd only print in an exercise that's explicitly about IO, not general examples or other exercises.

The example itself works but perhaps it could be turned into a function `password_check(password)`?

SaschaMann:
> I've added an example using only an if statement in the about.md page. Please let me know if that's okay? Thanks for your valued feedback.

Works great!

concepts.wip/if-expressions/about.md
SJ-Nosrat: Thank you for explaining this. Understood, I've converted the example into a `function`, as suggested. Please let me know if the example is clear.

SaschaMann: That's clearer, though I'd change it to return a string

concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,87 @@
...
+        end
+
+# Assume system_password is "abc123"
+julia> password_checker("abc123")
+Welcome back!
+
+julia> password_checker("xyz321")
+
+```
+`println("Welcome back!")` is run, if the `password == system_password`, otherwise it is skipped, and nothing is printed to the screen.
```
cmcaine, line 32:
```suggestion
`"Welcome back!"` is returned if `password == system_password`, otherwise that block is skipped and `nothing` is returned.
```

Func returns a string now. Maybe leave out the bit about `nothing` being returned.

concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,87 @@
...
+    println("n is in the interval [-0.5, 0).")
+elseif n > 0.5
+    println("n is greater than 0.5.")
+else
+    println("n is less than -0.5.")
+end
 ```
 
-<!-- TODO: Add that fancy concept highlight embed thing to boolean expression -->
+In the above chained expression, we can have mulitple `elseif` statements, with only one `else` statement at the end, but it's not necessary.
```
cmcaine, line 52:
```suggestion
In a chained if/elseif/else expression, we can have zero or more `elseif` expressions, and at most one `else` expression at the end.
```

concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,87 @@
...
-<!-- TODO: Add that fancy concept highlight embed thing to boolean expression -->
+In the above chained expression, we can have mulitple `elseif` statements, with only one `else` statement at the end, but it's not necessary. 
+
+## Nested expressions example
 
-If the boolean expression following the `if` evaluates to `true`, the first block of code is run and the second block is skipped.
-If the boolean expression following the `if` evaluates to `false`, the first block of code is skipped and the second block is run.
-The program continues running at the first line of code after the `end` keyword.
+```julia
+# We can refine our coniditonal evaluation
```
cmcaine, line 57: Probably better to have the comments in the markdown rather than as code comments?

concepts.wip/if-expressions/introduction.md
```diff
@@ -1,57 +1,86 @@
 # Introduction
 
-`if`-expressions in Julia are similar to those seen in other languages:
+In Julia, `if`-expressions, are similar to those seen in other languages.
+
+~~~~exercism/note
+[`if`-expressions](https://docs.julialang.org/en/v1/manual/control-flow/#man-conditional-evaluation) as defined here will encompass the following syntaxes:`if`, `if-else` and `if-elseif-else` statements.
```
cmcaine, line 6:
```suggestion
[`if`-expressions](https://docs.julialang.org/en/v1/manual/control-flow/#man-conditional-evaluation) as defined here will encompass the following syntaxes:`if`, `if-else` and `if-elseif-else`.
```

Not statements

concepts.wip/if-expressions/introduction.md
```diff
@@ -1,57 +1,86 @@
...
+~~~~
+
+`if`-expressions provide the ability to execute certain statements in a _code block_ given that the condition is satisfied. In other words, it allows for branching in our programs.
+
+## `if` statement example
```
cmcaine, line 11:
```suggestion
## `if` example
```

concepts.wip/if-expressions/introduction.md
```diff
@@ -1,57 +1,86 @@
...
+
+say_if_positive (generic function with 1 method)
+
+# Again, -10 < 0, the if block is skipped,
+# else block is executed, as the only other 
+# alternative.
+julia> say_if_positive(-10)
+n is negative!
+```
+## `if`-`elseif`-`else` statement example
```
cmcaine, line 61:
```suggestion
## `if`-`elseif`-`else` example
```

concepts.wip/if-expressions/introduction.md
```diff
@@ -1,57 +1,86 @@
...
-If the boolean expression following the `if` evaluates to `false`, the first block of code is skipped and the second block is run.
-The program continues running at the first line of code after the `end` keyword.
 
 ~~~~exercism/note
 In Julia, the `end` keyword signifies the end of all block expressions.
 This syntax is not specific to `if`-expressions or function definitions.
 ~~~~
 
-In cases where the second block of code would be just another `if`-expression, `elseif` allows us to avoid nesting `if`-expressions within the block:
+## `if`-`else` statement example
```
cmcaine, line 42:
```suggestion
## `if`-`else` example
```

concepts.wip/if-expressions/introduction.md
```diff
@@ -1,57 +1,86 @@
...
+~~~~
+
+`if`-expressions provide the ability to execute certain statements in a _code block_ given that the condition is satisfied. In other words, it allows for branching in our programs.
+
+## `if` statement example
 
 ```julia
-julia> function say_if_positive(n)
+# Defining a function, allows to pass different values
+# to test. The if statement is the code block inside
```
cmcaine, line 15:
```suggestion
# to test. The if expression is the code block inside
```

concepts.wip/if-expressions/introduction.md
```diff
@@ -1,57 +1,86 @@
...
-               println("n is not positive!")
            end
        end
-say_if_positive (generic function with 1 method)
+say_positive (generic function with 1 method)
 
-julia> say_if_positive(10)
+# 10 > 0 which enters the if block
+# and executes the println("n is positive!") 
+# statement.
```
cmcaine, line 26:
```suggestion
# expression.
```

concepts.wip/if-expressions/.meta/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for if-expressions concept",
-  "authors": [],
+  "blurb": "if-expressions control the flow of a code block by allowing to execute certain statements, given that the conditional expression is satisfied.",
```
cmcaine, line 2:
```suggestion
  "blurb": "if-expressions let the program choose whether to execute an expression or not based on some condition.",
```

concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,87 @@
 # About
 
-`if`-expressions in Julia are similar to those seen in other languages:
+In Julia, like many other languages, allow for the following with `if`-expressions:
```
cmcaine, line 3:
```suggestion
Julia, like many other languages, allows for the following `if`-expressions:
```

concepts.wip/if-expressions/about.md
```diff
@@ -1,52 +1,87 @@
...
+
+~~~~exercism/note
+[`if`-expressions](https://docs.julialang.org/en/v1/manual/control-flow/#man-conditional-evaluation) as defined here will encompass the following syntaxes:`if`, `if-else` and `if-elseif-else` statements.
+~~~~
+
+`if`-expressions allows for branching in our programs, given a condition is satisfied.
+
+## A first example
+
+The `function` defined below is not very realistic, but it shows the anatomy of an `if` statement.
```
cmcaine, line 16:
```suggestion
The `function` defined below is not very realistic, but it shows the anatomy of an `if` expression. 
```

PR #564, opened 2023-03-14
largest-series-product: add exercise

ErikSchierboom:


config.json
```diff
@@ -801,6 +801,25 @@
...
         "name": "Alphametics",
         "difficulty": 10
+      },
+      {
+        "uuid": "4ce25733-55ab-48fd-84e2-c4b7f2f3f529",
+        "practices": [],
+        "prerequisites": [],
+        "slug": "largest-series-product",
+        "name": "largest-series-product",
+        "difficulty": 4,
```
ErikSchierboom, line 811: Let me know what you think of this difficulty

exercises/practice/largest-series-product/runtests.jl
```diff
@@ -0,0 +1,65 @@
...
+
+@testset "reports 1 for empty string and empty product 0 span" begin
+    @test largest_product("", 0) == 1
+end
+
+@testset "reports 1 for nonempty string and empty product 0 span" begin
+    @test largest_product("123", 0) == 1
+end
+
+@testset "error handling" begin
```
ErikSchierboom, line 49: I don't know if it makes sense to use other test sets too?

exercises/practice/largest-series-product/.meta/example.jl
```diff
@@ -0,0 +1,16 @@
...
+
+    for c in s
+        isdigit(c) || throw(DomainError(s, "String must only contain digits"))
+        popfirst!(digits)
+        push!(digits, c - '0')
+        largest = max(largest, prod(digits))
+    end
+
+    largest
+end
```
ErikSchierboom, line 16: I'm new to Julia so let me know if this is bad

exercises/practice/largest-series-product/.docs/instructions.md
```diff
@@ -0,0 +1,14 @@
+# Instructions
+
+Given a string of digits, calculate the largest product for a contiguous substring of digits of length n.
+
+For example, for the input `'1027839564'`, the largest product for a series of 3 digits is 270 `(9 * 5 * 6)`, and the largest product for a series of 5 digits is 7560 `(7 * 8 * 3 * 9 * 5)`.
+
+Note that these series are only required to occupy *adjacent positions* in the input; the digits need not be *numerically consecutive*.
+
+For the input `'73167176531330624919225119674426574742355349194934'`,
```
SaschaMann, line 9: I know this is probably due to prob-specs formatting, but perhaps we should drop the single quotes to prevent any misconceptions about the type of the inputs.

exercises/practice/largest-series-product/.meta/example.jl
```diff
@@ -0,0 +1,16 @@
+function largest_product(s::AbstractString, span::Int)
```
SaschaMann, line 1:
```suggestion
function largest_product(s, span)
```

Type annotation should mainly be used for dispatch, which isn't the case here. It's often a good idea to leave them as general as possible, so the examples on exercism afair mostly don't specify the type.

exercises/practice/largest-series-product/runtests.jl
SaschaMann: I think it's fine.

exercises/practice/largest-series-product/runtests.jl
```diff
@@ -0,0 +1,65 @@
...
+@testset "can find the largest product of 3" begin
+    @test largest_product("1027839564", 3) == 270
+end
+
+@testset "can find the largest product of 5 with numbers in order" begin
+    @test largest_product("0123456789", 5) == 15_120
+end
+
+@testset "can get the largest product of a big number" begin
+    @test largest_product("73167176531330624919225119674426574742355349194934", 6,) == 23_520
```
SaschaMann, line 30:
```suggestion
    @test largest_product("73167176531330624919225119674426574742355349194934", 6) == 23_520
```

exercises/practice/largest-series-product/.meta/example.jl
SaschaMann: I don't see anything critically wrong with it.

You could move the `isdigit` check up in the style of a guard clause (`all(isdigit, collect(s)) || throw(...)`). It's slower but might be cleaner.

exercises/practice/largest-series-product/.docs/instructions.md
ErikSchierboom: Yeah, this is coming from prob-specs. The single quotes will be there to show it is a string I assume, but maybe we should use double quotes?

cmcaine: Double quotes are more stringy, I think. In Julia and Elixir at least, double quotes are for strings and single quotes are for something else (characters in Julia, charlists in Elixir). I can't think of any languages that allow single quotes for strings and not double quotes.

exercises/practice/largest-series-product/runtests.jl
```diff
@@ -0,0 +1,65 @@
...
+    @test largest_product("", 0) == 1
+end
+
+@testset "reports 1 for nonempty string and empty product 0 span" begin
+    @test largest_product("123", 0) == 1
+end
+
+@testset "error handling" begin
+    @testset "span longer than string length" begin
+        @test_throws DomainError largest_product("123", 4)
```
cmcaine, line 51: I think we should probably require throwing ArgumentError instead.

DomainError is a bit awkward because you should probably to provide the value that is outside the domain as the first argument, which is ugly when the input is very long.

exercises/practice/largest-series-product/.docs/instructions.md
```diff
@@ -0,0 +1,14 @@
...
+the largest product for a series of 6 digits is 23520.
+
+For a series of zero digits, the largest product is 1 because 1 is the multiplicative identity.
+(You don't need to know what a multiplicative identity is to solve this problem;
+it just means that multiplying a number by 1 gives you the same number.)
```
cmcaine, line 14: It's a bit confusing that we say they don't need to know and then explain it (kinda incorrectly, because the concept isn't just for numbers).

For me at least, it was initially confusing that the product of no numbers is 1 (or why any([]) == false and all([]) == true). I think the description should explain why we do this (so prod([]) * prod(v) == prod(v) because v == [[]..., v...]).

cmcaine:
Sorry, I think my commit removing the skeleton has confused GitHub and now your commit removing the quotes wasn't applied. You may need to commit again or something.

exercises/practice/largest-series-product/runtests.jl
cmcaine: Off topic:

I misremembered, you can't attach arbitrary data to ArgumentError, it just contains a string. DomainError takes a value and an optional message string because it used to be (maybe still is?) more efficient to avoid string error messages because you're then not inlining any string stuff or allocations in hot loops that call e.g. sqrt or log. I think that was maybe fixed when structs became immutable and maybe as the compiler improved.

https://github.com/JuliaLang/julia/issues/10617

exercises/practice/largest-series-product/.docs/instructions.md
ErikSchierboom: I've opened a PR: https://github.com/exercism/problem-specifications/pull/2233

ErikSchierboom, line 9:
```suggestion
For the input `"73167176531330624919225119674426574742355349194934"`,
```

exercises/practice/largest-series-product/.docs/instructions.md
```diff
@@ -0,0 +1,14 @@
+# Instructions
+
+Given a string of digits, calculate the largest product for a contiguous substring of digits of length n.
+
+For example, for the input `'1027839564'`, the largest product for a series of 3 digits is 270 `(9 * 5 * 6)`, and the largest product for a series of 5 digits is 7560 `(7 * 8 * 3 * 9 * 5)`.
```
ErikSchierboom, line 5:
```suggestion
For example, for the input `"1027839564"`, the largest product for a series of 3 digits is 270 `(9 * 5 * 6)`, and the largest product for a series of 5 digits is 7560 `(7 * 8 * 3 * 9 * 5)`.
```

exercises/practice/largest-series-product/.docs/instructions.md
ErikSchierboom: Would you mind opening a PR on problem-specifications? (I would be hard pressed to phrase it well)

exercises/practice/largest-series-product/runtests.jl
SaschaMann: Ah, we had this before on Exercism too: https://github.com/exercism/julia/pull/286

Seems like an ArgumentError is more appropriate indeed.

ErikSchierboom: I'll update tomorrow!

exercises/practice/largest-series-product/.meta/example.jl
cmcaine: There's nothing especially wrong with it, though I think most people would use a series of views rather than pushing and popping from a vector.

I wrote an approaches doc for this exercise and you can see a few variations on how I solved the exercise there: https://github.com/exercism/julia/pull/564/files#diff-1e0a7d63ad5259cf1372acd0aedb41baae1d5ebc78add316c25473f3d4995b33

exercises/practice/largest-series-product/.approaches/introduction.md
```diff
@@ -0,0 +1,120 @@
...
+        best = max(best, prod(view(digits, left:right)))
+    end
+
+    return best
+end
+```
+
+This example uses some techniques or syntax that may be unfamiliar to you:
+
+- `span ∉ 0:length(str)` is equivalent to `!(span in 0:length(str))`. `∉` is the mathematical symbol for "not in".
```
SaschaMann, line 35:
```suggestion
- `span ∉ 0:length(str)` is equivalent to `!(span in 0:length(str))`. `∉` is the mathematical symbol for "not in" and can be typed in Julia editors using `\notin<tab>`.
```

cmcaine:
```suggestion
- `span ∉ 0:length(str)` is equivalent to `!(span in 0:length(str))`. `∉` is the mathematical symbol for "not in" and can be typed at the Julia REPL (and possibly in your editor) as `\notin<tab>`
```

exercises/practice/largest-series-product/.docs/instructions.md
cmcaine: Sure. https://github.com/exercism/problem-specifications/pull/2234

config.json
cmcaine: Seems okay to me.

exercises/practice/largest-series-product/runtests.jl
ErikSchierboom: Updated

cmcaine (commented):
It looks good to me, except the examples in the approaches file will need DomainError changed to ArgumentError too.

I can do that today, probably.

ErikSchierboom:
@cmcaine I've just updated the approaches

SaschaMann (approved):

ErikSchierboom:
Thanks both!

cmcaine:
Thank you!

PR #574, opened 2023-03-14
sync sum-of-multiples instructions

cmcaine:
Also removed .append as I don't think we need it any more.


kytrinyx:
The reason my script didn't do it is that it didn't find a `sum-of-multiples` entry in the `config.json` of the Julia track.

I'm now second-guessing everything I think I know about how this works 😂

kytrinyx:
Oh! It was added yesterday! I will change my script to pull from main before checking the config.json (which it should have done in the first place).

cmcaine:
I'm glad I was able to sorta help :)

SaschaMann (approved):

PR #592, opened 2023-03-17
saddle-points: add exercise

ErikSchierboom:


config.json
```diff
@@ -820,6 +820,25 @@
...
           "algorithms"
         ]
+      },
+      {
+        "uuid": "f26bf680-b8fb-40dd-a82c-884ae391b8b6",
+        "practices": [],
+        "prerequisites": [],
+        "slug": "saddle-points",
+        "name": "Saddle Points",
+        "difficulty": 4,
```
ErikSchierboom, line 830: Let me know if this is a reasonable difficulty

exercises/practice/saddle-points/runtests.jl
```diff
@@ -0,0 +1,48 @@
+using Test
+
+include("saddle-points.jl")
+
+@testset "Can identify single saddle point" begin
+    matrix = [9 8 7; 5 3 2; 6 6 7]
+    @test saddlepoints(matrix) == [(2, 1)]
```
ErikSchierboom, line 7: I opted for the return value to be a vector of tuples (and not `CartesianIndex` values) to give the student some freedom in how to approach this. Let me know if that is okay.

exercises/practice/saddle-points/.meta/example.jl
```diff
@@ -0,0 +1,8 @@
+function saddlepoints(matrix)
+    isempty(matrix) && return []
+
+    (cols_min,) = findmin(matrix, dims=1)
+    (rows_max,) = findmax(matrix, dims=2)
+
+    [Tuple(i) for i = CartesianIndices(matrix) if matrix[i] <= cols_min[i[2]] && matrix[i] >= rows_max[i[1]]]
```
ErikSchierboom, line 7: No idea if this is a reasonable implementation.

exercises/practice/saddle-points/runtests.jl
```diff
@@ -0,0 +1,48 @@
...
+    @test saddlepoints(matrix) == [(3, 3)]
+end
+
+@testset "Can identify saddle points in a non square matrix" begin
+    matrix = [3 1 3; 3 2 4]
+    @test saddlepoints(matrix) == [(1, 1), (1, 3)]
+end
+
+@testset "Can identify that saddle points in a single column matrix are those with the minimum value" begin
+    matrix = reshape([2; 1; 4; 1], :, 1)
```
ErikSchierboom, line 41: I had to do this to force it into a `1x4` vector, which I needed because of my usage of `CartesianIndices`, but this may be a red flag.

exercises/practice/saddle-points/.meta/example.jl
SaschaMann: It might be faster to not find the `rows_max` but only check if a column minimum is also a maximum in its row but I haven't tested it.

exercises/practice/saddle-points/runtests.jl
SaschaMann, line 1: Same renaming here

exercises/practice/saddle-points/runtests.jl
SaschaMann, line 41: You can also write this as `[2; 1; 4; 1;;]` if you prefer. For some reason that's slower but it could be easier to figure out because one doesn't have to learn about `reshape`?

cmcaine:
```suggestion
    matrix = reshape([2, 1, 4, 1], :, 1)
```

There's no need for the semicolons. Vectors in Julia are already column vectors (and reshape replaces the old shape of the matrix anyway).

I think reshape is the standard way to do this. The technique with extra semicolons gives different answers for different Julia versions.

SaschaMann: > Vectors in Julia are already column vectors 

The example doesn't pass the tests when doing this, it needs the one column matrix for the indices to work. But that probably means the example should be changed, not the tests.

cmcaine: I think it's fine for the example to require the input to be an AbstractMatrix and would be confusing for students if they had to support being given "matrices" that actually only have one dimension.

exercises/practice/saddle-points/saddle-points.jl
```diff
@@ -0,0 +1,2 @@
+function saddlepoints(matrix)
```
cmcaine, line 1:
```suggestion
function saddlepoints(M)
```

If renaming as Sascha suggested.

exercises/practice/saddle-points/runtests.jl
cmcaine:
```suggestion
    matrix = reshape([2, 1, 4, 1], :, 1)
```

ErikSchierboom:
I've updated

exercises/practice/saddle-points/.meta/example.jl
cmcaine: I think the original way is better. Better worst case behaviour. If we write an approaches intro we can talk about how to do it differently.

config.json
cmcaine: Seems.good to me.

exercises/practice/saddle-points/.meta/example.jl
ErikSchierboom: Okay cool! I'll revert

SaschaMann (approved):

PR #593, opened 2023-03-17
Explain how to automatically pick between a BitSet and a Set{Int}

cmcaine:
I'm not sure whether this should be included (or if it should be included, whether it should be included here), but I wrote the code for it while I was testing different approaches for the exercise, so I thought I might as well write it up.


SaschaMann (approved):

SaschaMann:
Looks like the bot can't handle it, could you rebase this @cmcaine? :)

cmcaine:
Done :)

Issue #567, opened 2023-03-14
learning mode: add stubs to exercises


ErikSchierboom:
Most concept exercises do not have stubs, which is contrary to our [specification](https://exercism.org/docs/building/tracks/concept-exercises#h-file-stub-implementation).
The exception to this rule is for newly introduced syntax, where we _want_ the student to write the syntax by themselves.

# Current status

This is the list of exercises for which stubs need to be added:

- [x] `annalyns-infiltration`
- [ ] `annalyns-infiltration2`
- [ ] `dnd-char`
- [x] `documented-lasagna`
- [ ] `elyses-analytic-enchantments`
- [x] `elyses-enchantments`
- [ ] `emoji-times`
- [x] `encounters` (has a minimal stub, but we want students to write the multiple dispatched functions themselves)
- [ ] `exercism-matrix`
- [ ] `fibonacci-iterator`
- [ ] `high-score`
- [x] `lasagna` (only comments as stub, we want students to write the functions themselves)
- [x] `leap` (will be deprecated)
- [ ] `name-badge`
- [ ] `stage-heralding`
- [ ] `tech-palace`
- [ ] `vehicle-purchase`

See https://github.com/exercism/julia/issues/563 for context

glennj:
> lasagna (no stub, but we want students to write the functions themselves)

I added a stub to #604. I figured as the very first learning exercise, having a blank canvas for the stub is too much to ask.

ErikSchierboom:
Well, maybe not entirely blank but something along the line of: https://github.com/exercism/elixir/blob/main/exercises/concept/lasagna/lib/lasagna.ex?

glennj:
lasagna can be ticked as done

colinleach:
See #615

keiravillekode:
`annalyns-infiltration` and `elyses-enchantments` can be marked as done.

In `currency-exchange` the stub function bodies contain 4 spaces (excellent) but in `cars-assemble` the stub functions bodies only contain 2 spaces.

colinleach:
@depial fixed this in #888. Thanks for reporting it. I have no idea why it happened, but that isn't the main point.

FYI, Julia is not picky about whitespace: the original design copied Matlab, not Python. However, we of course try to observe the 4-space convention.

PR #604, opened 2023-03-21
Lasagna concept exercise

glennj:
Reworked the `functions` concept a bit.
Added `integer-introduction` to fulfill the lasagna design doc. 
Moved comments from functions to its own concept.

I added a `config.json.wip` file to track the syllabus.
Closes #565

PR #615, opened 2023-03-26
Concept exercise stubs

colinleach:
Partial fix for #567, mostly early-stage exercises that are close to completion.


SaschaMann (commented):
Many of these would likely not be defined using `function` but inline as `f(x) = ...` instead. Stubs like these actively push students towards a direction when another direction would be just as valid or better. I say we ignore the spec and leave these empty as I think the stubs actively make the exercises worse. Stubs seem more useful in languages that require a lot of boilerplate code around the solutions, which is not the case for Julia.

Alternatively we could consider adding the functions but without any methods as stubs, e.g. `function is_foggy end`. Though that introduces a new concept on its own.

cmcaine:
I don't think we should have stubs that define functions without methods. That doesn't seem useful and would definitely be confusing.

I think the stubs are unnecessary but harmless. Yes, it encourages using the more verbose form, but students can easily change their solutions to use the shorter form if they want it. For students who don't want the stubs it's very easy to delete them.

colinleach:
I agree that stubs are unnecessary but harmless. They are also Exercism policy for concept exercises, so we're kind of stuck with them.

The different ways to define functions were covered in `lasagna`, which students had to complete before unlocking any of these exercises. Several other language tracks (including those highlighted in Functional February) have equivalent long/short forms, and there is an informal convention to stub the long form and let students shorten it if appropriate. I think the principle of minimum surprise can be useful in teaching new content.

The exception is when function syntax is the point of the exercise. This was the case with `lasagna`, and will be true again when we reach multiple-dispatch. On that point, I have some doubts about stubbing the two `loot` methods in `annalyns-infiltration2`, but that exercise still has multiple issues to deal with and is well down the concept exercise flowchart, so we can maybe save the argument for another day.

SaschaMann:
> I agree that stubs are unnecessary but harmless. They are also Exercism policy for concept exercises, so we're kind of stuck with them.

Empty stubs satisfy the requirement that they must be present and serve the purpose for making it clear where the student needs to add code. The other statements in the spec leave room for interpretation and imho empty stubs are in line with the spec. Clearly we shouldn't provide stubs for `if`s either, even if the exercise doesn't introduce them.

> I think the principle of minimum surprise can be useful in teaching new content.

Spaced repetition is one of the most effective learning methods and while the function syntax is fairly trivial, you're still taking away learning opportunities from students by removing the need to practice what has been introduced before (when to use what syntax? Should a function use a keyword arg or be two definitions? etc).

I stand by that these stubs make the exercise worse and take away learning opportunities from the student. If the spec is bad in that regard, we should ignore it (even if I think empty stubs are in line with it). It's not like leaving them empty will break the site.

SaschaMann:
> Maybe we could compromise by keeping stubs for early exercises where students may be really new to the language: documents-lasagna, annalyns-infiltration, tech-palace, and removing them for more advanced exercises (either blank or with comments). In most cases, instructions.md gives the format of each function, and runtests.jl will always show it.

I'd be fine with that though it still makes learning through spaced repetition harder as you want to repeat things very frequently early on. `documented-lasagna` is an edge case that should have the examplar solution of `lasagna` as stub anyway, regardless of its position in the track.

colinleach:
If you really prefer to avoid stubs, it's probably best to close the PR without merging. It may be worth a short comment in issue #567 so other people know.

SaschaMann:
> If you really prefer to avoid stubs, it's probably best to close the PR without merging. 

I intentionally didn't request changes to block it or close the PR. While I strongly object to the stubs, if others disagree and want it merged despite those objections, so be it.

cmcaine:
Let's leave it closed. Thanks for the effort, @colinleach!

iHiD:
Hi. I'm reopening this. As Colin (Leach) states above, stubs are part of the spec for Learning Exercises, and so should be added. 

While I appreciate you might not like them and feel it's best not to have them, but they're there for a reason and much discussion went into that decision when it was made. Exercism operates on a "Disagree and commit" basis - it's fine to disagree when the spec is being made, but if its in the spec it still needs to be followed. (Sascha - I appreciate that you want to do what's best for the student and that you didn't block the PR, but equally, I'd rather us just get on with getting the Learning Mode launched even if its not exactly how you would choose it to be if you were running Exercism.)

SaschaMann:
Fwiw, the style of stubs in #604 would seem like a sensible compromise. It doesn't suggest one way of defining a function over the other while also clearly stating what has to be defined. And it matches the example from the holy text in terms of what information is there.

ErikSchierboom:
We'd like to merge this PR as is, as providing function definitions for concept exercises is something we do consistently across all tracks and it has been found to be helpful to students.

exercises/concept.wip/exercism-matrix/exercism-matrix.jl
```diff
@@ -0,0 +1,16 @@
+E = nothing
```
cmcaine, line 1:
```suggestion
const E = nothing
```

PR #619, opened 2023-04-03
Fixes #618 Binary Search tests broken for the multiple_targets case

### LINKED ISSUE #618: Buggy tets for Binary Search with multiple targets
Binary Search sports several _optional_ tests, covering a number of features beyond the basic search. Any student can get a working practice problem without enabling them, but if they do, `multiple_targets` happens to be broken... because the multiple targets flag is not set:

```
    if get(tested_benus_tasks, :multiple_matches, false)
        @testset "multiple matches" begin
            @test binarysearch([1, 3, 4, 6, 6, 9, 11], 6) == 4:5
            @test binarysearch([1, 1, 4, 6, 8, 9, 11], 1) == 1:2
            @test binarysearch([1, 3, 4, 6, 8, 11, 11], 11) == 6:7
            @test binarysearch([1, 3, 5, 8, 13, 21, 34, 55, 144, 144, 144, 377, 634],144,) == 9:11
            @test binarysearch([1, 3, 5, 21, 21, 21, 21, 21, 89, 144, 233, 377], 21) == 4:8
        end
    end
```

rcardenes:
Pretty much self explanatory. Some of the tests don't test what they're supposed to. The fix is trivial.

### DIFF
--- a/exercises/practice/binary-search/runtests.jl
+++ b/exercises/practice/binary-search/runtests.jl
@@ -81,11 +81,11 @@ end
 
     if get(tested_benus_tasks, :multiple_matches, false)
         @testset "multiple matches" begin
-            @test binarysearch([1, 3, 4, 6, 6, 9, 11], 6) == 4:5
-            @test binarysearch([1, 1, 4, 6, 8, 9, 11], 1) == 1:2
-            @test binarysearch([1, 3, 4, 6, 8, 11, 11], 11) == 6:7
-            @test binarysearch([1, 3, 5, 8, 13, 21, 34, 55, 144, 144, 144, 377, 634],144,) == 9:11
-            @test binarysearch([1, 3, 5, 21, 21, 21, 21, 21, 89, 144, 233, 377], 21) == 4:8
+            @test binarysearch([1, 3, 4, 6, 6, 9, 11], 6, multiple_matches = true) == 4:5
+            @test binarysearch([1, 1, 4, 6, 8, 9, 11], 1, multiple_matches = true) == 1:2
+            @test binarysearch([1, 3, 4, 6, 8, 11, 11], 11, multiple_matches = true) == 6:7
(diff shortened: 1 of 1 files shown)


cmcaine:
Where do we specify that there should be a multiple_matches keyword arg?

rcardenes:
Hi. In the practice exercise itself there's a NamedTuple with some flags controlling what gets tested:

```
tested_benus_tasks = (rev = true, by = true, lt = true, multiple_matches = false)
```
By default everything is `false`, and `binarysearch` can have the following signature:

```
function binarysearch(haystack, needle)
```

But one can write a full-fledged `binarysearch` function like this:

```
function binarysearch(haystack, needle; rev=false, by=identity, lt = <, multiple_matches=false)
   ...
end
```

The problem I found is that the tests, when `multiple_matches` are enabled, do **not** pass the `multiple_matches` flag meaning that the search is done as if only one matching value was expected. But the test expects the proper extended ranges. Like so:

```
    if get(tested_benus_tasks, :multiple_matches, false)
        @testset "multiple matches" begin
            @test binarysearch([1, 3, 4, 6, 6, 9, 11], 6) == 4:5
            @test binarysearch([1, 1, 4, 6, 8, 9, 11], 1) == 1:2
            @test binarysearch([1, 3, 4, 6, 8, 11, 11], 11) == 6:7
            @test binarysearch([1, 3, 5, 8, 13, 21, 34, 55, 144, 144, 144, 377, 634],144,) == 9:11
            @test binarysearch([1, 3, 5, 21, 21, 21, 21, 21, 89, 144, 233, 377], 21) == 4:8
        end
    end
```

My fix just adds the `, multiple_matches=true` flag to the method calls, but now it fails... which makes me suspect we are supposed to _detect_ the presence of additional results.

I'll cancel the pull request, because it looks like that's the case.

PR #620, opened 2023-04-05
Convert word-count mentoring notes to approaches intro

cmcaine:
Original file:
https://github.com/exercism/website-copy/blob/main/tracks/julia/exercises/word-count/mentoring.md

I've revised the text and solutions a bit and removed one solution that works badly with the new instructions.

### DIFF
--- a/exercises/practice/word-count/.approaches/introduction.md
+++ b/exercises/practice/word-count/.approaches/introduction.md
@@ -1,11 +1,32 @@
 # Introduction
 
+## Specification
+
+The specification of this exercise has changed several times, slightly changing what counts as a word.
+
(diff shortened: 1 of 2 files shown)


exercises/practice/word-count/.approaches/introduction.md
```diff
@@ -0,0 +1,224 @@
...
+end
+```
+
+Note that this solution was written for a previous version of the instructions and counts a lot of words that the new specification does not allow.
+See if you can find some!
+
+<details>
+<summary>Answer</summary>
+
+This solution will count words containing more than one apostrophe as a single word, which is against the specification.
```
vaeng, line 117: Do you think this note is also needed in the other solutions that allow for the multiple-apostrophe behaviour? Like the very last one?

ErikSchierboom (approved):
Approving with the remark that we'd still like to see separate approaches, but this is a good starting point

PR #632, opened 2023-04-15
Remove order dependency

DrDrDrDrDrAlhazenEuler:
The test shall pass regardless of the order of the entries. This will work because doublings will already have been filtered out by the tests where there is only one solution.

### DIFF
--- a/exercises/practice/pythagorean-triplet/runtests.jl
+++ b/exercises/practice/pythagorean-triplet/runtests.jl
@@ -18,11 +18,11 @@ end
 end
 
 @testset "returns all matching triplets" begin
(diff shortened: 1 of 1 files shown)


cmcaine:
Thanks for your interest, but I'd rather keep the problem specification as it is (unique triples in ascending lexicographic order).

DrDrDrDrDrAlhazenEuler:
Hello there,

where would I find this problem specification? Certainly not in the problem instructions.

cmcaine:
You are right. In Exercism v2 (which this exercise was written for), the tests formed part of the specification. We should mention it in the instructions as well, though. We could relax the requirement, but if we did that, it would be simpler to sort the student's solution than to use Sets.

I've made a PR to change the instructions.

PR #635, opened 2023-04-21
Square Root exercise

### LINKED ISSUE #633: Adding the square-root exercise
Hi!

I'd like to add the square-root exercise to the track.

golanor:
Hi,

This is an attempt at adding the square-root exercise to Julia. This is my first PR, so please let me know (and be forgiving) if I missed something. Should close #633 

Thanks!


exercises/practice/square-root/square-root.jl
```diff
@@ -0,0 +1,3 @@
+function square_root(n::Int64; tol)
```
cmcaine, line 1: Do we need to provide a tolerance parameter? The built-in sqrt doesn't use one.

If we do provide a tolerance parameter then I think we should specify if it is an absolute tolerance (atol) or relative tolerance (rtol) to match base Julia's naming patterns.

exercises/practice/square-root/.meta/example.jl
```diff
@@ -0,0 +1,18 @@
...
+    x = initial_guess(n)
+    while abs(x^2 - n) > tol
+        x = 0.5 * (x + n / x)
+    end
+    round(x; digits=digs)
+end
+
+
+function initial_guess(n::Int64)
+    m = Int(floor(log10(n) / 2))
```
cmcaine, line 15: This isn't important, but this pattern is a bit more idiomatic, I think, and better matches your line 2.

```suggestion
    m = floor(Int, log10(n) / 2)
```

exercises/practice/square-root/.meta/config.json
```diff
@@ -0,0 +1,20 @@
+{
+  "blurb": "Given a natural radicand, return its square root.",
```
cmcaine, line 2: With less jargon?

```suggestion
  "blurb": "Calculate a positive integer's square root",
```

exercises/practice/square-root/square-root.jl
```diff
@@ -0,0 +1,3 @@
...
+    
+end
```
cmcaine, line 3: Stubs should end with a blank line, I think, to match unix expectations.

```suggestion
end

```

cmcaine (commented):
Thanks!

exercises/practice/square-root/square-root.jl
cmcaine: Here's an implementation that will converge on the exact answers for integer square numbers:

```julia
function mysqrt2(S)
    x = S
    while (S/x + x) / 2 ≠ x
        x = (S/x + x) / 2
    end
    return x
end


findall(sqrt(n^2) ≠ mysqrt2(n^2) for n in 1:1000)
```

golanor: I thought having a keyword argument would be a nice practice that is usually not done in these exercises, but that's not a must.

cmcaine, line 1:
```suggestion
function square_root(n)
```

exercises/practice/square-root/runtests.jl
```diff
@@ -0,0 +1,35 @@
...
+    @testset "Domain error for negative number" begin
+        @test_throws DomainError square_root(-65536)
+    end
+end
+
+
+@testset "Random tests" begin
+    for s in rand(10, 1:1000000)
+        @test square_root(s^2) == s
+    end
```
cmcaine, line 34:
```suggestion
    for _ in 1:1000
        s = rand(1:isqrt(typemax(Int64)))
        @test square_root(s^2) == s
    end
```

exercises/practice/square-root/square-root.jl
cmcaine: Cool, well I think it's better to not have it in this exercise. Thanks for removing it!

cmcaine (approved):

exercises/practice/square-root/.meta/example.jl
```diff
@@ -0,0 +1,17 @@
...
+    end
+    x
+end
+
+
+function initial_guess(n::Int64)
+    m = floor(Int, log10(n) / 2)
+    a = n ÷ (10 ^ m)
+    (a < 10 ? (0.89 + 0.28*a) : (2.8 + 0.089*a)) * 10^m
+end
```
cmcaine, line 17: This equation is a little messed up, btw:

```suggestion
function initial_guess(n::Int64)
    m = floor(Int, log10(n))
    a = n / (10 ^ m)
    (a < 10 ? (0.89 + 0.28*a) : (2.8 + 0.089*a)) * 10^(m ÷ 2)
end
```

PR #647, opened 2023-06-22
Make rational-numbers exercise easier

### LINKED ISSUE #646: rational-numbers Ordering testset too hard
To correctly solve the exercise you need to handle positive and negative infinities. We could remove that requirement by changing the "Ordering" testset to iterate `1:5` instead of `-5:5`.

The "Ordering" testset also needs the user to define `<(::RationalNumber, ::Int)`, etc, or to define some promotion_rules correctly, which we don't explain in the readme.

We should probably relegate the Ordering testset to bonus tasks and split it up so that the student doesn't need to learn about handling infinities at the same time as they try to get conversion and promotion/comparison functions working.

cmcaine:
These revisions make the rational numbers exercise easier by:

- introducing more tests (to guide the student to good solutions)
- making instructions more explicit
- removing tests that required correct handling of infinities (not specified in instructions)
- removing tests that required a custom `show` method (just not a good idea).

TODO: add an approach.

Closes #646

### DIFF
--- a/exercises/practice/rational-numbers/runtests.jl
+++ b/exercises/practice/rational-numbers/runtests.jl
@@ -122,7 +122,7 @@ end
 # https://github.com/JuliaLang/julia/blob/52bafeb981bac548afd2264edb518d8d86944dca/LICENSE.md
 @testset "Ordering" begin
     for a in -5:5, b in -5:5, c in -5:5
-        a == b == 0 && continue
+        b == 0 && continue
         
         r = RationalNumber(a, b)
 
@@ -134,7 +134,7 @@ end
         @test (r >  c) == (a / b >  c)
 
         for d in -5:5
-            c == d == 0 && continue
+            d == 0 && continue
 
             s = RationalNumber(c, d)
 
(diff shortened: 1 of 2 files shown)


exercises/practice/rational-numbers/runtests.jl
```diff
@@ -5,6 +5,18 @@ include("rational-numbers.jl")
...
+@testset "Conversion and promotion" begin
+    # To pass these tests you should define a promote_rule method and a
+    # constructor, you don't need to define methods such as
+    # `==(a::RationalNumber{Int}, b::Int)`, `promote_type`, `convert`, etc.
+    #
+    # Read the manual for how to do this and why:
+    # https://docs.julialang.org/en/v1/manual/conversion-and-promotion/
+    @test RationalNumber{Int}(10) == 10
+    @test promote_type(Int, RationalNumber{Int}) == RationalNumber{Int}
+    @test convert(RationalNumber{Int8}, 2) <: RationalNumber{Int8}
```
cmcaine, line 17:
```suggestion
    @test convert(RationalNumber{Int8}, 2) isa RationalNumber{Int8}
```

ErikSchierboom (approved):

PR #661, opened 2023-08-28
Update runtests.jl

DrDrDrDrDrAlhazenEuler:
Fixed some flawed numerical test values.

### DIFF
--- a/exercises/practice/luhn/runtests.jl
+++ b/exercises/practice/luhn/runtests.jl
@@ -54,11 +54,11 @@ end
 
 @testset "valid" begin
     @testset "valid Canadian SIN" begin
-        @test luhn("055 444 285")
+        @test luhn("055 444 283")
     end
 
     @testset "a simple valid SIN that remains valid if reversed" begin
-        @test luhn("059")
+        @test luhn("59")
     end
 
     @testset "a simple valid SIN that becomes invalid if reversed" begin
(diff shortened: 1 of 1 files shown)


BNAndras:
I'm not a maintainer of the Julia track so this is just a friendly comment, but those test values come from https://github.com/exercism/problem-specifications which contains a shared set of markdown files and test data for all Exercism tracks. Tracks don't need to use that data, but it's strongly encouraged since maintaining a track is a lot simpler if you don't also need to maintain your bespoke test cases. If you believe these tests should be updated, I'd suggest proposing your changes at @exercism/problem-specifications since then multiple tracks could benefit instead of a single track.

DrDrDrDrDrAlhazenEuler:
> I'm not a maintainer of the Julia track so this is just a friendly comment, but those test values come from https://github.com/exercism/problem-specifications which contains a shared set of markdown files and test data for all Exercism tracks. Tracks don't need to use that data, but it's strongly encouraged since maintaining a track is a lot simpler if you don't also need to maintain your bespoke test cases. If you believe these tests should be updated, I'd suggest proposing your changes at @exercism/problem-specifications since then multiple tracks could benefit instead of a single track.

If I propose the change over there, should I then close this pull request? Is it ensured that the changes then apply to the Julia track also?

DrDrDrDrDrAlhazenEuler:
I officially just failed to create a pull request to the folder you mentioned because apparently right now the company does not accept contributions from non-employees.

PR #662, opened 2023-08-28
Sync exercise markdown files and metadata

BNAndras:


BNAndras:
@exercism/julia, can I get a review? I pulled down the latest docs and metadata.

cmcaine:
Can you provide the exact code that you ran to do the sync so that I can confirm I get the same output without having to eyeball that huge diff?

exercises/practice/alphametics/.docs/instructions.md
```diff
@@ -28,5 +28,4 @@ Each letter must represent a different digit, and the leading digit of a multi-d
 
 Write a function to solve alphametics puzzles.
 
-You may (or may not!) want to call the function `permutations(a, t)` from [Combinatorics.jl](https://github.com/JuliaMath/Combinatorics.jl) in your solution.
-If you would like to use it, you can include the **read-only** file `permutations.jl` with `include("permutations.jl")`.
```
cmcaine, line 32: Julia-specific changes like this should generally be retained.

exercises/practice/bob/.docs/instructions.md
```diff
@@ -1,16 +1,19 @@
...
-
-He says 'Fine. Be that way!' if you address him without actually saying
-anything.
-
-He answers 'Whatever.' to anything else.
-
-Bob's conversational partner is a purist when it comes to written communication and always follows normal rules regarding sentence punctuation in English.
+- **"Sure."**
+  This is his response if you ask him a question, such as "How are you?"
+  The convention used for questions is that it ends with a question mark.
```
cmcaine, line 9: This has poor grammar (questions is plural => "they" rather than "it"). Please consider opening a PR on the problem-spec repo.

```suggestion
Questions end with a question mark.
```

exercises/practice/isbn-verifier/.docs/instructions.md
```diff
@@ -1,54 +1,42 @@
...
-
-* Define a string macro so that `isbn"3-598-21507-X" == ISBN("3-598-21507-X")`.
-
-* Let the constructor accept ISBN-13s as well. The same ISBN should compare equal regardless of format: `ISBN("1-234-56789-X") == ISBN("978-1-234-56789-7")`.
-
-* Define a function or iterator to generate a valid ISBN, maybe even from a given starting ISBN.
+Now, it's even trickier since the check digit of an ISBN-10 may be 'X' (representing '10').
+For instance `3-598-21507-X` is a valid ISBN-10.
 
-* Can you store the ISBN as an integer rather than a string and would that ever be a good idea? If you did, could you still print the ISBN (including check-digit) as a string?
```
cmcaine, line 54: Shame to lose these bits. Please try adding them to an append file or similar.

exercises/practice/phone-number/.docs/instructions.md
```diff
@@ -27,10 +32,3 @@ should all produce the output
 `6139950253`
 
 **Note:** As this exercise only deals with telephone numbers used in NANP-countries, only 1 is considered a valid country code.
-
-When an input is invalid your function must throw an `ArgumentError`.
-Invalid inputs include:
-- `apple`
-- `100`
-- `2 555 555 1234`
-- `100 200 3000`
```
cmcaine, line 36: This needs to go in an append file. Tests expect that error, iirc.

exercises/practice/rational-numbers/.docs/instructions.md
```diff
@@ -28,16 +28,15 @@ Exponentiation of a real number `x` to a rational number `r = a/b` is `x^(a/b) =
...
-
-In Julia, there are fallback methods already defined for many arithmetic methods, comparisons, etc, especially for subtypes of `Real`.
-You do not need to write your own method for `>=` if you have already defined `<`; nor do you need to define `^(::RationalNumber, ::Integer)` if you have provided a method for multiplication; etc.
-You can examine the fallback methods for a given function by entering `methods(abs)` or `edit(abs, (RationalNumber, RationalNumber))` at the REPL.
-
-Your implementation of rational numbers should always be reduced to lowest terms. For example, `4/4` should reduce to `1/1`, `30/60` should reduce to `1/2`, `12/8` should reduce to `3/2`, etc. To reduce a rational number `r = a/b`, divide `a` and `b` by the greatest common divisor (gcd) of `a` and `b`. So, for example, `gcd(12, 8) = 4`, so `r = 12/8` can be reduced to `(12/4)/(8/4) = 3/2`.
-The reduced form of a rational number should be in "standard form" (the denominator should always be a positive integer). If a denominator with a negative integer is present, multiply both numerator and denominator by `-1` to ensure standard form is reached. For example, `3/-4` should be reduced to `-3/4`
```
cmcaine, line 41: There's Julia-specfic stuff in here that should be retained. Without these hints the exercise is harder and students will probably learn less.

cmcaine (changes requested):
Please retain the Julia-specific changes. Thanks!

exercises/practice/alphametics/.docs/instructions.md
BNAndras: Yeah, those should be in an append for ease of maintenance.

BNAndras:
> Can you provide the exact code that you ran to do the sync so that I can confirm I get the same output without having to eyeball that huge diff?

`configlet sync —docs —metadata -uy` from what I recall off hand.

exercises/practice/rational-numbers/.docs/instructions.md
cmcaine: The first set of bullet points also contains content that should be retained:

```
- conversion and promotion of integers to rationals. Read the manual's chaper on [conversion and promotion](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/) for details about how and why to do this.
- comparison (less than, equals, etc.) of two rational numbers, comparison of a rational number and an integer
```

IMO, it is important to keep hinting to the student that conversion and promotion is the first thing they should do. If you want to move this into the .append file then please find some way of making it more prominent.

Otherwise, maybe just revert the changes to the files in this directory.

cmcaine (changes requested):
Please revert the changes to rational-number, or merge in the Julia-specific changes without a .append or find a good way of capturing all the Julia-specific stuff in the .append.

exercises/practice/rational-numbers/.docs/instructions.md
BNAndras: I'd rather move the Julia-specific information to an append as long as we're not duplicating information in the instructions. That's the suggested workflow in the Exercism docs and I don't see a good reason to deviate. Editing the instructions in place just delays the issue.

cmcaine: I appreciate your work in revising this PR. Are you saying that you intend to move those two lines into a .append file?

As I say below, I think that's fine so long as the advice to start with conversion and promotion is made prominent (ideally even more prominent than it is now).

Re: reverting, you don't see a good reason to deviate from the suggested workflow.

The rational-numbers exercise, like several others on the Julia track, has a bunch of intentional differences from the upstream definitions in problem-specs. I think the julia-specific changes are good and that if we remove them it should be because they are bad for students or mentors, not because they are inconsistent with problem-specifications.

What is the cost of letting rational-numbers continue to differ from the problem-spec version? I don't work on this repo very much any more, is there a developer experience issue that I'm not seeing?

BNAndras:
For rational-numbers, the append currently contains the Julia-specific content discussing fallback methods. I'm not seeing anything else that needs to be captured in the append.

cmcaine:
The append does not contain the extra bullet points that go under "implement the following...": https://github.com/exercism/julia/pull/662/files#r1470022604

Please either revert the changes to the rational numbers folder (contributors can just deal with configlet sync warning them about rational numbers) or find a way to include that information in a natural and high-profile way.

BNAndras:
I reverted the sync for the rational-numbers documentation.

cmcaine:
Thanks again for this contribution.

Because this is a force push, I can't easily see what changed in the commit, but I'm just going to trust you on it and merge it.

In future, please help me out by avoiding force pushes or rebases and making separate commits when responding to review.

cmcaine (approved):

PR #678, opened 2023-12-10
New practice exercise - Perceptron

depial:
This is a first attempt at contributing an exercise, so there may be some deficiencies. I haven't received a reply to the issue I opened, so I will leave this as a PR in case it's of interest to Exercism.

- Since this is my first contribution, I would appreciate any help in bettering the exercise so I can contribute more efficiently in the future.
- I was unable to get configlet to automatically create the UUIDs for the tests, so I've done it manually (using the UUID function).
- The test setup uses pseudo-random dataset generation. I haven't seen this technique in other Exercism exercises, so I don't know if this creates a the above problem.

### DIFF
--- a/exercises/practice/perceptron/.docs/instructions.md
+++ b/exercises/practice/perceptron/.docs/instructions.md
@@ -1,0 +1,34 @@
+# Instructions
+
+### Introduction
+[Perceptron](https://en.wikipedia.org/wiki/Perceptron) is one of the oldest and bestestly named machine learning algorithms out there. Since perceptron is also quite simple to implement, it's a favorite place to start a machine learning journey. As a linear classifier, if a linear decision boundary (e.g. a line in 2D or hyperplane in general) can be drawn to separate two labled classes of objects, perceptron is guaranteed to do so. This can help in predicting what an unlabeled object would likely be classified as by seeing which side of the decision boundary it is on.
+
+### Details
+The basic idea is fairly straightforward. We cycle through the objects and check if they are on the correct side of our hyperplane. If one is not, we make a correction to the hyperplane and continue checking the objects against the new hyperplane. Eventually the hyperplane is adjusted to correctly separate all the objects and we have our decision boundary!
+
+#### A Brief Word on Hyperplanes
+How do you pick your starting hyperplane? It's up to you! Be creative! Or not... Actually perceptron's convergence times are sensitive to conditions such as the initial hyperplane and even the order the objects are looped through, so you might not want to go too wild.
+
+We will be dealing with a two dimensional space, so our divider will be a line. The standard equation for a line is usually written as $y = ax+b$, where $a,b \in \Re$, however, in order to help generalize the idea to higher dimensions, it's convenient to reformulate this equation as $w_0 + w_1x + w_2y = 0$. This is the form of the [hyperplane](https://en.wikipedia.org/wiki/Hyperplane) we will be using, so your output should be $[w_0, w_1, w_2]$. In machine learning, ${w_0,w_1,w_2}$ are usually referred to as weights. 
+
+While hyperplanes are equivalent under scalar multiplication, there is a difference between $[w_0, w_1, w_2]$ and $[-w_0, -w_1, -w_2]$ in that the normal to the hyperplane points in opposite directions. By convention, the perceptron normal points towards the class defined as positive, so this property will be checked but not result in a test failure.
+
+#### Updating
+Checking if an object is on one side of a hyperplane or another can be done by checking the normal vector which points to the object. The value will be positive, negative or zero, so all of the objects from a class should have normal vectors with the same sign. A zero value means the object is on the hyperplane, which we don't want to allow since its ambiguous. Checking the sign of a normal to a hyperplane might sound like it could be complicated, but it's actually quite easy. Simply plug in the coordinates for the object into the equation for the hyperplane and check the sign of the result. For example, we can look at two objects $v_1,v_2$ in relation to the hyperplane $[w_0, w_1, w_2] = [1, 1, 1]$:
+
+$$v_1$$ $$[x_1, y_1] = [2, 2]$$ $$w_0 + w_1*x_1 + w_2*y_1 = 1 + 1*2 + 1*2 = 5 > 0$$
+
+
+$$v_2$$ $$[x_2,y_2]=[-2,-2]$$ $$w_0 + w_1*x_2 + w_2*y_2 = 1 + 1*(-2) + 1*(-2) = -3 < 0$$
+
+If $v_1$ and $v_2$ have different labels, such as $1$ and $-1$ (like we will be using), then the hyperplane $[1, 1, 1]$ is a valid decision boundary for them. 
+
+Now that we know how to tell which side of the hyperplane an object lies on, we can look at how perceptron updates a hyperplane. If an object is on the correct side of the hyperplane, no update is performed on the weights. However, if we find an object on the wrong side, the update rule for the weights is:
+
+$$[w_0', w_1', w_2'] = [w_0 \pm l_{class}, w_1 \pm x*l_{class}, w_2 \pm y*l_{class}]$$
+
+Where $l_{class}=\pm 1$, according to the class of the object (i.e. its label), $x,y$ are the coordinates of the object, the $w_i$ are the weights of the hyperplane and the $w_i'$ are the weights of the updated hyperplane. The plus or minus signs are homogenous, so either all plus or all minus, and are determined by the choice of which class you define to be on the positive side of the hyperplane. Beware that only two out of the four possible combinations of class on positive side of the hyperplane and the plus/minus in the update are valid ($\pm \pm, \mp \mp$), with the other two ($\pm \mp, \mp \pm$) leading to infinite loops.
+
+This update is repeated for each object in turn, and then the whole process repeated until there are no updates made to the hyperplane. All objects passing without an update means they have been successfully separated and you can return your decision boundary!
+
+Note: Although the perceptron algorithm is deterministic, a decision boundary depends on initialization and is not unique in general, so the tests accept any hyperplane which fully separates the objects.
--- a/exercises/practice/perceptron/.meta/config.json
+++ b/exercises/practice/perceptron/.meta/config.json
@@ -1,0 +1,20 @@
+{
+  "authors": [
+    "depial"
+  ],
+  "contributors": [
+    "cmcaine"
+  ],
+  "files": {
+    "solution": [
+      "perceptron.jl"
+    ],
+    "test": [
+      "runtests.jl"
+    ],
+    "example": [
+      ".meta/example.jl"
+    ]
+  },
+  "blurb": "Given points and their labels, provide a hyperplane which separates them"
+}
--- a/exercises/practice/perceptron/.meta/example.jl
+++ b/exercises/practice/perceptron/.meta/example.jl
@@ -1,0 +1,8 @@
+function perceptron(points, labels)
+    θ, pnts = [0, 0, 0], vcat.(1, points)
+    while true
+        θ_0 = θ
+        foreach(i -> labels[i]*θ'*pnts[i] ≤ 0 && (θ += labels[i]*pnts[i]), eachindex(pnts))
+        θ_0 == θ && return θ
+    end
+end
--- a/exercises/practice/perceptron/.meta/tests.toml
+++ b/exercises/practice/perceptron/.meta/tests.toml
@@ -1,0 +1,25 @@
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
+
+[728853d3-24de-4855-a452-6520b67dec23]
+description = "Initial set"
+
+[ed5bf871-3923-47ca-8346-5d640f9069a0]
+description = "Initial set w/ opposite labels"
+
+[15a9860e-f9be-46b1-86b2-989bd878c8a5]
+description = "Hyperplane cannot pass through origin"
+
+[52ba77fc-8983-4429-91dc-e64b2f625484]
+description = "Hyperplane nearly parallel with y-axis"
+
+[3e758bbd-5f72-447d-999f-cfa60b27bc26]
+description = "Increasing Populations"
--- a/exercises/practice/perceptron/perceptron.jl
+++ b/exercises/practice/perceptron/perceptron.jl
@@ -1,0 +1,3 @@
+function perceptron(points, labels)
+    # Perceptronize!
+end
--- a/exercises/practice/perceptron/runtests.jl
+++ b/exercises/practice/perceptron/runtests.jl
@@ -1,0 +1,43 @@
+using Test
+
+include("perceptron.jl")
+include("testtools.jl")
+
+@testset "Low population" begin
+    @testset "Initial set" begin
+        points = [[1, 2], [3, 4], [-1, -2], [-3, -4], [2, 1], [1, 1]]
+        labels = [1, 1, -1, -1, 1, 1]
+        reference = [1, 2, 1]
+        hyperplane = perceptron(points, labels)
+        @test dotest(points, labels, hyperplane, reference)
+    end
+    @testset "Initial set w/ opposite labels" begin
+        points = [[1, 2], [3, 4], [-1, -2], [-3, -4], [2, 1], [1, 1]]
+        labels = [-1, -1, 1, 1, -1, -1]
+        reference = [-1, -2, -1]
+        hyperplane = perceptron(points, labels)
+        @test dotest(points, labels, hyperplane, reference)
+    end
+    @testset "Hyperplane cannot pass through origin" begin
+        points = [[1, 2], [3, 4], [-1, -2], [-3, -4], [2, 1], [-1, -1]]
+        labels = [1, 1, -1, -1, 1, 1]
+        reference = [-1, 3, 3]
+        hyperplane = perceptron(points, labels)
+        @test dotest(points, labels, hyperplane, reference)
+    end
+    @testset "Hyperplane nearly parallel with y-axis" begin
+        points = [[0, 50], [0, -50], [-2, 0], [1, 50], [1, -50], [2, 0]]
+        labels = [-1, -1, -1, 1, 1, 1]
+        reference = [2, 0, -1]
+        hyperplane = perceptron(points, labels)
+        @test dotest(points, labels, hyperplane, reference)
+    end
+end
+
+@testset "Increasing Populations" begin
+    for n in 10:50
+        points, labels, reference = population(n, 25)
+        hyperplane = perceptron(points, labels)
+        @test dotest(points, labels, hyperplane, reference)
+    end
+end
--- a/config.json
+++ b/config.json
@@ -883,6 +883,21 @@
           "conditionals",
           "variables"
         ]
+      },
+      {
+        "uuid": "b43a938a-7bd2-4fe4-b16c-731e2e25e747",
+        "practices": [],
+        "prerequisites": [],
+        "slug": "perceptron",
+        "name": "Perceptron",
+        "difficulty": 3,
+        "topics": [
+          "machine learning",
+          "loops",
+          "arrays",
+          "logic",
+          "math"
+        ]
       }
     ]
   },
--- a/exercises/practice/perceptron/testtools.jl
+++ b/exercises/practice/perceptron/testtools.jl
@@ -1,0 +1,43 @@
+using Random
+
+function dotest(points, labels, hyperplane, reference)
+    # Tests if a hyperplane linearly separates labeled points
+    # Returns true or false
+
+    points = vcat.(1, points)
+    test = reduce(hcat, points)' * hyperplane .* labels
+    if all(>(0), test)
+        println("Reference hyperplane = $reference\nYour hyperplane = $hyperplane\nSeparated! And the normal points towards the positively labeled side\n")
+        return true
+    elseif all(<(0), test)
+        println("Reference hyperplane = $reference\nYour hyperplane = $hyperplane\nSeparated! But the normal points towards the negatively labeled side\n")
+        return true
+    else
+        println("Reference hyperplane = $reference\nYour hyperplane = $hyperplane\nThe sides are not properly separated...\n")
+        return false
+    end
+end
+
+Random.seed!(42) # set seed for deterministic test set
+
+function population(n, bound)
+    # Builds a population of n points with labels {1, -1} in area bound x bound around a reference hyperplane
+    # Returns linearly separable points, labels and reference hyperplane
+
+    vertical = !iszero(n % 10) #every tenth test has vertical reference hyperplane
+    x, y, b = rand(-bound:bound), rand(-bound:bound)*vertical, rand(-bound÷2:bound÷2)
+    y_intercept = -b ÷ (iszero(y) ? 1 : y)
+    points, labels, hyperplane = [], [], [b, x, y]
+    while n > 0
+        # points are centered on y-intercept, but not x-intercept so distributions can be lopsided
+        point = [rand(-bound:bound), y_intercept + rand(-bound:bound)]
+        label = point' * [x, y] + b
+        if !iszero(label)
+            push!(points, point)
+            push!(labels, sign(label))
+            n -= 1
+        end
+    end
+
+    points, labels, hyperplane
+end


depial:
It looks like a lot of checks failed, so I apologize since I'm not sure what happened.

yctai1994:
Would you mind explaining what you intend to do in your commit? Particularly for the test cases, it would be better to provide any math explanation. If so, I can try to help with your pull request after next week.

depial:
TLDR: I have experience creating coding exercises, but, while I've done my best to read the Exercism docs, I still have quite limited experience with GitHub and the Exercism contribution procedure, so that is likely where I'll need the most help.

Excuse me if I misunderstand your question, since I'm new in this process, but I'll give a bit more mathematical description of the functions I use in creating the test cases and testing the returned values.

In the test cases, there are two types: fixed and pseudorandom. The fixed cases are just an array of six points and an array of labels corresponding to the points' class. The pseudorandom tests use the function `population` in `testtools.jl` to create linearly separable arrays of many more points and labels, but a seed is used, so the same test set is returned every time for reliability. Effectively, the function first creates a line (pseudorandomly), then uses it as a reference to create the separated populations of the two classes to be passed to the student. Care has been taken to keep the populations close to the origin, while allowing for some lopsidedness (more of one class than the other). If you would like more technical mathematical description, please let me know.

The task, put simply, is to return a line (in hyperplane format) which can completely separate the points with one class on one side and the other class on the other side. The other function in `testtools.jl`, `dotest`, tests the hyperplane returned by the student via an elementwise dot product and scalar multiplication which can determine on which side of the hyperplane the points are. If $h = hyperplane$, $p=point$ and $l = label$, then the equation to see if the point is on the "correct" side is $$(h \cdot p) l>0$$ However, a hyperplane still separates the points if $(h \cdot p) l<0$ **for all points**. However, this isn't the convention in ML, so this is tested, but does not result in a test failure. For more mathematical detail on these aspects, please see `instructions.md`

That being said, I believe there should be no problem with the mechanics/mathematics of the exercise, since I've tested it quite thoroughly a couple of places. I figure there might be an issue with configuration, since this is my first contribution here and I've likely messed something up. For example, I'm not sure if Exercism supports this type of pseudorandom test generation since I've only seen the fixed test type which usually get a separate UUID for each one in `tests.toml` (meanwhile the whole pseudorandom group of 40 tests gets a singe UUID in my test setup). One other thing is that there could perhaps be version issues, since I wrote this in Julia 1.8 and I see 1.6 and 1 in the failed checks above. While I didn't have any issues when testing with 1.5, I can no longer be sure I haven't made any version specific changes since those tests. 

Please feel free to ask any further clarifications.

depial:
After looking over the errors more, it looks like `runtests.jl` fails to include `testtools.jl` and the test runner check fails with error message: `"LoadError: SystemError: opening file \".//testtools.jl\": No such file or directory...`. 

I've seen extra files added to the two normal files (`slug.jl` and `runtests.jl`) such as in `Alphametics` which has the `permutations.jl` file added. However, `permutations.jl` is not called in the test runner (it's just meant to be accessed by students), so I figure the issue is if including an 'outside' file from the test runner is possible. 

In other words, is it possible to get `testtools.jl` to be visible to `runtests.jl` in the test runner?

cmcaine:
I pushed a merge commit so we can see how CI likes it when running on a newer version of the repo. If that is inconvenient for you, please let me know and I won't do it again :)

depial:
Thanks Colin! It looks like there's still an issue with testing. However, I think it's a different (but possibly related) error from what I saw before. The `LoadError` for `testtools.jl` is gone and now I'm seeing `UndefVarError: dotest not defined`, where `dotest` is a helper function in `testtools.jl`

I didn't want to put the testing function `dotest` directly into `runtests.jl` because it has code which could be copied into a solution. That's why, instead of having all the testing helper functions in `runtests.jl`, I created `testtools.jl` in the same folder hoping the functions could be called from `runtests.jl`. I'm not sure if this is possible or if all functions have to be defined explicitly in `runtests.jl`.

cmcaine:
You're welcome :)

The error is probably because we override `include()` when running the tests for the whole track at once.

I'm not sure there is a good way to hide the test code from the student. I'd just put any spoilery functions at the bottom of the test file with a note saying "Reading past here will give you the answer, so watch out!"

cmcaine:
The main motivation is that you want testtools.jl not to appear in the web UI, right?

I'm not sure that's practical while also having it be downloaded for people who use the exercism CLI.

depial:
> I'm not sure that's practical while also having it be downloaded for people who use the exercism CLI.

Oh, good point! I hadn't considered compatibility with the CLI. Cheers!

I've put the test set in a function and called it at the end of the file, as you suggested, and it looks like all the tests have passed. I've also deleted the `testtools.jl` file since it's superfluous now.

@cmcaine One other thing I would like to ask about is attribution. I had put my name as author and yours as contributor just to not leave those fields blank, but is it okay to leave them blank? I don't really care to be listed as an author and I don't know if you want to be credited as a contributor or not.

cmcaine:
Cool, thanks for fixing these things up :) I'll read through it properly in the next couple of days!

I don't know what happens if no authors are listed. If you strongly prefer not to be listed as the author on the website then you could try to credit `@JuliaLang` or `@exercism`. I don't know if that would work, though.

depial:
Thanks for the help! :)

I'm quite flexible on the author situation, so if it's any hassle at all, I certainly don't mind going with the norm. Feel free to modify/leave that (or anything else) as you see fit!

exercises/practice/perceptron/.meta/tests.toml
```diff
@@ -0,0 +1,25 @@
+# This is an auto-generated file.
```
cmcaine, line 1: Is it? I don't see a perceptron exercise in problem-specifications.

exercises/practice/perceptron/runtests.jl
```diff
@@ -0,0 +1,86 @@
+using Test, Random
+include("perceptron.jl")
+
+function runtestset()
```
cmcaine, line 4: I think we should include some tests with manually specified input data of just a few points to make this more approachable and also as good practice (e.g. first few tests could be spaces with just 2-4 points placed manually).

exercises/practice/perceptron/.docs/instructions.md
```diff
@@ -0,0 +1,34 @@
...
+
+#### A Brief Word on Hyperplanes
+How do you pick your starting hyperplane? It's up to you! Be creative! Or not... Actually perceptron's convergence times are sensitive to conditions such as the initial hyperplane and even the order the objects are looped through, so you might not want to go too wild.
+
+We will be dealing with a two dimensional space, so our divider will be a line. The standard equation for a line is usually written as $y = ax+b$, where $a,b \in \Re$, however, in order to help generalize the idea to higher dimensions, it's convenient to reformulate this equation as $w_0 + w_1x + w_2y = 0$. This is the form of the [hyperplane](https://en.wikipedia.org/wiki/Hyperplane) we will be using, so your output should be $[w_0, w_1, w_2]$. In machine learning, ${w_0,w_1,w_2}$ are usually referred to as weights. 
+
+While hyperplanes are equivalent under scalar multiplication, there is a difference between $[w_0, w_1, w_2]$ and $[-w_0, -w_1, -w_2]$ in that the normal to the hyperplane points in opposite directions. By convention, the perceptron normal points towards the class defined as positive, so this property will be checked but not result in a test failure.
```
cmcaine, line 14: I think this should maybe be a test failure. It would simplify the messaging for the student in the instruction and from the tests. Do you think that would be okay?

cmcaine: Also, negative numbers are scalars (at least to me?), so I would probably drop the first part of the first sentence.

exercises/practice/perceptron/.docs/instructions.md
```diff
@@ -0,0 +1,34 @@
...
+$$v_1$$ $$[x_1, y_1] = [2, 2]$$ $$w_0 + w_1*x_1 + w_2*y_1 = 1 + 1*2 + 1*2 = 5 > 0$$
+
+
+$$v_2$$ $$[x_2,y_2]=[-2,-2]$$ $$w_0 + w_1*x_2 + w_2*y_2 = 1 + 1*(-2) + 1*(-2) = -3 < 0$$
+
+If $v_1$ and $v_2$ have different labels, such as $1$ and $-1$ (like we will be using), then the hyperplane $[1, 1, 1]$ is a valid decision boundary for them. 
+
+Now that we know how to tell which side of the hyperplane an object lies on, we can look at how perceptron updates a hyperplane. If an object is on the correct side of the hyperplane, no update is performed on the weights. However, if we find an object on the wrong side, the update rule for the weights is:
+
+$$[w_0', w_1', w_2'] = [w_0 \pm l_{class}, w_1 \pm x*l_{class}, w_2 \pm y*l_{class}]$$
```
cmcaine, line 28: This equation and the next two paragraphs are not clear enough, imo.

The update rule is where the magic happens, so it would be nice to explain this better and try to give the student some insight into how the rule works.

The $$\pm\mp$$ will also be impenetrable to most readers, I think.

exercises/practice/perceptron/runtests.jl
depial: Sorry, I'm not sure if I've understood, but there are four tests with (six) manually specified points to illustrate a couple of different possible orientations of a hyperplane (testset "Low population"). After that the 40 pseudorandomly generated tests begin (testset "Increasing Populations"). Was this what you meant?

exercises/practice/perceptron/.meta/tests.toml
depial: I'm not sure if I fixed this or not. I've deleted the auto-generated comment. I had to manually create the UUID since when I ran configlet, it kept saying everything was done even when there were no UUIDs in this file (sorry I can't remember the exact message). Let me know if I need to do something else

exercises/practice/perceptron/runtests.jl
cmcaine: I meant we could take some of the manually specified examples out of the runtestset() function and test them without the support functions, just so that it's less mystifying to a student reading the tests.

And maybe we should have the student write their own function for finding the computed label of a point?

e.g.

```jl
# Student must implement both `perceptron(points, labels)` and `classify(boundary, point)`

# Originally provided points should be classified correctly
boundary = perceptron([[0,0], [3, 3]], [1, -1])
@test classify(boundary, [0, 0]) == 1
@test classify(boundary, [3, 3]) == -1

# Adding more points constrains the location of the boundary so that we can test
# the classification of unseen points.
boundary = perceptron([[0,0], [1, 0], [0, 1]], [-1, 1, 1])
@test classify(boundary, [0, 0]) == -1
@test classify(boundary, [2, 5]) == 1
```

exercises/practice/perceptron/.meta/tests.toml
cmcaine: Sorry, I'm not very familiar with what configlet needs when we're writing a practice exercise from scratch (one where there is not an upstream definition in problem-specifications). Someone in @exercism/reviewers might know.

exercises/practice/perceptron/.docs/instructions.md
depial: >I think this should maybe be a test failure. It would simplify the messaging for the student in the instruction and from the tests. Do you think that would be okay?

I didn't want to strictly penalize a flipped normal since it is just a convention (logical though it may be) and it would at still tell the student they have successfully found a decision boundary. In any case, I think it would be nice to leave this in, but if you think it's better to enforce the convention (which of course is good practice), we can turn this into a failed test, with it's own error message.

>Also, negative numbers are scalars (at least to me?), so I would probably drop the first part of the first sentence.

I've rephrase the scalar multiplication part to make more clear what I was trying to say (that a hyperplane looks the same multiplied by any scalar, but negative scalars flip what we define to be the normal). Hopefully that's clearer :)

exercises/practice/perceptron/.docs/instructions.md
```diff
@@ -0,0 +1,34 @@
+# Instructions
+
+### Introduction
+[Perceptron](https://en.wikipedia.org/wiki/Perceptron) is one of the oldest and bestestly named machine learning algorithms out there. Since perceptron is also quite simple to implement, it's a favorite place to start a machine learning journey. As a linear classifier, if a linear decision boundary (e.g. a line in 2D or hyperplane in general) can be drawn to separate two labled classes of objects, perceptron is guaranteed to do so. This can help in predicting what an unlabeled object would likely be classified as by seeing which side of the decision boundary it is on.
```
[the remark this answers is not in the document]
depial: I've tried to clarify and add a gif. Let me know if that is better :)

exercises/practice/perceptron/.docs/instructions.md
```diff
@@ -0,0 +1,34 @@
...
+
+While hyperplanes are equivalent under scalar multiplication, there is a difference between $[w_0, w_1, w_2]$ and $[-w_0, -w_1, -w_2]$ in that the normal to the hyperplane points in opposite directions. By convention, the perceptron normal points towards the class defined as positive, so this property will be checked but not result in a test failure.
+
+#### Updating
+Checking if an object is on one side of a hyperplane or another can be done by checking the normal vector which points to the object. The value will be positive, negative or zero, so all of the objects from a class should have normal vectors with the same sign. A zero value means the object is on the hyperplane, which we don't want to allow since its ambiguous. Checking the sign of a normal to a hyperplane might sound like it could be complicated, but it's actually quite easy. Simply plug in the coordinates for the object into the equation for the hyperplane and check the sign of the result. For example, we can look at two objects $v_1,v_2$ in relation to the hyperplane $[w_0, w_1, w_2] = [1, 1, 1]$:
+
+$$v_1$$ $$[x_1, y_1] = [2, 2]$$ $$w_0 + w_1*x_1 + w_2*y_1 = 1 + 1*2 + 1*2 = 5 > 0$$
```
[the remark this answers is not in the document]
depial: I think the problem was with no spaces between factors and `*`. I've changed this up and I think it should be rendering fine on Github now :)

exercises/practice/perceptron/.docs/instructions.md
cmcaine: I think if we ask the student to define a classify(boundary, point) function then they'll have to get consistent about the boundary anyway?

See https://github.com/exercism/julia/pull/678#discussion_r1477341834

exercises/practice/perceptron/.meta/config.json
```diff
@@ -0,0 +1,20 @@
...
+      "perceptron.jl"
+    ],
+    "test": [
+      "runtests.jl"
+    ],
+    "example": [
+      ".meta/example.jl"
+    ]
+  },
+  "blurb": "Given points and their labels, provide a hyperplane which separates them"
```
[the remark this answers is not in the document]
depial: Done! That does sound more appealing :)

exercises/practice/perceptron/runtests.jl
depial: Sure! I can throw a couple in there like your first examples, but I've got a question about the `classify`:
Due to the wide range of possible decision boundaries returned by Perceptron, beyond carefully selected examples, I'm not sure if testing classification of unseen points is viable. Also, since the algo is effectively just the three parts of classify + update + repeat, couldn't requiring a separate classification function introduce an unnecessary constraint on the task?

cmcaine: Cool!

I don't see why it would constrain the student or the task for us to ask them to provide a classify function. What do you mean by that?

I agree that testing with unseen points is a bit tricky, but it's not super hard given that we control the input data and it's not too bad for us to either only include hand-written tests or to do some geometry to ensure our unseen pointd will always be classified correctly by a valid linear boundary.

depial: I guess my thoughts were that, since classification is already an integral part of the algorithm, requiring a separate classification function doesn't really add anything substantive to the task. One thing this could constrain is that it could encourage a modular implementation over a unitary one, as opposed to leaving this option more open to the student. This is not a big issue, but I don't really see the benefit of adding the extra requirement since it's already built in.

cmcaine: I think a good reason is that it helps explain how these things work. A real supervised learning interface would usually have a function for training the classifier and another for using the output of the training to do some classification.

exercises/practice/perceptron/.docs/instructions.md
cmcaine: I'll see if I have time and energy to attempt the exercise myself during the week and maybe that will help me clarify my ideas :)

exercises/practice/perceptron/.docs/instructions.md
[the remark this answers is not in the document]
cmcaine: Great!

exercises/practice/perceptron/runtests.jl
cmcaine: I think keeping it at functions might be better, so that there's only one confusing thing to work out at a time:

```jl
fit(points, labels) -> boundary
predict(boundary, point) -> label

# or possibly
predict(boundary, points) -> labels
```

If we go with a struct, then I think we'd end up with something like this if we followed what's done in MLJ and Lux.jl:

```jl
struct Perceptron
  # parameters for customising the perceptron go here, e.g. learning rate r
  r::Float64
  Perceptron(; r) = new(r)
end

struct Model
  w::Vector{Float64}
  # ...
end

model = fit(Perceptron(r=0.2), points, labels)
predicted_labels = predict(model, unseen_points)
```

Which is maybe a bit too complicated?

depial: Since the only non-trivial hyperparameter that the basic Perceptron algo has is the initial hyperplane, I think we could get away with pre-defining a single struct and leaving the student to define the `fit` and `predict` methods. Maybe making the slug something like:

```julia
mutable struct Perceptron
    # instantiates Perceptron with a decision boundary (this struct can remain unmodified)
    dbound::Vector{Float64}
    Perceptron() = new([0., 0., 0.])
end

function fit!(model::Perceptron, points, labels)::Vector{Float64}
    # updates and returns the decision boundary: model.dbound

end

function predict(model::Perceptron, points::Vector{Vector{Float64}})::Vector{Float64}
    # returns a vector of the predicted labels of points against the model's decision boundary

end
```
This would leave only methods to be defined while still giving a look into how structs work and a more realistic interface (whose use will be able to be seen in the tests).

On whether the `predict` method should be for a single point or multiple, I would say that multiple is better, since it's easier to wrap a single point in a Vector as input for `predict` rather than to ask the student to handle this difference inside the method (not to mention it's closer to what they'd see in practice).

exercises/practice/perceptron/.meta/tests.toml
depial: Looking at some problem-specifications, I'm wondering if psedorandom test generation is possible on the platform. Checking `canonical-data.json` for Alphametics shows each exercise with data such as a unique UUID, plus the input and expected result. If this file is necessary and needs the hardcoded input and expected output, although possible, it'll be a bit more work to produce.

cmcaine: canonical-data isn't required. You can see parametric tests in the rational-numbers exercise and elsewhere.

exercises/practice/perceptron/runtests.jl
cmcaine: Is the learning rate trivial?

I'm fine with either of these approaches (two methods, with or without a struct). Are you happy to make the changes to the PR?

exercises/practice/perceptron/.meta/tests.toml
depial: Great! I'd forgotten about the testset at the end of Rational Numbers :)

exercises/practice/perceptron/runtests.jl
depial: >Is the learning rate trivial?

Sorry, over-generalization from me there :) I guess I'm considering the basic (pedagogical) Perceptron, i.e. one provided with small, dense, separable populations, since, under separability, the learning rate doesn't affect either the final loss or the upper bound of the number of errors the algorithm can make. That said, I was wrong to say the initial hyperplane affects these, since it doesn't either. It just seems non-trivial to me because it can be returned.

I'll try to make the necessary changes to the PR today and/or tomorrow :)

depial: >I agree that testing with unseen points is a bit tricky, but it's not super hard given that we control the input data and it's not too bad for us to either only include hand-written tests or to do some geometry to ensure our unseen pointd will always be classified correctly by a valid linear boundary.

I was thinking about this, and it appears it is redundant to test unseen points which are guaranteed to be on the correct side of any possible decision boundary for a population. That's because there are two hyperplanes which define the limits of a cone of possibility, and each of these limits can be defined by two points, one from each class. These two points work like support vectors, and are robust to any additional (guaranteed valid) unseen points. This means only the support vectors need to be classified correctly to verify that the hyperplane is a valid decision boundary, and this is already done under the current testing.

For this reason I think it would be best to leave out testing of unseen points, since, for students encountering this for the first time, it could give the unintended impression that Perceptron will always correctly classify unseen points.

Any thoughts?

exercises/practice/perceptron/.docs/instructions.md
depial: Edit: moving this down here to split up topics

> I agree that testing with unseen points is a bit tricky, but it's not super hard given that we control the input data and it's not too bad for us to either only include hand-written tests or to do some geometry to ensure our unseen pointd will always be classified correctly by a valid linear boundary.

I was thinking about this, and it appears it is redundant to test unseen points which are guaranteed to be on the correct side of any possible decision boundary for a population. That's because there are two hyperplanes which define the limits of a cone of possibility, and each of these limits can be defined by two points, one from each class. These two points work like support vectors, and are robust to any additional (guaranteed valid) unseen points. This means only the support vectors need to be classified correctly to verify that the hyperplane is a valid decision boundary, and this is already done under the current testing.

For this reason I think it would be best to leave out testing of unseen points, since, for students encountering this for the first time, it could give the unintended impression that Perceptron will always correctly classify unseen points.

Any thoughts?

exercises/practice/perceptron/runtests.jl
depial: Is there anything you would add (e.g. subtyping) to my suggestion for a slug, or do you think I could just copy and paste?

cmcaine: I don't think the struct is really adding much, so I'd remove it. You can pick what you like, tho.

If I try the exercise and hate the struct then I might veto it, but not until then.

exercises/practice/perceptron/.docs/instructions.md
cmcaine: I don't think these objections are relevant?

The tests exist to guide the students to a correct solution. It's fine and expected if the points we provide are guaranteed to be correctly classified *by a correct implementation*. The point of the tests is just to catch bad implementations.

cmcaine: Thanks for explaining :)

I want us to have at least one test with unseen points (and ideally more) because I believe that:

1. the tests will make more sense to students who read them
2. confused students won't be able to encode the seen points in the perceptron and pass the tests.

I agree that the tests should be unambiguous. We can have unseen points without producing ambiguous tests by picking points that are geometrically guaranteed to be on the right side of any valid linear boundary between the classes.

E.g. for linearly separable datasets, I think the centroid of any triangle drawn from points of the same set will be classified correctly by any correct implementation.

Another option would be to train on and classify larger datasets and assert that the accuracy is above some threshold that it is either impossible or implausible for a correct implementation to miss.

We don't need to provide the code for generating the test data to the student. We can precompute some datasets and include them in the test file.

I don't think that it's necessary to test the training and classification separately, though we can choose to do so. I think it's fine to just call both functions for every test, as shown above.

Does that make sense?

depial: I believe I understand your intentions, but I want to check if I'm getting everything:

> 2. Confused students won't be able to encode the seen points in the perceptron and pass the tests.

I can possibly understand this two ways, so I'll take a stab at both: 

- Do you mean students won't have seen-points to use locally to debug? If so, the first set of fixed tests have the points and labels shown explicitly. 
- Or, do you mean they might hard code the answers into their function in order to pass the tests (i.e. 'cheat')? If so, the pseudorandom tests don't explicitly show the seen points, and someone would have to hard code over 1200 points across 40+ tests to pass, which, while doable, seems unlikely (also couldn't they just look at other students' solutions after submitting once?).

> We can have unseen points without producing ambiguous tests by picking points that are geometrically guaranteed...

This actually still leaves an ambiguity. If a student presents an incorrect decision boundary (DB), the unseen point is not guaranteed to be either correctly or incorrectly classified. It all depends on the particular DB. Furthermore, if the point happens to be classified correctly, it can give the impression that the DB is _correct_ instead of properly relaying the actual information of being just _not incorrect_, thereby misleading the student.

Just a note: the reason I said that any test that can be done with unseen points can be done as well or better with seen-points is that correct classification of a maximum of four special points (support vectors) out of the seen-points is both necessary and sufficient to tell if the DB is 100% correct or not. _Any other test is technically deficient on its own, and deficiency introduces ambiguity_. On the other hand, any other in conjunction with classifying the support vectors is superfluous. However, finding them is a non-trivial task, so every point is exhaustively classified in testing (to make sure they are found and tested).

> Another option would be to train on and classify larger datasets and assert that the accuracy is above some threshold that it is either impossible or implausible for a correct implementation to miss.

I believe this is what the pseudorandom test set does and accuracy is required to be 100%. Is there something further that I'm missing?

> I think it's fine to just call both functions for every test.

I believe this adds unnecessary confusion when there is a failed test as to which function is causing the failure.

> 1. the tests will make more sense to students who read them.

At this point, even if this is the case, the added possibility of confusion created through various ambiguities, and/or the difficulties in trying to mitigate them, make including these features seem to be not worth it. I don't mind including ambiguity in exercises, since it provides 'real world' experience, but it depends on the audience. From what I gather from your concern in this point, you are looking out for less experienced students, so I would presume less ambiguity would be better?

>E.g. for linearly separable datasets, I think the centroid of any triangle drawn from points of the same set will be classified correctly by any correct implementation.

Thanks! I was trying to think of the simple way you had mentioned, but I had lost the trees for the forest :)

depial: > I don't think that it's necessary to test the training and classification separately, though we can choose to do so. I think it's fine to just call both functions for every test, as shown above.

Sorry, I just wanted to clarify, I am on board with a classification function if it's tested separately before being used in conjunction with the perceptron function. It's really the unseen points I'm most hesitant to accept.

cmcaine: I think your belief that unseen points will cause ambiguity (test failures for valid decision boundaries) is making you think that they're a big problem.

But, as I hope the triangle centroid thing shows, we can generate unseen points with zero chance of them causing false test failures.

If you don't believe the triangle thing, then maybe you can believe that, for separable datasets, any point contained by the convex hull of the points of one label will be given that label by any boundary-based classifier that has zero error on the training set. For a linear boundary and a 2D space that one should be provable geometrically with a ruler and some drawings.

depial: I think I have failed to unmuddle my concern :) I'm okay with testing to demonstrate how classification of unseen points works, but testing as proposed is too easily confused with testing unseen points to verify the correctness of a decision boundary. My concerns about why testing unseen points to verify correctness is undesirable follow:

I'm with you on the centroid idea (which I quite liked) and familiar with the other arguments you've given, so I'm not concerned about test failures for _valid_ decision boundaries. The minor issue I have in the case of valid decision boundaries is that testing with unseen points to validate a decision boundary is simply redundant. Why it is redundant can be gleaned from your arguments on why and how we can choose correct unseen points: namely they are chosen entirely in reference to seen points.

One issue that concerns me involves _invalid_ decision boundaries. With an invalid decision boundary, there is no guarantee how an unseen point will be classified (just as there is no guarantee how _any_ point will be classified). There are two cases:

1. The unseen point is classified incorrectly and there is a test failure. The student is alerted to the fact that their decision boundary is incorrect. This I'm okay with, although it is still redundant.
2. The unseen point is classified correctly and the test passes. The student is not alerted to the fact their decision boundary is incorrect and either receives conflicting information from other tests or may indeed end up believing that their decision boundary is correct. This is the first aspect I dislike.

The other issue that concerns me is on a more psychological note: People often take partial information, over-generalize and create false impressions (e.g. stereotypes/prejudices/etc). That's how some of the more naive or inattentive students, seeing testing on unseen points in this way, could walk away with the mistaken impression that Perceptron can correctly classify _any_ unseen point (not just carefully selected points). This is the second aspect I dislike.

So, with what the possible outcomes are from testing unseen points, redundant good info or unnecessary potentially misleading info, and the possible misconception about Perceptron that can be conveyed simply by including the testing, I don't feel the testing of unseen points in the proposed manner brings enough to the table to be included.

**To boil it all down: I feel that 'testing' unseen points to demonstrate how classification works has value, but I think this is too easily confused with testing to verify correctness of the decision boundary. I currently can't think of tests which can eradicate my concerns above.** However, there are related ideas we could demonstrate.

For example, we could show that the introduction of a new point to the data will (in all likelihood) result in a different decision boundary. This doesn't involve classification of unseen points though. As an aside: with SVM, we could introduce an element of a demonstration of unseen point classification switching with a before and after approach, but it just doesn't work with Perceptron because of the non-uniqueness of the decision boundary (i.e. we can't be guaranteed a switch in classification unless we control the initial hyperplane the student uses).

Do you have any other ideas on ways to demonstration classification of unseen points?

depial: The cost vs benefit analysis that is going on in my head is the following:

1. Testing unseen points to verify correctness is both redundant and carries downsides
2. Testing unseen points to demonstrate classification is desirable, but unnecessary to satisfy the aim of the exercise.

The exercise is first and foremost meant to have students learn about the Perceptron algorithm. Including a demonstration of how classification works would be a bonus, but I just am not sure this type of testing environment properly supports demonstration of this kind.

exercises/practice/perceptron/.docs/instructions.md
ErikSchierboom: > $$[w_0', w_1', w_2'] = [w_0 \pm l_{class}, w_1 \pm x*l_{class}, w_2 \pm y*l_{class}]$$

We also don't support rendering equations as far as I can recall.

exercises/practice/perceptron/.docs/instructions.md
```diff
@@ -0,0 +1,50 @@
...
+
+We will be playing in a two dimensional space, so our separating hyperplane will simply be a 1D line. You might remember the standard equation for a line as $y = ax+b$, where $a,b \in \Re$, however, in order to help generalize the idea to higher dimensions, it's convenient to reformulate this equation as $w_0 + w_1x + w_2y = 0$. This is the form of the hyperplane we will be using, so your output should be $[w_0, w_1, w_2]$. In machine learning, ${w_0,w_1,w_2}$ are usually referred to as weights. 
+
+Scaling a hyperplane by a positive value gives an equivalent hyperplane (e.g. $[w_0, w_1, w_2] = [\alpha \cdot w_0, \alpha \cdot w_1, \alpha \cdot w_2]$ with $\alpha > 0$). However, it should be noted that there is a difference between the normal vectors (the green arrow in illustration above) of hyperplanes scaled by a negative value (e.g. $[w_0, w_1, w_2]$ vs $[-w_0, -w_1, -w_2]$) in that the normal to the negative hyperplane points in opposite direction of that of the positive one. By convention, the perceptron normal points towards the class defined as positive, so this property will be checked but not result in a test failure.
```
ErikSchierboom, line 20: I think this is probably best extracted to a separate `introduction.md` file (not all practice exercises use it, but we recommend using it when possible).

exercises/practice/perceptron/.docs/instructions.md
```diff
@@ -0,0 +1,50 @@
...
+
+#### Updating
+Checking if an object is on one side of a hyperplane or another can be done by checking the normal vector which points to the object. The value will be positive, negative or zero, so all of the objects from a class should have normal vectors with the same sign. A zero value means the object is on the hyperplane, which we don't want to allow since its ambiguous. Checking the sign of a normal to a hyperplane might sound like it could be complicated, but it's actually quite easy. Simply plug in the coordinates for the object into the equation for the hyperplane and check the sign of the result. For example, we can look at two objects $v_1,v_2$ in relation to the hyperplane $[w_0, w_1, w_2] = [1, 1, 1]$:
+
+$$\large v_1$$ 
+
+$$[x_1, y_1] = [2, 2]$$ 
+
+$$w_0 + w_1 \cdot x_1 + w_2 \cdot y_1 = 1 + 1 \cdot 2 + 1 \cdot 2 = 5 > 0$$
```
ErikSchierboom, line 29: We don't support rendering equations I think, so that would make this text quite hard to read. You'd also have to consider that people working locally in their editor of choice might also not see the equations rendered. Could you rewrite this to regular text?

exercises/practice/perceptron/runtests.jl
ErikSchierboom: > I think we should include some tests with manually specified input data of just a few points to make this more approachable and also as good practice (e.g. first few tests could be spaces with just 2-4 points placed manually).

Agree with this

depial: > Agree with this

Could you be more specific? There are already four tests with six manually specified points which check for different possible orientations of a decision boundary. 

Beyond this, we've had an extensive conversation about using "unseen points". (TLDR: I believe testing unseen points to be potentially more detrimental to understanding than beneficial)

The other ideas for tests were of the type to see if the student is returning the correct object (vector of three real numbers), etc. Is there something else you were thinking?

exercises/practice/perceptron/.docs/instructions.md
depial: > We also don't support rendering equations as far as I can recall.

Fixed.

exercises/practice/perceptron/.docs/instructions.md
depial: To be honest, I'm not 100% sure how to do this, mainly because there is continuity between the sections. Do the separate files `introduction.md` and `instructions.md` need to be self-contained? Or can they reference each other, as these sections do?

exercises/practice/perceptron/.meta/example.jl
```diff
@@ -0,0 +1,8 @@
+function perceptron(points, labels)
+    θ, pnts = [0, 0, 0], vcat.(1, points)
+    while true
+        θ_0 = θ
+        foreach(i -> labels[i]*θ'*pnts[i] ≤ 0 && (θ += labels[i]*pnts[i]), eachindex(pnts))
+        θ_0 == θ && return θ
+    end
+end
```
kotp, line 8:
```suggestion
end

```

exercises/practice/perceptron/perceptron.jl
```diff
@@ -0,0 +1,3 @@
+function perceptron(points, labels)
+    # Perceptronize!
+end
```
kotp, line 3:
```suggestion
end

```

exercises/practice/perceptron/.docs/instructions.md
cmcaine: Maybe use unicode subscripts?

depial: Done! And I cleaned up a couple of other things using the appropriate unicode characters.

cmcaine:
Sorry for radio silence. I've not had the energy to do the exercise yet. Maybe next week. If anyone else fancies giving it a go and posting their comments about it then that's welcome and about the only thing that will speed up the process.

Otherwise, no worries, I'll get to it when I can.

depial:
> Sorry for radio silence. I've not had the energy to do the exercise yet.

No worries! I'm in no hurry and have been learning a fair bit, which leaves me quite content :)

I've updated a few files:
- Example solution in `example.jl` has been exploded (i.e. ala Russian nesting doll), so it is in a more modular form. That way, it wouldn't have to be changed if a `classify` function gets included in the tests for the exercise.
- `runtests.jl` has had the element type and length tests added, and `tests.toml` has had manually created UUIDs inserted for said tests.
- `runtests.jl` has had the manually specified six-point populations changed to four-point populations, so they are easier for the student to visualize. Also the `dotest` function has been renamed to `isvalidboundary`, to be more informative in testing.
- Minor variable name changes have been made throughout to be more informative.

After a lot of contemplation on the other suggested tests, I've come to the conclusion that the potential collateral effects are too detrimental to this particular learning environment for me to justify adding them. Of course, discussion on this is still welcome.

I'll look at splitting the `introduction.jl` into `introduction.jl` and `instructions.jl` files, but I'm still not sure if they have to be self-contained or can reference one another. I believe referencing is not a problem on the website, since the two seem to show up one after the other, but I'm unsure about how it would look on the CLI since I'm not very familiar with it.

cmcaine:
Glad to hear you're content and those changes sound fine :)

For the unseen points, I think your complaint is that we can pick valid unseen points but we can't ensure that an invalid decision boundary or implementation will fail those tests. Whereas if we test only the support vectors then a decision boundary that passes is by definition valid.

Is that your position?

I agree that that is true. I don't think it's a problem, though. It's extremely common for an incorrect solution to pass some tests and not others. My suggestion is that we test the support vectors (or all the labeled points, whatever, and also some unseen points).

I can't help regarding the introduction/instructions split. CLI users probably get either two files or the two files are concatenated into a single README.md

ErikSchierboom:
The CLI will concatenate them

depial:
> For the unseen points, I think your complaint is that we can pick valid unseen points but we can't ensure that an invalid decision boundary or implementation will fail those tests. Whereas if we test only the support vectors then a decision boundary that passes is by definition valid.

This is one worry. However, I don't worry that the student will get past all the other tests, but more about what implicit information is being given to the student by using this testing. For example, my biggest concern is that some (inattentive/naive) students, seeing testing of this type, can form the misconception that Perceptron will always classify unseen points correctly. 

Don't get me wrong, I like to induce misconceptions when teaching in person, but that's because dispelling them often produces a memorable lesson. However, here, we don't have the luxury of noticing or dispelling misconceptions. That's why I think it's better to mitigate the possibility of their inclusion.

Another possible misconception that could come from the proposed `classify` function, comes from the fact that we would essentially be testing the students' code against itself, since classification is a vital part of the Perceptron algorithm (e.g. code function A, then use function A in function B, then test function B with function A). This can leave some students thinking it's okay to test their code against itself, and this is the reason that I insisted on fully testing the `classify` function before using it in the proposed manner.

In short, I think that the other tests are overstepping the bounds of what can be reliably accomplished with this type of exercise on this platform in that they are getting into where we would need to test for understanding.

depial:
I've realized that I haven't detailed the logic of the current testing sequence, which is the following progression:

1. Four points centered on the origin (e.g. optimal decision boundary would be a line through origin with slope = -1)
2. Same four points, but with labels negated, which emphasizes the importance of the correct normal
3. Same four points as Test 1, but offset, so a decision boundary cannot pass through the origin
4. Four points which require a vertical or nearly vertical decision boundary (discourages geometric solutions)
5. The remaining parametric tests present increasing populations, while incorporating the aspects from the first four tests

Is there something significant the proposed tests add which warrants all the drawbacks or which couldn't be remedied by further testing of the sort above?

cmcaine:
> **my biggest concern is that some (inattentive/naive) students, seeing testing of this type, can form the misconception that Perceptron will always classify unseen points correctly.**

Cool, I just don't think this is likely and it's not clear to me why this would be worse than students getting confused about why we only ever test with the same data we provided as input.

> I've realized that I haven't detailed the logic of the current testing sequence, which is the following progression:

These sound good and well-thought out :) I would like unseen points to be tested as well, but whatever, let's leave that disagreement until the rest of the exercise is reviewed.

depial:
Maybe there is a slight source of confusion in using an output as a proxy to test for the *correctness* of an algorithm versus using an output as a proxy to test the *accuracy* of an algorithm? The former is the intended scope of this exercise, while I believe testing unseen points does the latter. A quick analogy to show the difference:

We can test the _Alphametics_ algorithm by inputting _a puzzle_, then taking the output _dictionary_ and testing it against _unseen puzzles_ which have the same _encoding_ as the original *puzzle*.

This test is redundant in the same way as the proposed testing with unseen points*. 

I believe the two testing paradigms aren't usually conflated on Exercism because most (if not all) the current exercises are not ML related, and testing for accuracy is the paradigm most associated with ML. While Perceptron is used in ML, I don't want to mix the two paradigms here because correctness should be shown before testing accuracy, and trying to do the two together introduces a level of complexity which I don't think is beneficial at this stage. We could always make another exercise which tests Perceptron accuracy, but it would likely involve (or be entirely centered around) hyperparameter tuning, like an actual use case in ML.

To briefly highlight the key differences in the testing paradigms:
- Correctness: An algorithm is fed many different inputs, resulting in many separate outputs which are then tested against a known standard.
- Accuracy: An algorithm is fed one input (whole or part of a population) and the output is tested against either the same input or (numerous) examples from the same population as the input.

I'll be happy to respond to your other comments after the rest of the exercise is reviewed as you suggested, but I just wanted leave this for your consideration because I feel it nicely encapsulates the underlying source of many of my misgivings.

*Most of the italicized word have fairly obvious analogues in the _Perceptron_ case, with the possible exception of _encoding_, whose analogue is _support vectors_.

depial:
I had made a new practice exercise from the library (binary-search-tree) and thought I would be able to open a new PR with it, but it got included in this one. I'm not sure if I've done something wrong or if this is expected. Let me know if there's anything I should do about it, but I'd like to explain the new exercise, since the tests from `canonical-data.json` seem to be fairly limited in scope (only testing instantiation and sorting), unless there's another way to implement them.

Also, it looks like there are checks that are not successful due to three tests in Circular Buffer failing in the `Exercise CI / Julia nightly - OS` group of tests. Does this have anything to do with the PR?

exercises/practice/binary-search-tree/binary-search-tree.jl
```diff
@@ -0,0 +1,2 @@
+# Create a (Mutable) Struct BinarySearchTree, which has fields: data, left, right
+# Also write a sort method, which returns a sorted array of the elements in a BinarySearchTree
```
kotp, line 2:
```suggestion
# Also write a sort method, which returns a sorted array of the elements in a BinarySearchTree.

```

It seems that this should be sentences in a comment, but surely the EOL is important as well for the last line of this stub.

depial:
Sorry, I feel like I've made a mess of things here. Closing

kotp:
> Sorry, I feel like I've made a mess of things here. Closing

Another option is if you have a commit that you can go back to that you can identify as the point at which you feel things are where they should be, you can reset hard and force push to your own branch.  That can bring you back to where you have less of that feeling, while keeping the work that you feel has some potential.

It might be better than opening a new PR, and can keep the communication in one place (those that are following alone, etc) will not have to re-subscribe to the work being done that they may be interested in.

Of  course, it is up to you how you would want to manage it.

cmcaine:
I still think this is good, I've just been too busy to properly review (and
now on holiday!!).

Please don't delete the branch: I would like the option of working on this
even if you decide you don't want to spend any more time on it :)

depial:
@kotp Thanks for the suggestion and comments! I'll consider how is best to move forward. I just felt the PR was becoming unwieldy because of the new exercise (which may also be able to use a good bit of feedback).

I'll keep the branch, as @cmcaine suggests, so anyone can use whatever I've done in the meantime. I don't know if there is anything more I could add to the Perceptron exercise, so feel free to take it and modify it as you please.

Likewise, the Binary Search Tree exercise is also available if anyone wants it. I feel it is rather 'bare-bones' since I merely followed what was in `canonical-data.json` for the tests. I may add some "optional tests" (ala Circular Buffer) for checking membership and adding elements, to allow for testing to be brought more in line with the description if desired.

depial:
Just an update: I've opened an [issue](https://github.com/exercism/julia/issues/714) with a list of the exercises I have finished making and a link to the branch in order to make more people aware of the available code. Hopefully this is an acceptable solution.

PR #693, opened 2024-01-25
Added protein translation practice exercise

golanor:
Here is a suggestion for the protein-translation exercise in Julia using string macros. Let me know what you think.

### DIFF
--- a/exercises/practice/protein-translation/.meta/example.jl
+++ b/exercises/practice/protein-translation/.meta/example.jl
@@ -23,7 +23,7 @@ struct TranslationError <: Exception
 end
 
 
-function rna_translator(str)
+function rna_to_amino_acids(str)
     n = 3
     result = []
     for i=1:n:length(str)
--- a/exercises/practice/protein-translation/.meta/example.jl
+++ b/exercises/practice/protein-translation/.meta/example.jl
@@ -23,6 +23,10 @@ struct TranslationError <: Exception
 end
 
 macro rna_str(str)
+    rna_translator(str)
+end
+
+function rna_translator(str)
     n = 3
     result = []
     for i=1:n:length(str)
@@ -35,7 +39,8 @@ macro rna_str(str)
         protein == "STOP" && break
         push!(result, protein)
     end
-    return result
+    result
+    
 end
 
 
--- a/exercises/practice/protein-translation/.meta/example.jl
+++ b/exercises/practice/protein-translation/.meta/example.jl
@@ -1,4 +1,4 @@
-codon_protein_dict = Dict(
+const codon_protein_dict = Dict(
     "AUG" => "Methionine",
     "UUU" => "Phenylalanine",
     "UUC" => "Phenylalanine",
@@ -22,9 +22,6 @@ struct TranslationError <: Exception
     message::String
 end
 
-macro rna_str(str)
-    rna_translator(str)
-end
 
 function rna_translator(str)
     n = 3
--- a/exercises/practice/protein-translation/.meta/example.jl
+++ b/exercises/practice/protein-translation/.meta/example.jl
@@ -26,7 +26,11 @@ macro rna_str(str)
     n = 3
     result = []
     for i=1:n:length(str)
-        substring = SubString(str, i, i+n-1)
+        substring = try
+            SubString(str, i, i+n-1)
+        catch
+            throw(TranslationError("invalid rna string"))
+        end
         protein = string_to_protein(substring)
         protein == "STOP" && break
         push!(result, protein)
--- a/exercises/practice/protein-translation/.meta/example.jl
+++ b/exercises/practice/protein-translation/.meta/example.jl
@@ -1,0 +1,42 @@
+codon_protein_dict = Dict(
+    "AUG" => "Methionine",
+    "UUU" => "Phenylalanine",
+    "UUC" => "Phenylalanine",
+    "UUA" => "Leucine",
+    "UUG" => "Leucine",
+    "UCU" => "Serine",
+    "UCC" => "Serine",
+    "UCA" => "Serine",
+    "UCG" => "Serine",
+    "UAU" => "Tyrosine",
+    "UAC" => "Tyrosine",
+    "UGU" => "Cysteine",
+    "UGC" => "Cysteine",
+    "UGG" => "Tryptophan",
+    "UAA" => "STOP",
+    "UAG" => "STOP",
+    "UGA" => "STOP"
+)
+
+struct TranslationError <: Exception
+    message::String
+end
+
+macro rna_str(str)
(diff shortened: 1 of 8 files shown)


cmcaine:
Hi, thanks for your interest. Please rebase and make the tests pass, then ping reviewers again.

golanor:
Hi @cmcaine ,
There seems to be an issue with some versions of Julia when running this test. Do you suggest I change the test so that it would always pass, or should we try fixing the runner so there won't be a difference? It works for 1, and for the nightly (and 1.10 on my machine)

cmcaine:
Julia 1.6 is the current LTS release, so I think we should still support it. The bug is probably because the macros are being expanded at different times on different releases, which might have been done deliberately or not.

We could fix that with eval (`@test_throws TranslationError eval(:(@macroexpand rna"foo"))`), but I think we should require the user to write a function that the string macro is then expected to call. I think this is usually the better option, and is definitely the better option in this case where it is very plausible that a user would want to find the translations of strings that are not known at compile time.

How does that sound to you?

golanor:
I made the changes, but then the string macro is a bit redundant, isn't it?

exercises/practice/protein-translation/.meta/example.jl
```diff
@@ -0,0 +1,51 @@
+codon_protein_dict = Dict(
```
cmcaine, line 1:
```suggestion
const codon_protein_dict = Dict(
```

exercises/practice/protein-translation/protein-translation.jl
```diff
@@ -0,0 +1,8 @@
+macro rna_str()
+    # I'm a ribosome macro!
+end
+
+
+function rna_translator()
+    # I'm a ribosome function!
+end
```
cmcaine, line 8:
```suggestion
function rna_translator()
    # I'm a ribosome function!
end

# Read the manual for more information on string macros:
# https://docs.julialang.org/en/v1/manual/metaprogramming/#meta-non-standard-string-literals
macro rna_str()
    # I'm a ribosome macro!
end
```

exercises/practice/protein-translation/runtests.jl
```diff
@@ -0,0 +1,84 @@
...
+        @test_throws TranslationError rna_translator("AAA")
+    end
+
+    @testset "Incomplete codon causes translation exception" begin
+        @test_throws TranslationError rna_translator("UGUU")
+    end
+
+    @testset "Incomplete RNA sequence can translate if given a stop codon" begin
+        @test rna"UGGUAGUAAAA" == ["Tryptophan"]
+    end
```
cmcaine, line 83: I suggest changing all instances of `rna"..."` to `rna_translator("...")` and adding a couple of tests of the macro at the end.

Reason: the function interface is easier to test and understand for the student.

Alternatively, we could remove the macro version entirely.

cmcaine:
The macro serves the same purpose as other string macros:

1. Mostly a convenient/pretty shorthand
2. Also a way to ensure a computation can happen at compile-time (good for e.g. Regex so that the moderately expensive PCRE regex compilation step can be done at the same time as the Julia compilation step).

What do you want the student to learn from the exercise?

I suggest changing all instances of rna"..." in the tests to rna_translator("...") and adding a couple of tests of the macro at the end.

Reason: the function interface is easier to test and understand for the student.

Alternatively, we could remove the macro version entirely.

golanor:
I wanted to teach both macro writing and creating an exception. I'll settle for only learning how to create a new exception.

exercises/practice/protein-translation/protein-translation.jl
```diff
@@ -0,0 +1,3 @@
+function rna_translator(rna::String)
```
cmcaine, line 1: Function names should describe an action, not an object, according to the Julia style guide and general practice.

How about one of these names? Presented in my preference order.

```suggestion
function rna_to_amino_acids(rna::String)
```

```suggestion
function rna_to_protein(rna::String)
```

```suggestion
function translate_rna(rna::String)
```

cmcaine:
Feel free to put the string macro back in as a bonus question, if you like. The rot13 practice exercise has an example of this.

golanor:
Renamed and added the macro as a bonus question

golanor:
Can we merge this?

cmcaine:
Sorry, I missed the updates. I'll try to take a look this week. Please ping me again if I don't.

cmcaine (commented):
Note to self. Last two commiys look good. Like the rename. Recheck the diff with master before merge.

golanor:
ping @cmcaine

golanor:
@cmcaine Can we merge?

exercises/practice/protein-translation/runtests.jl
```diff
@@ -0,0 +1,88 @@
...
+        @test rna_to_amino_acids("UGGUGUUGA") == ["Tryptophan", "Cysteine"]
+    end
+
+    @testset "Non existent codon causes translation exception" begin
+        @test_throws TranslationError rna_to_amino_acids("AAA")
+    end
+
+    @testset "Incomplete codon causes translation exception" begin
+        @test_throws TranslationError rna_to_amino_acids("UGUU")
+    end
```
cmcaine, line 75: Requiring the user to define a custom exception type makes this exercise less approachable for beginners and may feel like make-work to more experienced programmers.

The norm in Julia is to throw an ArgumentError or a DomainError:

```
julia> parse(Int, "lmao")
ERROR: ArgumentError: invalid base 10 digit 'l' in "lmao"
```

My preference is that we take one of these two options:

1: require user to throw an argumenterror (will require some minor changes to example.jl)

```suggestion
    @testset "Non existent codon causes translation exception" begin
        @test_throws ArgumentError rna_to_amino_acids("AAA")
    end

    @testset "Incomplete codon causes translation exception" begin
        @test_throws ArgumentError rna_to_amino_acids("UGUU")
    end
```

2: require them to throw some kind of exception, but don't worry about the details:

```suggestion
    @testset "Non existent codon causes translation exception" begin
        @test_throws Exception rna_to_amino_acids("AAA")
    end

    @testset "Incomplete codon causes translation exception" begin
        @test_throws Exception rna_to_amino_acids("UGUU")
    end
```

I don't have a preference between those two options.

The other option is that we explain in the instructions.append.md how and why you should define your own exception type. I can't do that because I don't believe that it's worthwhile to define your own exception types :D

cmcaine (changes requested):
Thanks for the reminder.

What do you think of these suggestions?

golanor:
I comitted the suggested changes

cmcaine:
Cool. The example.jl will need to be changed to throw Argument error.

colinleach:
This exercise was added in #746.

PR #694, opened 2024-01-28
Add space-age exercise

BNAndras:
@exercism/julia, part of #48in24.

### DIFF
--- a/config.json
+++ b/config.json
@@ -32,29 +32,12 @@
       ".meta/exemplar.jl"
     ]
   },
-  "concepts": [
-    {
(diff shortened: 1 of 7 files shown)


cmcaine (approved):
Approved, thanks!

In the future please consider either not making semantically irrelevant formatting changes or open a separate PR for formatting/mechanical changes and include the exact code that you have run to reformat the files so that reviewers can mechanically confirm that the diff is fine without having to read the whole thing.

BNAndras:
That was the result of `configlet create —practice-exercise <slug>` which formats the track config. That shouldn’t be an issue in the future since the config is now formatted.

Issue #675, opened 2023-11-25
Port additional exercises?


BNAndras:
I'd like to start porting these exercises over the next few weeks.

- [ ] Bottle Song
- [ ] High Scores
- [ ] Micro Blog
- [ ] Strain
- [ ] Queen Attack
- [x] Eliud's Eggs
- [x] Protein Translation
- [x] Resistor Color
- [x] Resistor Color Duo
- [x] Space Age
- [x] Two Fer

rainerrodrigues:
Hi @BNAndras , can I contribute?

BNAndras:
Sure, this list might be out of date though so double check that exercises aren’t implemented anyway

PR #700, opened 2024-02-05
Add two-fer exercise

BNAndras:
Related to #675

### DIFF
--- a/config.json
+++ b/config.json
@@ -873,6 +873,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 2
+      },
+      {
+        "slug": "two-fer",
+        "name": "Two-Fer",
+        "uuid": "91a8a9c0-041c-4e2e-97fc-dae149ef69f9",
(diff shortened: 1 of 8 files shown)


cmcaine (approved):
Same as eliuded's eggs. Seems fine. Would be nice to have a brief approaches file explaining how to solve it and naming the concepts (default values/optional arguments and string interpolation) and linking to their docs in the manual.

https://docs.julialang.org/en/v1/manual/functions/#Optional-Arguments
https://docs.julialang.org/en/v1/manual/strings/#string-interpolation

BNAndras:
Sure, I'd add a string concatenation approach for completeness, highlighting that interpolation is often more readable.

cmcaine:
For this and the other exercises, do you want me to merge them as-is or would you like to contribute approaches files first? No rush.

BNAndras:
You can merge them. I’m still planning on getting to them.

PR #701, opened 2024-02-05
add eliuds-eggs exercise

BNAndras:
Related to #675


BNAndras:
@exercism/Julia, this is also a #48in24 exercise, but it's not featured until week 43.

cmcaine (approved):
Thanks, seems fine! Would be nice to have an approaches file, but not necessary.

BNAndras:
Sure. Approaches would probably be something like recursion, while loops, and then the built-in I use.

exercises/practice/eliuds-eggs/runtests.jl
```diff
@@ -0,0 +1,21 @@
+using Test
+
+include("eliuds-eggs.jl")
+
+@testset verbose = true "Eliud's Eggs" begin
```
ErikSchierboom, line 5:
```suggestion
@testset verbose = true "tests" begin
```

PR #702, opened 2024-02-05
Add resistor-color exercise

BNAndras:
Related to #675

### DIFF
--- a/exercises/practice/resistor-color/.meta/example.jl
+++ b/exercises/practice/resistor-color/.meta/example.jl
@@ -1,0 +1,20 @@
+COLORS = [
+    "black",
+    "brown",
+    "red",
+    "orange",
+    "yellow",
+    "green",
+    "blue",
+    "violet",
+    "grey",
+    "white"
+]
+
+function colorcode(color)
+    findfirst(==(color), COLORS) - 1
+end
+
+function colors()
+    COLORS
+end
--- a/exercises/practice/resistor-color/.meta/example.jl
+++ b/exercises/practice/resistor-color/.meta/example.jl
@@ -1,4 +1,4 @@
-COLORS = [
+const COLORS = (
     "black",
     "brown",
     "red",
@@ -9,7 +9,7 @@ COLORS = [
     "violet",
     "grey",
     "white"
-]
+)
 
 function colorcode(color)
     findfirst(==(color), COLORS) - 1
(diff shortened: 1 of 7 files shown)


exercises/practice/resistor-color/.meta/example.jl
```diff
@@ -0,0 +1,20 @@
+COLORS = [
```
cmcaine, line 1:
```suggestion
const COLORS = [
```

exercises/practice/resistor-color/.meta/example.jl
```diff
@@ -0,0 +1,20 @@
...
+    "brown",
+    "red",
+    "orange",
+    "yellow",
+    "green",
+    "blue",
+    "violet",
+    "grey",
+    "white"
+]
```
cmcaine, line 12:
```suggestion
const COLORS = (
    "black",
    "brown",
    "red",
    "orange",
    "yellow",
    "green",
    "blue",
    "violet",
    "grey",
    "white"
)
```

Would probably be better. Not that it matters for the example.jl, but if we include it in an approaches intro or if a student goes looking then it would be nice to avoid issues around mutable globals.

exercises/practice/resistor-color/runtests.jl
```diff
@@ -0,0 +1,32 @@
...
+            "brown",
+            "red",
+            "orange",
+            "yellow",
+            "green",
+            "blue",
+            "violet",
+            "grey",
+            "white"
+        ]
```
cmcaine, line 30: We should allow the student to return a variety of ordered collections (definitely vectors or tuples).

```suggestion
        expected_colors = [
            "black",
            "brown",
            "red",
            "orange",
            "yellow",
            "green",
            "blue",
            "violet",
            "grey",
            "white"
        ]
        @test all(colors() .== expected_colors)
```

cmcaine (changes requested):
Thanks for the PR.

Please apply changes from review, otherwise same as others: approaches file would be nice, but not necessary.

cmcaine (approved):
I applied the changes. Please either:

- merge
- or add approaches and re-request review
- or dispute changes

BNAndras:
I don't have write access to the repo so I can't merge. However, since you wanted approaches written, I can take a look at that. I'm not sure how interesting that would be. I think it'd just be use a collection or use the if-else constructs. I can work on that next week though.

PR #703, opened 2024-02-05
Add resistor-color-duo

BNAndras:
Related to #675. I'm moving this exercise in the config before resistor-color-trio. That lumps the three resistor color exercises together so the student can find the next one on the website easily.


exercises/practice/resistor-color-duo/.meta/example.jl
```diff
@@ -0,0 +1,23 @@
...
+    "brown",
+    "red",
+    "orange",
+    "yellow",
+    "green",
+    "blue",
+    "violet",
+    "grey",
+    "white"
+]
```
cmcaine, line 12:
```suggestion
const COLORS = [
    "black",
    "brown",
    "red",
    "orange",
    "yellow",
    "green",
    "blue",
    "violet",
    "grey",
    "white"
)
```

exercises/practice/resistor-color-duo/runtests.jl
```diff
@@ -0,0 +1,33 @@
+using Test
+
+include("resistor-color-duo.jl")
+
+@testset verbose = true "Resistor Color Duo" begin
+    @testset "Brown and black" begin
+        @test colorcode(["brown", "black"]) == 10
```
cmcaine, line 7: It's still weird that this and the trio exercise pass in a vector rather than separate arguments. Is this related to the learning objectives for the exercise?

BNAndras: As far as I know, that was by design. Notably, Haskell does separate arguments (see https://github.com/exercism/haskell/blob/main/exercises/practice/resistor-color-duo/src/ResistorColors.hs and https://github.com/exercism/haskell/blob/main/exercises/practice/resistor-color-trio/src/ResistorColors.hs). However since we already use a vector in resistor-color-trio, it makes sense to do the same here for consistency.

exercises/practice/resistor-color-duo/.meta/example.jl
```diff
@@ -0,0 +1,23 @@
...
```
cmcaine, line 12:
```suggestion
const COLORS = (
    "black",
    "brown",
    "red",
    "orange",
    "yellow",
    "green",
    "blue",
    "violet",
    "grey",
    "white"
)
```

exercises/practice/resistor-color-duo/runtests.jl
cmcaine: Sure, maybe it'll help people who haven't worked out how to index collections yet or something.

cmcaine (approved):
Approved. Add an approaches file if you feel like it, otherwise feel free to merge this if the github UI lets you.

exercises/practice/resistor-color-duo/runtests.jl
BNAndras: Should resistor-color-trio be updated then? You'll break solutions, but the whole resistor color series will be consistent on how they handle arguments.

BNAndras:
> Approved. Add an approaches file if you feel like it, otherwise feel free to merge this if the github UI lets you.

We could show the mathematical approach I used or a reduce-based approach. If we switch this exercise to use two separate arguments for the colors, the reduce approach is still worthwhile but not as "clean" since we'd need to create a collection to then reduce. Otherwise, I don't think we need approaches for how to retrieve the individual band values since that'd be covered by resistor-color.

exercises/practice/resistor-color-duo/runtests.jl
cmcaine: No, I don't think so. I was saying "sure let's do the same as trio for consistency".

exercises/practice/resistor-color-duo/runtests.jl
```diff
@@ -0,0 +1,33 @@
+using Test
+
+include("resistor-color-duo.jl")
+
+@testset verbose = true "Resistor Color Duo" begin
```
ErikSchierboom, line 5:
```suggestion
@testset verbose = true "tests" begin
```

PR #718, opened 2024-03-29
New practice exercise connect

depial:
This is the practice exercise `connect` taken from `problem-specifications` with scaffolding created by configlet.

Possibly relevant points for reviewers to consider:
- Difficulty has been set at 5 (Medium) due to use of a basic graph traversal algorithm in `example.jl`.
- The name of the function tested is `connect`.
- Verbose unit testing implemented
- There is one extra edge case added to the tests: a 1x1 empty board, `[ "." ]`, meant to complement the 1x1 full boards, `[ "X" ]` and `[ "O" ]`. Nothing was added to `tests.toml` in regard to this.

There are further possible edge cases related to the added one above, such as:
-  0x0 board: `[ "" ]`
- 1xn board, `0` wins: `[ ". . . O . X . ." ]` 
- 1xn board, `X` wins: `[ "X X X X" ]`
- And nx1 variations...

These are mostly to prevent the use of length checks , but the already added edge case above probably does enough to discourage that (e.g. `isone(length(board)) && return first(board)` no longer works). Otherwise, testing seems to be sufficient to check for correctness, but a couple more tests could be added to provide more symmetry. For example, more `O` wins could be added since, of the 11 original tests, there are only two `O` wins, and one of those is an edge case.


depial:
If there is anyone @exercism/reviewers who would be interested in working with me to review some practice exercises to add to the track, please let me know.

exercises/practice/connect/.meta/config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "authors": [],
```
BNAndras, line 2: Missing author name here

depial: I've added the original author's name

kotp (approved):

exercises/practice/connect/.meta/config.json
BNAndras: It should be your GitHub name if you're the one porting the exercise to this track.

depial: Is that required?

depial:
I've added my name to `config.json`, but I would rather prefer it be left off if possible.

exercises/practice/connect/.meta/config.json
BNAndras: You can leave it off if you want. You just won’t get the additional rep points on Exercism for authoring the exercise (30 I believe). You’d only get rep for opening this PR.

depial: Thanks for the clarification!

ErikSchierboom: To be clear, it's your github _username_ that's listed here.

ErikSchierboom (approved):
LGTM

PR #733, opened 2024-05-01
Upgrade julia version

ErikSchierboom:
I'm testing if this fixes the macOS build.

### DIFF
--- a/.github/workflows/exercise-tests.yml
+++ b/.github/workflows/exercise-tests.yml
@@ -24,6 +24,9 @@ jobs:
       matrix:
         julia-version: ["1.6", "1", nightly]
         os: [ubuntu-22.04, windows-2022, macos-14]
+        exclude:
+          - julia-version: 1.6
+            os: macos-14
 
     steps:
       - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b


ErikSchierboom:
@cmcaine Upgrading to test with Julia 1.10 fixes the macOS issue where CI would fail for macos. Is there any particular reason we shouldn't do this?

cmcaine:
We test Julia 1.6 because that is the most recent long-term support release. Testing Julia 1.10 should be redundant with testing Julia 1 (which I expect to redirect to the latest stable version of Julia 1, currently 1.10.3).

I think that if it is easy to do then we should test on only Julia 1 and nightly on MacOS. If that's not convenient, then we can test only 1 and nightly on all operating systems.

cmcaine:
The setup-julia action is probably failing for Julia 1.6 on macOS because Julia 1.6 doesn't support ARM versions of macOS. 

The x86_64 binary did work through rosetta, iirc, but maybe an update to the setup-julia action or some other infra has made it not give an x86_64 binary to newer macOS versions.

ErikSchierboom:
> I think that if it is easy to do then we should test on only Julia 1 and nightly on MacOS. If that's not convenient, then we can test only 1 and nightly on all operating systems.

I think this is easy enough to do. I'll look into it! Thanks for explaining.

cmcaine:
Thanks!

PR #739, opened 2024-05-24
Binary search tree

depial:
The `runtests.jl` include explicit testing of instantiation and searching (making up the first two test groups) which go beyond the tests in `canonical-data.json`.

I do have another testset which strictly follows `canonical-data.json`, and it can be used if instantiation and searching isn't explicitly tested (in fact, a non-mutable struct could be used), but I don't think that would be in the spirit of the exercise.

Other notes:
- The testest uses indentation to signify levels of the binary tree which are being tested.
- Difficulty is set at 8 due to use of mutable struct and tree traversal
- Field access methods have been supplied in the stub, but the struct definition is left to the student.


exercises/practice/binary-search-tree/.meta/config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "authors": [],
```
ErikSchierboom, line 2: Could you add yourself as the author?

config.json
```diff
@@ -913,6 +913,14 @@
...
         "prerequisites": [],
         "difficulty": 5
+      },
+      {
+        "slug": "binary-search-tree",
+        "name": "Binary Search Tree",
+        "uuid": "39998558-213a-4f92-953e-5764d15ddd7c",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 8
```
ErikSchierboom, line 923: This might a tad too high I feel. Maybe 6 or 7?

depial: I agree 8 was too high, so I lowered it to 6. In the end, the struct definition is not too difficult (e.g. no need for promotion rules, etc.) and the tree traversal is likely medium difficulty both algorithmically and in terms of language sophistication.

exercises/practice/binary-search-tree/.meta/config.json
depial: I would prefer to leave this blank. Is that alright, or is it necessary?

ErikSchierboom: That's fine

ErikSchierboom:
@depial I'll let @cmcaine look at the Julia code. If this isn't reviewed in a week or so, please ping me.

depial:
@ErikSchierboom there still hasn't been any further review. I think this is the exercise that is the most involved, as far as the code goes, so if you think it would be easier, I could try opening a PR with an exercise that has more straightforward code, and we could leave this one for later?

ErikSchierboom:
I will look at this next week when I return to work!

ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #750, opened 2024-07-10
add new practice exercise dominoes

depial:
There are two suggestions for testing in the problem specifications. The more rigorous choice, of testing
a returned chain for validity, was used here in `runtests.jl` and requires a helper function to evaluate the users function. Main points: 

- Difficulty has been set at 5 due to a pruned dfs approach with returned chain
- Helper function needed for evaluation

Below is another implementation of the algorithm which simply returns true/false if a chain is possible.
The accompanying test set is below that. The change to the algorithm below is very minor, but this change may still be easier for some students. Due to this change being for less experienced students I have explicitly tested as `@test dominoes(input) == true/false` for clarity, instead of using the return value directly (i.e. `@test dominoes(input)/!dominoes(input))`. This points at another potential benefit which is that no helper function is needed for testing the validity of the chain.

Alternate `example.jl`:

```julia
function dominoes(stones)
    function dfs(stones, chain, indices)
        length(indices) == length(stones) && first(first(chain)) == last(last(chain)) && return true
        for (i, stone) in enumerate(stones)
            if i ∉ indices && last(last(chain)) ∈ stone
                check = dfs(stones, push!(chain, last(last(chain)) == first(stone) ? stone : reverse(stone)), push!(indices, i))
                check && return true
                delete!(indices, i)
                pop!(chain)
            end
        end
        false
    end
    isempty(stones) || dfs(stones, [first(stones)], Set(1))
end

```

Alternate testset for above algo:

```julia
@testset "empty input = empty output" begin
    @test dominoes([]) == true
end

@testset "singleton input = singleton output" begin
    @test dominoes([1, 1]) == true
end

@testset "singleton that can't be chained" begin
    @test dominoes([1, 2]) == false
end

@testset "three elements" begin
    @test dominoes([[1, 2], [3, 1], [2, 3]]) == true
end

@testset "can reverse dominoes" begin 
    @test dominoes([[1, 2], [1, 3], [2, 3]]) == true
end

@testset "can't be chained" begin
    @test dominoes([[1, 2], [4, 1], [2, 3]]) == false
end

@testset "disconnected - simple" begin
    @test dominoes([[1, 1], [2, 2]]) == false
end

@testset "disconnected - double loop" begin
    @test dominoes([[1, 2], [2, 1], [3, 4], [4, 3]]) == false
end

@testset "disconnected - single isolated" begin
    @test dominoes([[1, 2], [2, 3], [3, 1], [4, 4]]) == false
end

@testset "need backtrack" begin
    @test dominoes([[1, 2], [2, 3], [3, 1], [2, 4], [2, 4]]) == true
end

@testset "separate loops" begin
    @test dominoes([[1, 2], [2, 3], [3, 1], [1, 1], [2, 2], [3, 3]]) == true
end

@testset "nine elements" begin
    @test dominoes([[1, 2], [5, 3], [3, 1], [1, 2], [2, 4], [1, 6], [2, 3], [3, 4], [5, 6]]) == true
end

@testset "separate three-domino loops" begin
    @test dominoes([[1, 2], [2, 3], [3, 1], [4, 5], [5, 6], [6, 4]]) == false
end
```

### DIFF
--- a/exercises/practice/dominoes/.docs/instructions.append.md
+++ b/exercises/practice/dominoes/.docs/instructions.append.md
@@ -1,0 +1,2 @@
+### Note: 
+In this version of the exercise, you are required to return `true` or `false` if a valid chain is possible or not.
--- a/exercises/practice/dominoes/.docs/instructions.append.md
+++ b/exercises/practice/dominoes/.docs/instructions.append.md
@@ -1,2 +1,2 @@
-### Note: 
+# Instructions append 
 In this version of the exercise, you are required to return `true` or `false` if a valid chain is possible or not.
--- a/config.json
+++ b/config.json
@@ -945,6 +945,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 6
+      },
+      {
+        "slug": "dominoes",
+        "name": "Dominoes",
+        "uuid": "76e27165-47f5-4c4f-b747-c047872ff02c",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5
       }
     ]
   },
--- a/exercises/practice/dominoes/.docs/instructions.md
+++ b/exercises/practice/dominoes/.docs/instructions.md
@@ -1,0 +1,13 @@
+# Instructions
+
+Make a chain of dominoes.
+
+Compute a way to order a given set of dominoes in such a way that they form a correct domino chain (the dots on one half of a stone match the dots on the neighboring half of an adjacent stone) and that dots on the halves of the stones which don't have a neighbor (the first and last stone) match each other.
+
+For example given the stones `[2|1]`, `[2|3]` and `[1|3]` you should compute something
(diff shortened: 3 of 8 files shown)


ErikSchierboom (approved):
I think my personal preference would be the simpler true/false option

depial:
I've change the testing to the alternate `true/false`, and modified the instructions a bit to make the return values clearer.

ErikSchierboom:
> and modified the instructions a bit to make the return values clearer.

Could you revert this change? The instructions are synced from https://github.com/exercism/problem-specifications, so if they need to be improved they should be improved there. What you _are_ allowed to do, is to create an `instructions.append.md` file, which contents will be appended to the `instructions.md` when the user views the exercise.

depial:
I've reverted the changes to `instructions.md` and added a short `instructions.append.md` file.

exercises/practice/dominoes/.docs/instructions.append.md
```diff
@@ -0,0 +1,2 @@
+### Note:
```
ErikSchierboom, line 1:
```suggestion
# Instructions append 
```

ErikSchierboom (approved):

PR #755, opened 2024-07-16
add practice exercise game-of-life

depial:
Adding `game-of-life` from `problem-specifications`. This is a basic implementations with a straightforward matrix traversal `example.jl`. Only two things to note:

- Matrices used instead of nested vectors
- Difficulty set at 2, but I could imagine 3 due to use of matrices

For this exercise, the only salient differences between using nested vectors and matrices is likely just to be how the elements are accessed and/or how they are traversed.  I think this would be a gentle introduction to using matrices, if a user is new to them.

### DIFF
--- a/config.json
+++ b/config.json
@@ -976,7 +976,7 @@
         "uuid": "31e83423-a0c3-4745-8fca-24125e8ef6ce",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 2
+        "difficulty": 3
       }
     ]
   },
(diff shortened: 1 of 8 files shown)


ErikSchierboom (commented):
> Matrices used instead of nested vectors

Good idea!

> Difficulty set at 2, but I could imagine 3 due to use of matrices

I think 3 makes sense for the reason mentioned.

ErikSchierboom (approved):
Awesome! You've been on a roll!

PR #764, opened 2024-08-28
add practice exercise meetup

colinleach:
A #48in24 exercise, highlighted in the week of August 20th.

Difficulty is provisionally set to 4, copying Python, C# and Crystal. I could easily be persuaded to drop to 3, copying Ruby.

The many tests are taking directly from `problem-specifications`. Unlike several other tracks, there are no additional tests of invalid input. Should these be added?

The file `generators/meetup.jl` created the tests adequately for present purposes, but is clunky. It's not clear to me that the track has agreed standards for these generator files?

### DIFF
--- a/exercises/practice/meetup/.meta/example.jl
+++ b/exercises/practice/meetup/.meta/example.jl
@@ -1,0 +1,3 @@
+function meetup(year, month, week, dayofweek)
+
+end
--- a/exercises/practice/meetup/.meta/example.jl
+++ b/exercises/practice/meetup/.meta/example.jl
@@ -1,3 +1,25 @@
-function meetup(year, month, week, dayofweek)
+using Dates
 
+const week_specifiers = Dict("first" => 1, "second" => 2, "third" => 3, "fourth" => 4, "fifth" => 5)
+
+function meetup(year, month, week, dayofweek)
+    currmonth = Date(year, month)
+    days = [Date(year, month, d) for d in 1:Dates.daysinmonth(currmonth)]
+    valid_dates = [d for d in days if Dates.dayname(d) == dayofweek]
+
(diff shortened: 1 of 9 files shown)


colinleach:
Clearly this needs some work!

I'll go away and think about why every test failed CI (returning `nothing`), despite passing locally. The only import is the `Dates` module, part of the standard library which seems to be available in the Exercism online editor.

exercises/practice/meetup/.meta/example.jl
```diff
@@ -0,0 +1,3 @@
+function meetup(year, month, week, dayofweek)
+
+end
```
SleeplessByte, line 3: This should become [exercises/practice/meetup/meetup.jl](https://github.com/exercism/julia/pull/764/files#diff-9e5779cec2b85194336bc974cd9f95cee7533d462fb88a6d6db7cd944c58f68a) and [exercises/practice/meetup/meetup.jl](https://github.com/exercism/julia/pull/764/files#diff-9e5779cec2b85194336bc974cd9f95cee7533d462fb88a6d6db7cd944c58f68a) should become this file.

How it works is:
- in the CI it will take the `example.jl` file and rename it to `meetup.jl`, and then run the tests on it
- in the web editor / CLI, it will download `meetup.jl` as the starting point for the student.

Switch the two files around and you should see the CI pass once again!

colinleach:
_Hmmm_ - I do coding in my retirement to get reassurance that I'm not yet senile. This is sending the message I was hoping to avoid!!!

Many thanks for your help in spotting this. I should have seen it, but read straight past it. Time to stop for the day, I think.

generators/meetup.jl
ErikSchierboom, line 1: I didn't even know Julia _had_ generators!

ErikSchierboom (approved):

PR #771, opened 2024-09-01
add practice exercise grade-school

colinleach:
Difficulty 3 copies Python and C#. Other tracks give it a higher difficulty, and increasing to 4 might be reasonable?

### DIFF
--- a/exercises/practice/grade-school/.meta/config.json
+++ b/exercises/practice/grade-school/.meta/config.json
@@ -3,7 +3,7 @@
     "colinleach"
   ],
   "contributors": [
-    "depial
+    "depial"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/grade-school/.meta/config.json
+++ b/exercises/practice/grade-school/.meta/config.json
@@ -2,9 +2,6 @@
   "authors": [
     "colinleach"
   ],
-  "contributors": [
-    "depial"
-  ],
   "files": {
     "solution": [
       "grade-school.jl"
--- a/exercises/practice/grade-school/.meta/config.json
+++ b/exercises/practice/grade-school/.meta/config.json
@@ -1,0 +1,18 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "files": {
+    "solution": [
+      "grade-school.jl"
+    ],
+    "test": [
+      "runtests.jl"
+    ],
+    "example": [
+      ".meta/example.jl"
+    ]
+  },
+  "blurb": "Given students' names along with the grade that they are in, create a roster for the school.",
+  "source": "A pairing session with Phil Battos at gSchool"
+}
--- a/exercises/practice/grade-school/.meta/config.json
+++ b/exercises/practice/grade-school/.meta/config.json
@@ -2,6 +2,9 @@
   "authors": [
     "colinleach"
   ],
+  "contributors": [
+    "depial
+  ],
   "files": {
     "solution": [
       "grade-school.jl"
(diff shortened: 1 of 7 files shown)


exercises/practice/grade-school/runtests.jl
depial, line 1: I found the use of a struct here a bit interesting, since it means students will have to read its definition to figure out that they need to access the fields directly in their functions (e.g. `student.name`).

I don't know if I've seen this setup in an exercise before, but I'm perfectly fine with making things a bit unpredictable for students, so I have no problems with just keeping the struct as is. However, if you would like to make it a bit more forgiving, another possible option could be to include iteration as a means of access. One way would be to use named tuples instead of the struct:
```julia
Student = NamedTuple{(:name, :grade), Tuple{String, Int}}

student = Student(("Aimee", 2)) #note the tuple input (i.e. extra parentheses) as compared to the struct
```
With this, both `student.name == "Aimee"` and `student[1] == "Aimee"` returns true (while the latter throws a method error with the struct as is)

Another way would be to define an iterate method for the `Student` struct, but that seems like overkill.

colinleach: Yes, I already wondered whether structs were the way to go, so I'm happy to discuss it. I'll create a revised version using NamedTuple, see how it looks and then push it.

colinleach: If it looks like I'm relatively inexperienced in Julia, that's because it's true. Feel free to criticize when I submit slightly-reformatted Python code!

depial: Cheers! Your code looks good to me, but I'm more of a Julia aficionado/hobbyist than any sort of expert myself, so feel free to take my suggestions with a dose of skepticism :D

colinleach:
The `NamedTuple` approach looks good, requiring no changes in my `example.jl` but giving the students more flexibility.

Thank you, I found that educational.

exercises/practice/grade-school/.meta/config.json
depial, line 1: Thank you for considering me as a contributor! However, if it's not too much trouble, I would prefer not to be listed as one. Silly quirk of mine maybe :)

colinleach: Sure, whatever you prefer. I assume this also applies to Kindergarten Garden?

depial: Yep, that'd be great. Thanks!

exercises/practice/grade-school/runtests.jl
```diff
@@ -0,0 +1,240 @@
+using Test
+
+include("grade-school.jl")
+
+Student = NamedTuple{(:name, :grade), Tuple{String, Int}}
```
ErikSchierboom, line 5: I wonder if this should maybe be part of the stub file, to help with discoverability for students.

depial: I initially considered the suggesting that, but then I thought that putting it in the stub it might be confusing to students since it's not actually needed in their functions (I usually assume that any boilerplate needs to be used somehow).

In the end, since it's a `NamedTuple`, it behaves just like a normal `Tuple`, but with the added option of accessing the fields directly, so I don't think it needs to be discovered and could be left as a kind of hidden gem.

ErikSchierboom: Okay!

ErikSchierboom (approved):

ErikSchierboom:
There is a merge conflict

PR #779, opened 2024-09-09
[Basics] new concept

colinleach:
For discussion on the [forum](https://forum.exercism.org/t/julia-learning-track-redux/12826/16).

This concept draft combines and rewrites 3 older concepts. It is paired with the Lasagna concept exercise.

PR #780, opened 2024-09-09
[Lasagna] revision to existing concept exercise

colinleach:
For discussion on the [forum](https://forum.exercism.org/t/julia-learning-track-redux/12826/16)

This revision (the most recent of many):
- Rewrites and shortens the `introduction.md` to meet Exercism guidelines.
- Enforces use of a constant, in place of magic numbers within the code.
- Points the top-level `config.sys` to the `basics` concept, instead of 3 older ones.

The old documentation isn't wasted: I moved some to `basics/about.md` and some to later concepts.

Including the `introduction.md.tpl` was my mistake. It should probably be removed now we no longer need to combine bits from 3 concepts.

### DIFF
--- a/exercises/concept/lasagna/.docs/hints.md
+++ b/exercises/concept/lasagna/.docs/hints.md
@@ -4,18 +4,23 @@
 
 - You will need to define [functions][functions] and use [arithmetic operators][arithmetics].
 
-## 1. Calculate the preparation time in minutes
+## 1. Store the expected bake time in a constant
+
+- Define a [const][const] to store this value.
+
+## 2. Calculate the preparation time in minutes
 
 - Use the [`times` arithmetic operator][arithmetics] to multiply the argument by `2`.
 
-## 2. Calculate the remaining oven time in minutes
+## 3. Calculate the remaining oven time in minutes
 
 - Use the [`minus` arithmetic operator][arithmetics] to subtract the argument from `60`.
 
-## 3. Calculate the total working time in minutes
+## 4. Calculate the total working time in minutes
 
-- Reuse the `preptime` function defined in the first step.
-- Use the [`plus` arithmetic operator][arithmetics] to add the return value of `preptime` to the argument.
+- Reuse the `preparation_time` function defined in the first step.
+- Use the [`plus` arithmetic operator][arithmetics] to add the return value of `preparation_time` to the argument.
 
+[const]: https://docs.julialang.org/en/v1/base/base/#const
 [functions]: https://docs.julialang.org/en/v1/manual/functions/
 [arithmetics]: https://docs.julialang.org/en/v1/manual/mathematical-operations/#Arithmetic-Operators-1
--- a/exercises/concept/lasagna/.docs/instructions.md
+++ b/exercises/concept/lasagna/.docs/instructions.md
@@ -2,9 +2,18 @@
 
 In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
 
-You have three tasks, all related to the time spent cooking the lasagna.
+You have four tasks, all related to the time spent cooking the lasagna.
 
-## 1. Calculate the preparation time in minutes
+## 1. 
+
+Define the `expected_bake_time` constant that returns how many minutes the lasagna should bake in the oven. According to the cooking book, lasagna needs to be in the oven for a total of 60 minutes.
+
+```julia-repl
+julia> expected_bake_time
+60
+```
+
+## 2. Calculate the preparation time in minutes
 
 Define the `preptime` function that takes the number of layers you added to the lasagna as an argument and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
 
@@ -13,9 +22,8 @@ julia> preptime(4)
 8
 ```
 
-## 2. Calculate the remaining oven time in minutes
+## 3. Calculate the remaining oven time in minutes
 
-According to the cooking book, lasagna needs to be in the oven for a total of 60 minutes.
 Define the `remaining_time` function that takes the actual minutes the lasagna has been in the oven as a parameter and returns how many minutes the lasagna still has to remain in the oven.
 
 ```julia-repl
@@ -23,7 +31,7 @@ julia> remaining_time(50)
 10
 ```
 
-## 3. Calculate the total working time in minutes
+## 4. Calculate the total working time in minutes
 
 Define the `total_working_time` function that takes two arguments: the first argument is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven.
 The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
--- a/exercises/concept/lasagna/.docs/instructions.md
+++ b/exercises/concept/lasagna/.docs/instructions.md
@@ -4,9 +4,11 @@ In this exercise you're going to write some code to help you cook a brilliant la
 
 You have four tasks, all related to the time spent cooking the lasagna.
 
-## 1. 
+## 1. Store the expected bake time in a constant
 
-Define the `expected_bake_time` constant that returns how many minutes the lasagna should bake in the oven. According to the cooking book, lasagna needs to be in the oven for a total of 60 minutes.
+Define the `expected_bake_time` constant that returns how many minutes the lasagna should bake in the oven.
+
+According to the cooking book, lasagna needs to be in the oven for a total of 60 minutes.
 
 ```julia-repl
 julia> expected_bake_time
@@ -15,10 +17,10 @@ julia> expected_bake_time
 
 ## 2. Calculate the preparation time in minutes
 
-Define the `preptime` function that takes the number of layers you added to the lasagna as an argument and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
+Define the `preparation_time` function that takes the number of layers you added to the lasagna as an argument and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
 
 ```julia-repl
-julia> preptime(4)
+julia> preparation_time(4)
 8
 ```
 
@@ -34,6 +36,7 @@ julia> remaining_time(50)
 ## 4. Calculate the total working time in minutes
 
 Define the `total_working_time` function that takes two arguments: the first argument is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven.
+
 The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
 
 ```julia-repl
--- a/exercises/concept/lasagna/.docs/instructions.md
+++ b/exercises/concept/lasagna/.docs/instructions.md
@@ -35,7 +35,7 @@ julia> remaining_time(50)
 
 ## 4. Calculate the total working time in minutes
 
-Define the `total_working_time` function that takes two arguments: the first argument is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven.
+Define the `total_working_time` function that takes two arguments: the first argument is the number of layers you added to the lasagna, and the second argument is the number of minutes the lasagna has been in the oven.
 
 The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
 
--- a/exercises/concept/lasagna/.docs/introduction.md
+++ b/exercises/concept/lasagna/.docs/introduction.md
@@ -3,9 +3,77 @@
 The entire Julia track will require you to treat your solution like small libraries, i.e. you need to define functions, types etc. which will then be run against a test suite.
 For that reason, we will introduce named functions as the very first concept.
 
-## Functions
+Julia is a dynamic, strongly-typed programming langauge.
+The programming style is mainly functional, though with more flexibility than in languages such as Haskell.
 
-### Defining functions
+## Comments
+
+Two options are possible in Julia:
+- Single-line comments start with `#`
+- Multi-line comments start with `#=` and end with `=#`. Nesting is allowed.
+
+```julia
+# This is a single-line comment
+
+x = 3  # This is an inline comment
+
+#=
+	Multi-line comments can be used for longer explanations.
+
+	They are especially useful to comment out blocks of code during debugging.
+=#
+```
+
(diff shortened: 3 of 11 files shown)


colinleach:
*Update:* one test is failing on Julia 1.6, the current LTS version, but passes on 1.10, the current latest release.

I'll add a guard clause to this test and push a revision.

FYI - 1.6 will probably be dropped by the end of this year, when 1.11 is released and 1.10 becomes LTS.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
 The entire Julia track will require you to treat your solution like small libraries, i.e. you need to define functions, types etc. which will then be run against a test suite.
 For that reason, we will introduce named functions as the very first concept.
 
-## Functions
+Julia is a dynamic, strongly-typed programming langauge.
+The programming style is mainly functional, though with more flexibility than in languages such as Haskell.
 
-### Defining functions
+## Comments
```
BNAndras, line 9: This section describes how to make comments, but not what they are. The previous section mentions functions being the next concept so going immediately into comments is a bit confusing. This exercise also doesn't involve working with comments beyond noting what needs to be implemented so I'm not convinced it's necessary. Named functions and constants seem to be the focus.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
-    This is most commonly used for one-line function definitions or mathematical functions, where the function body is a single expression.
+	They are especially useful to comment out blocks of code during debugging.
+=#
+```
 
-### Invoking functions
+## Variables and assignment
 
-Invoking a function is done by specifying its name and passing arguments for each of the function's parameters:
+To create a variable, just assign a value to it:
```
BNAndras, line 29: If `x` exists already and I assign a value to it, I didn't just create `x` so this sentence might be misleading.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
+42
 
-# and of course you can invoke a function within the body of another function:
-square_plus_one(x) = muladd(x, x, 1)
+julia> name = "Maria"  # strings are surrounded by double-quotes ""
+"Maria"
 ```
 
-## Integers and Arithmetic operations
+Types are an important subject in Julia, but for now it is best to ignore them them.
```
BNAndras, line 39: Types aren't referenced before so this might be confusing to highlight this especially since there's no context for what a type is. Is it something to do with functions? Variables?

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
-Underscores may be used as a digit separator.
+## Constants
 
-```julia-repl
-julia> 1
-1
-
-julia> -1234
--1234
+Global variables, created outside any function, are:
```
BNAndras, line 44: The heading refers to constants but we don't mention constant until later. Perhaps we should introduce what a constant is first and then discuss global variables to show why a constant is preferred.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
+    ```julia
+    function muladd(x, y, z)
+        x * y + z
+    end
+    ```
 
-julia> div(16, 6)
-2
+    Indentation by 4 spaces is conventional for readability, but the compiler ignores this.
+    The `end` keyword is essential: more like Ruby than like Python.
```
BNAndras, line 89: Referring to Ruby or Python might be confusing. A student would need to know the syntax from both languages to understand this. We can simply say `end` is part of the required syntax for defining the function.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
+2. Using the "assignment form"
 
-[unicode]: https://docs.julialang.org/en/v1/manual/unicode-input/#Unicode-Input
-~~~~
+    ```julia
+    muladd(x, y, z) = x * y + z
+    ```
 
-## Comments
+    This is most commonly used for one-line function definitions or mathematical functions, where the function body is a single expression.
```
BNAndras, line 101: What's the difference between one-line function definitions and mathematical functions? We don't give context for what a mathematical function is. Perhaps we can say this is for making concise single-expression functions and that'd be clear enough.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
-## Types
+## Naming conventions
+
+Like many languages, Julia requires that names (of variables, functions, and many other things) start with a letter, followed by any combination of letters, digits and underscores.
+
+By convention, variable, constant, and function names are *lowercase*, with underscores kept to a reasonable minimum.
 
-Depending on which other programming languages you know, you may expect parameters, variables or return values to have explicit type annotations.
-For now, assume that Julia will infer the types automagically and don't worry about them, we will get to the specifics of the type system in later exercises.
+However, Julia uses Unicode throughout, so "letter" is interpreted quite flexibly: not just *English* letters.
```
BNAndras, line 122: We could provide some examples of valid Unicode function names here.

exercises/concept/lasagna/.meta/exemplar.jl
```diff
@@ -1,3 +1,7 @@
+expected_bake_time = 60
```
BNAndras, line 1: Shouldn't this be a constant?

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,120 @@
...
+- Sometimes necessary.
+- Usually discouraged (only within `*.jl` files; the REPL operates differently).
 
-julia> 1_234_567_890
-1234567890
-```
+If a value needs to be available throughout the program, but is not expected to change, use a constant instead.
 
-### Arithmetic operations
+Prefacing the assignemt with the `const` keyword allows the compiler to generate more efficient code.
```
BNAndras, line 51:
```suggestion
Prefacing the assignment with the `const` keyword allows the compiler to generate more efficient code.
```

exercises/concept/lasagna/.docs/instructions.md
```diff
@@ -2,9 +2,18 @@
...
+Define the `expected_bake_time` constant that returns how many minutes the lasagna should bake in the oven. According to the cooking book, lasagna needs to be in the oven for a total of 60 minutes.
+
+```julia-repl
+julia> expected_bake_time
+60
+```
+
+## 2. Calculate the preparation time in minutes
 
 Define the `preptime` function that takes the number of layers you added to the lasagna as an argument and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
```
BNAndras, line 18: Shouldn't this be `prep_time` globally to match the other functions?

exercises/concept/lasagna/.docs/introduction.md
colinleach: Agree. This should be in `about.md`, not `introduction.md`

exercises/concept/lasagna/.docs/introduction.md
colinleach: Agree

exercises/concept/lasagna/.docs/introduction.md
colinleach: Agree

exercises/concept/lasagna/.docs/introduction.md
colinleach: Agree

exercises/concept/lasagna/.docs/introduction.md
colinleach: Agree

exercises/concept/lasagna/.docs/introduction.md
colinleach: Again, probably better for `about.md` than `introduction.md`

exercises/concept/lasagna/.docs/introduction.md
colinleach: Opposite issue here. I think that's for `about.md`. I don't want to get into this detail for `lasagna`, and maybe removing the line completely would be better?

exercises/concept/lasagna/.meta/exemplar.jl
colinleach: Totally, and at one stage it was. Some sort of odd reversion...

exercises/concept/lasagna/.docs/instructions.md
colinleach: Good catch. It's old code that I didn't read carefully enough.

colinleach:
There's stuff I need to do for the next couple of hours. I'll go through your suggestions again later today and push the changes.

This exercise has been rumbling around for about 5 years. so it would be good to finally get it right!

exercises/concept/lasagna/.docs/introduction.md
BNAndras: We don't use Unicode function names in the exercise itself so removing it makes sense.

exercises/concept/lasagna/.docs/instructions.md
colinleach: For real consistency, I went with `preparation_time` in the revisions I'll push next.

exercises/concept/lasagna/.docs/introduction.md
colinleach: Removed the entire subsection.

exercises/concept/lasagna/.docs/introduction.md
colinleach: Reworded in the latest commit.

exercises/concept/lasagna/.docs/introduction.md
colinleach: Removed in the latest commit.

exercises/concept/lasagna/.docs/introduction.md
colinleach: Switched the latest commit to your shorted wording.

exercises/concept/lasagna/.docs/introduction.md
colinleach: Removed.

exercises/concept/lasagna/.docs/instructions.md
colinleach: This changes touches most of the files. A quick `grep -r preptime` reassures me that I caught them all.

I caught a few other errors while doing this, especially in `hints.md`.

exercises/concept/lasagna/.docs/introduction.md
colinleach: Extensively reworded. What do you think of the new attempt?

exercises/concept/lasagna/.docs/introduction.md
colinleach: Removed the bit about other languages.

Bethany is keen to see cross language comparison, and Julia is unlikely to be any student's first programming language. However, this is for `about.md`, not an introduction to the very first exercise.

exercises/concept/lasagna/.meta/exemplar.jl
colinleach: Definitely there now.

colinleach:
Once we agree on the `introduction.md` here, I'll copy (most of) that to `basics`.

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -4,18 +4,23 @@
...
 - You will need to define [functions][functions] and use [arithmetic operators][arithmetics].
 
-## 1. Calculate the preparation time in minutes
+## 1. Store the expected bake time in a constant
+
+- Define a [const][const] to store this value.
+
+## 2. Calculate the preparation time in minutes
 
 - Use the [`times` arithmetic operator][arithmetics] to multiply the argument by `2`.
```
depial, line 13: If it wouldn't be too much information, we could consider including how Julia can do implicit multiplication here too (maybe as an aside)

```julia
c = 4
2c == 2*c == 8 # evaluates as true
```

colinleach:
I'm trying to keep this first exercise to a minimum, so leaving out implicit multiplication was deliberate. I'm thinking of adding a `numbers` concept, to deal with this, the various sorts of division, float vs int vs bignum, with a brief mention of rational. Python uses the `currency-exchange` exercise for this, which may be worth copying.

I haven't made progress on this in the last few days, other than reading around some other tracks. Blame some passing infection and associated fever (the sort of non-specific thing that epidemiologists lump together as ILI (Influenza-Like Infection) and normal people might call "something going around"). Not an aid to clear thinking.

depial:
I hope you feel better soon!

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,98 @@
...
 ```
 
-## Types
+## Naming conventions
+
+Like many languages, Julia requires that names (of variables, functions, and many other things) start with a letter, followed by any combination of letters, digits and underscores.
 
-Depending on which other programming languages you know, you may expect parameters, variables or return values to have explicit type annotations.
-For now, assume that Julia will infer the types automagically and don't worry about them, we will get to the specifics of the type system in later exercises.
+By convention, variable, constant, and function names are *lowercase*, with underscores kept to a reasonable minimum.
```
BNAndras, line 100: Maybe some examples just to be clear?

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -3,122 +3,98 @@
...
-The standard prefix and infix operations are available: `+`, `-`, `*`, `%`.
+## Arithmetic operators
 
-```julia-repl
-julia> +16
-16
-
-julia> -16
--16
+These mostly work conventionally:
```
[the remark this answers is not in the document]
colinleach: I'll update tomorrow.

exercises/concept/lasagna/.docs/introduction.md
colinleach: This feels like something that should be expanded in `basics/about.md`. I'm trying hard to keep `introduction.md` to the bare minimum needed to solve the exercise. Jeremy has said repeatedly that this is what he wants, and I'm not disagreeing.

colinleach:
Reading through all this again, months later, I think all the comments are addressed?  You can approve when you're happy, but please don't try to merge yet - it will complain about the missing `basics` concept. I'll try to get that concept (#779) updated and ready to merge first.

colinleach:
Oops, no. I had (inevitably) missed one requested change.

depial (approved):
This semi-blank slug seems appropriate for the intentions of the exercise. Everything else also look okay to me

colinleach:
Yes, an exercise to learn how to add functions is the exception to the rule we discussed earlier!

PR #781, opened 2024-09-09
[Booleans] new concept

colinleach:
For discussion on the [forum](https://forum.exercism.org/t/julia-learning-track-redux/12826/16).

This concept draft borrows heavily from `exercises/concept.wip/.docs/introduction.md`, with some edits and new additions. It pairs with `annalyns-infiltration`.


concepts/booleans/about.md
```diff
@@ -0,0 +1,107 @@
...
+If `x` was 3 they would both be `true`.
+We can express statements like "is x less than 5 and not equal to y?" using [Boolean operators][boolean-operators]: `!` (not), `&&` (and), `||` (or).
+
+In Julia (and many other programming languages), `&&` has a [higher precedence][operator-precedence] than `||` (in the same way that `*` is applied before `+`).
+This means that `true || false && true` evaluates to `true` because it is parsed as `(true || false) && true`.
+It is common to include explicit brackets anyway so that the reader doesn't need to think about this.
+
+### Logical _not_
+
+`!` represents the logical "not" operation in  Julia.
```
BNAndras, line 31:
```suggestion
`!` represents the logical "not" operation in Julia.
```

concepts/booleans/about.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Boolean logic
+
+Imagine we have the following Boolean expressions in Julia: `5 > x` and `x != 0`.
+If `x` was 3 they would both be `true`.
+We can express statements like "is x less than 5 and not equal to y?" using [Boolean operators][boolean-operators]: `!` (not), `&&` (and), `||` (or).
+
+In Julia (and many other programming languages), `&&` has a [higher precedence][operator-precedence] than `||` (in the same way that `*` is applied before `+`).
+This means that `true || false && true` evaluates to `true` because it is parsed as `(true || false) && true`.
+It is common to include explicit brackets anyway so that the reader doesn't need to think about this.
```
BNAndras, line 27:
```suggestion
It is common to include explicit parentheses anyway so that the reader doesn't need to think about this.
```

concepts/booleans/about.md
```diff
@@ -0,0 +1,107 @@
...
+julia> 2 < 1
+false
+
+julia> (5 * 5 == 25) || (2 < 1)
+true
+```
+
+### Short-circuit evaluation
+
+When `&&` or `||` are used to link Boolean expressions, Julia evaluates from left to right, and stops when it finds an unambiguous answer.
```
BNAndras, line 85:
```suggestion
When `&&` or `||` are used to link Boolean expressions, Julia evaluates from left to right, and stops when it finds an unambiguous answer.
```

We don't define what an unambiguous answer is before the example so it's a little hard to follow until the explanation. Should we just Julia always evaluates the left side and sometimes  the right side depending on the value of the left side and the operator used?

concepts/booleans/about.md
```diff
@@ -0,0 +1,107 @@
...
+
+Consider this expression:
+
+```julia-repl
+julia> (5 > 12) && (1 != 0)
+true
+```
+
+Because `5 > 12` is `false`, clearly the combined expression must be `false`.
```
BNAndras, line 94: I don't believe we can say the combined expression is clearly `false` since we haven't introduced why the expression would be `false`. We also risk reinforcing whatever interpretation the student got from line 85 because they might think Julia stops when it hits a false value regardless of which operator is used so that's what "unambiguous" means.

concepts/booleans/about.md
```diff
@@ -0,0 +1,107 @@
...
+The `1 != 0` expression is not relevant and is never evaluated.
```
BNAndras, line 95: `1 != 0` isn't evaluated when the left side evaluates to false. It is evaluated if the left side evaluates to true so we can't say it's never evaluated without qualifying that never.

concepts/booleans/introduction.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+## Boolean logic
+
+Imagine we have the following Boolean expressions in Julia: `5 > x` and `x != 0`.
+If `x` was 3 they would both be `true`.
+We can express statements like "is x less than 5 and not equal to y?" using [Boolean operators][boolean-operators]: `!` (not), `&&` (and), `||` (or).
+
+In Julia (and many other programming languages), `&&` has a [higher precedence][operator-precedence] than `||` (in the same way that `*` is applied before `+`).
+This means that `true || false && true` evaluates to `true` because it is parsed as `(true || false) && true`.
```
BNAndras, line 26:
```suggestion
This means that `true || false && true` evaluates to `true` because it is parsed as `true || (false && true)`.
```

However, we haven't properly introduced the logical operators at this point so students who didn't follow the Boolean operators link will likely be confused how we evaluated that expression.

concepts/booleans/introduction.md
```diff
@@ -0,0 +1,84 @@
...
+true
+
+julia> 1 != 0
+true
+
+julia> (5 > 3) && (1 != 0)
+true
+```
+
+Parentheses are optional and can make the code easier to read.
```
BNAndras, line 66: This should be included under the logical OR section. However, we've already mentioned this in the Boolean logic section so it can be dropped.

concepts/booleans/introduction.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Boolean logic
+
+Imagine we have the following Boolean expressions in Julia: `5 > x` and `x != 0`.
+If `x` was 3 they would both be `true`.
+We can express statements like "is x less than 5 and not equal to y?" using [Boolean operators][boolean-operators]: `!` (not), `&&` (and), `||` (or).
+
+In Julia (and many other programming languages), `&&` has a [higher precedence][operator-precedence] than `||` (in the same way that `*` is applied before `+`).
+This means that `true || false && true` evaluates to `true` because it is parsed as `(true || false) && true`.
+It is common to include explicit brackets anyway so that the reader doesn't need to think about this.
```
BNAndras, line 27:
```suggestion
It is common to include explicit parentheses anyway so that the reader doesn't need to think about this.
```

colinleach:
I need to think about this and try again.

colinleach:
Your key point is that comparison operators aren't taught till the next concept. I'm a bit less worried about this in `about.md`, which is allowed to forward-reference, but these should definitely be removed from `introduction.md`.

One take-home message: *don't* blindly reuse text from the older exercise drafts, as I did here.. There are reasons why this syllabus hasn't launched.

colinleach:
I did a major rewrite (essentially from scratch) of both Markdown files, in response to the all-too-valid criticisms from @BNAndras. 

Further comments welcome, from anyone.

concepts/booleans/about.md
```diff
@@ -0,0 +1,120 @@
...
+
+julia> typeof(true)
+Bool
+```
+
+In contrast to several other languages, Julia deliberately has no concept of "truthiness".
+Only expressions which evaluate to `true` or `false` will be treated as a `Bool`.
+
+Specifically, empty arrays or strings will *not* be interpreted as `false`.
+There must be an appropriate test such as `isempty()` if you want special handlind for empty values.
```
BNAndras, line 23:
```suggestion
There must be an appropriate test such as `isempty()` if you want special handling for empty values.
```

concepts/booleans/about.md
```diff
@@ -0,0 +1,120 @@
...
+
+```julia-repl
+julia> !true
+false
+
+julia> !false
+true
+```
+
+These operators will be familiar to users of many other languages, though *not* Python.
```
BNAndras, line 62: This sentence is confusing. `and` and `or` exist in other languages. Perhaps we say the operators and these concepts should be familiar coming from other languages but they might have different names there.

BNAndras (commented):
Not at a computer but these jumped out at me.

colinleach:
Aargh! I have done something very, very wrong with this, and the files changed now include GH workflows and all sorts of nonsense.

Unless I can dig myself out of this soon, I may close this PR and open a new one. It was just about ready to review, before I broke it!

colinleach:
Closing now, I wrecked this PR. Continues with #831.

PR #784, opened 2024-09-11
Fix JSON file formatting

colinleach:
This is purely the result of running `bin/configlet fmt -u -y`.

The changes are largely cosmetic. The most important thing is they remove the noise from `configlet fmt` output, which should now show any new problems without having to scroll through 79 old ones.

### DIFF
--- a/exercises/practice/acronym/.meta/config.json
+++ b/exercises/practice/acronym/.meta/config.json
@@ -1,5 +1,4 @@
 {
-  "blurb": "Convert a long phrase to its acronym.",
   "authors": [
     "PseudoCodeNerd"
   ],
@@ -17,6 +16,7 @@
       ".meta/example.jl"
     ]
   },
+  "blurb": "Convert a long phrase to its acronym.",
   "source": "Julien Vanier",
   "source_url": "https://github.com/monkbroc"
 }
--- a/exercises/practice/all-your-base/.meta/config.json
+++ b/exercises/practice/all-your-base/.meta/config.json
@@ -1,5 +1,4 @@
 {
-  "blurb": "Convert a number, represented as a sequence of digits in one base, to any other base.",
   "authors": [
     "serialhex"
   ],
@@ -17,5 +16,6 @@
     "example": [
       ".meta/example.jl"
     ]
-  }
+  },
+  "blurb": "Convert a number, represented as a sequence of digits in one base, to any other base."
 }
--- a/exercises/practice/allergies/.meta/config.json
+++ b/exercises/practice/allergies/.meta/config.json
@@ -1,5 +1,4 @@
 {
-  "blurb": "Given a person's allergy score, determine whether or not they're allergic to a given item, and their full list of allergies.",
   "authors": [
     "DarkoGNU"
   ],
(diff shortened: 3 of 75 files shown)


depial:
> Should this be marked as `[no important files changed]`? That's something I've not dealt with before, and need to get my head around.

I suspect it doesn't need that. Here's a bit of the [docs](https://exercism.org/docs/building/tracks#h-maintenance) which gives some explanation, but in short the `[no important files changed]` seems to help "avoid triggering unnecessary test runs".

colinleach:
@depial, do you have authority to approve this, or do we need a cross-track-maintainer?

depial (approved):
lgtm

colinleach:
thanks!

depial:
Cheers! 

If anything else is sufficiently reviewed that we can merge, you can feel free to let me know. I'm still not used to having merging powers, so I'm hesitant to use them

colinleach:
> I'm still not used to having merging powers

I know the feeling!

depial:
@colinleach I've just noticed that there seems to be two versions of the Leap exercise on the [website](https://exercism.org/tracks/julia/exercises). Can you see this too or is it something with my browser?

From what I can tell, they are clones, with the only difference I can see being the blurb (though I only made a cursory exploration). Even all of the solutions seem to be the same (at least I assume they are the same since they both have 69 pages of solutions and the same top solutions)

I can't see anything in the `.json` files that would be responsible for that, but I'm leaving the message here because this is my best guess as to what could have lead to it.

Comparative weirdness:
- On the [overview tab](https://exercism.org/tracks/julia) there is only one version and total exercise count is 95
- On the [practice tab](https://exercism.org/tracks/julia/exercises) there are two versions and total exercise count is 96

colinleach:
This is a mess, not of our making.

If you look at it in an incognito window, everything is good. As maintainers, we're seeing a Leap clone because the previous maintainers wanted to use this as a Concept exercise. This is *strictly forbidden* by Jeremy, who insists on keeping Concept and Practice exercises totally separate. I hear rumors that there have been bad-tempered arguments about this in the past, which he refuses to repeat (and he's the CEO, so tends to have the last word).

The files are in [`exercises/concept.wip`](https://github.com/exercism/julia/tree/main/exercises/concept.wip/leap). My recommendation is to submit a PR to remove that duplicate Leap directory. Even if we somehow wanted to use it, we wouldn't be allowed. Rightly, IMHO.

depial:
Now that you mention that, it reminds me of an open [PR #365](https://github.com/exercism/julia/pull/365). It looks like they planned to rename it and then deprecate it, but that was almost 3 1/2 years ago. I don't know how much of that process is still relevant (e.g. Do we have to deprecate something or can we simply remove it?). Also, the file mentioned in the PR, `exercises/concept/leap/`, is no longer there (presumably moved into `concept.wip`)

colinleach:
Just deleting it should be OK.

depial:
I'll try to take a look at doing that tomorrow. 

> Closing https://github.com/exercism/julia/pull/365 may also be a good idea.

I feel like there's quite a few old PRs that are probably able to be closed as obsolete, but I'm not sure if it's the norm to keep them around for some reason

PR #782, opened 2024-09-09
[Annalyns Infiltration] concept exercise

colinleach:
For discussion on the [forum](https://forum.exercism.org/t/julia-learning-track-redux/12826/16).

This existing draft was moved from `exercises/concept.wip`, with some tidying.

PR #785, opened 2024-09-12
Sort practice exercises by difficulty

colinleach:
The ordering of exercises on the [website page](https://exercism.org/tracks/julia/exercises) has become fairly random. It would be good to have them sorted easy-medium-hard like most other tracks.

I got a sorted list with jq:  `.exercises.practice | sort_by(.difficulty)`. There was some manual copying/pasting after that (I forgot a lot of jq, sorry!), but I was encouraged to see that I ended up with the same line count, and `configlet lint` was happy with the result. I think this gives us a list with difficulties 1-10, ascending. The Exercism website should follow this, perhaps after some delay.

Addressing the question of whether difficulties are correctly assigned is a whole other challenge.

### DIFF
--- a/config.json
+++ b/config.json
@@ -60,19 +60,6 @@
         ]
       },
       {
-        "slug": "killer-sudoku-helper",
-        "name": "Killer Sudoku Helper",
-        "uuid": "e0142403-af9e-406e-827a-1fd37187d1ea",
-        "practices": [],
-        "prerequisites": [],
-        "difficulty": 5,
-        "topics": [
-          "iterator-protocol",
-          "vectors",
-          "conditionals"
-        ]
-      },
-      {
         "slug": "leap",
         "name": "Leap",
         "uuid": "8740c914-f008-4b6b-8e7d-7470e78e6a8e",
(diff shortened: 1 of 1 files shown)


depial (approved):
This will make things nicer! I'd always wondered how the ordering was determined on the website.

colinleach:
The web page looks good in an incognito window, all nicely sorted. I panicked when I looked at it from my own account - it's showing me all the old concept exercise drafts from `exercises/concepts.wip` even though they are not in `config.sys`. That seems to be a maintainer-only thing.

depial:
I noticed those as well and plan to work through them soon. I saw that LaTeX was being used in the instructions/introduction of at least one of them, so I'll probably be opening an issue about migrating those to unicode, since LaTeX doesn't seem to render on the website.

colinleach:
> I noticed those as well and plan to work through them soon.

Honestly, I can't recommend doing much work on the old drafts. I tried using `annalyns-infiltration` in PR #782, as it seemed relatively uncontroversial, but even that didn't go well. My current feeling (backed by feedback from other reviewers) is that we can probably reuse the examplar and runtests code, but all the markdown needs rewriting from scratch. That's what I'm aiming to do, making it closer to the Python syllabus.

See also my (not-yet-written) comments on #780.

Issue #789, opened 2024-09-19
Move old `exercises/concept.wip` drafts 


colinleach:
These older drafts are causing confusion as we try to get a new learning syllabus together. They are hidden from students, but show up to maintainers in distracting and often puzzling ways.

Proposals:

- Delete the duplicate `leap` exercise completely (as already discussed with @depial in #784).
- Move the others to somewhere they will be ignored by the website: out of the `exercises` directory, to some neutral name like `drafts`.
- Re-use the exercise `runtests.jl` and `example.jl` as appropriate, but expect to rewrite the markdown files pretty much completely, based on reviewer feedback.
- With a bit less urgency, also move the `concepts.wip` directory from the top level (note the plural `concepts` in this case)

depial:
It looks like I opened [PR #790](https://github.com/exercism/julia/pull/790) around the time you were raising this issue, so I thought I'd mention it here.

I wouldn't be adverse to moving them to another location where they will be ignored so they are still around for future reference also (I saved a copy of Leap on my machine as a backup before submitting the PR just in case)

colinleach:
I just merged #790. I'll aim to move the remaining items tomorrow.

depial:
What are your thoughts on how to introduce learning mode material? I feel what they appear to have been doing before (staging the learning mode stuff in `concepts.wip`, then, once there's enough material, launching it all at once) seems reasonable.

If you think that is an okay idea, maybe we could keep the non overlapping exercises in the `concepts.wip`, so they aren't wiped from the site, and then work on improving or replacing them individually. I definitely liked some of the exercises in there, so I'd be happy to have a go at getting them ready.

colinleach:
I was just thinking about this, without coming to any clear conclusions. I've been playing about with a few things (concepts more than exercises) on a private repo, intending to copy them to the main repo once I see where things might be going. It would be good to have a staging area where the two of us could discuss them before asking for a full review.

After losing 10 days to illness, I've just rewritten `booleans`. I'm out of time this evening, so I'll read through it again tomorrow then push the new version to #781. A power outage today didn't help - I was just getting restarted on this when everything shut down.

I'll probably do a `numbers` concept next, based on the Python version, so we can see where this fits into the syllabus. I'm undecided between `cars-assemble` or `currency exchange` to go with it, so that's a choice we need to make sometime.

[...]

PR #791, opened 2024-09-20
Move older concept/exercise drafts to older-drafts folder

colinleach:
[no important files changed]

I moved all the unmerged older learning syllabus material to `older-drafts/`, with subdirectories for `concepts` and `exercises`. The aim is that the Exercism website will now ignore these, even for maintainers.

No file *contents* were changed, only the *location*.

I added `.gitkeep` files to `concepts.wip/` and `exercises/concept.wip/` so they stay around, even when empty. I guess we will want to use these as a staging area for new drafts?

Comments welcome on whether this is the best way to nake the change.

I'd love to move the 3 old concepts currently in `concepts/`: these are going to be replaced by the new `basics` concept when it is merged. I'm hesitant:

- The old 3 are merged, with entries in `config.sys`.
- The Exercism website probably expects to find them?

Does anyone know how to deprecate a concept that is no longer needed? Exercises have a `status` field, but there seems to be nothing equivalent for concepts.


colinleach:
I should have said: this follows on from Issue #789.

depial:
> The aim is that the Exercism website will now ignore these, even for maintainers.

I've noticed that the `Leap` that was deleted is still on the website, under the `Practice` tab, while now there is now also a duplicate of `Annalyn's Infiltration` there. I'm wondering why the old ones aren't being deleted from the site?

Another note: I was hesitant to wipe them from the site before we have replacements, but now I feel it wouldn't make a big difference, since I don't think there are many completions anyway.

colinleach:
The issue of duplicates and old concept exercises hanging around only seems to affect maintainers. If you look at the exercises in incognito, it all looks correct.

For Leap, I suspect (but can't be sure) that I still see the Concept version on the Practice tab because I solved Leap in the past: only as a practice exercise, but these are now hopelessly muddled. I assume that successful solutions go into a MySQL table, with (student, track, exercise) fields, so the website still shows it to me as something I published. It would need @iHiD to untangle that at the SQL level, and my guess is that he will see it as more trouble than it's worth - given only the two of us are affected,

It could be worse. I just filtered the practice exercise page to show Annalyn's Infiltration, and I'm seeing a magnificent 7 copies!  I'm hoping this PR will fix that problem, but we'd need to test that empirically.

iHiD:
I think any exercises also need to be marked as deprecated in the config.json before we can merge this. Otherwise I think things will just break.

colinleach:
> exercises also need to be marked as deprecated in the config.json before we can merge this

That's part of the confusion - there is no mention of these draft concept exercises in the `config.json`. They are in a directory called `exercises/concept.wip`, and the website seems to find them and display them to maintainers. We have no way to deprecate them, other than moving them to somewhere more hidden. All suggestions welcome, if I'm missing something.

Leap is a bigger tangle, because it was created at one time as a concept exercise, as well as a practice exercise. That got the software as confused as I am (which is saying something).

BethanyG:
While the `concept` key is set to false, your `config.json` still has a `concept exercise` array, with a number of exercises flagged `wip`.  I _think_ that's what the website is picking up, and what you have to label as deprecated.

Here's the [block](https://github.com/exercism/julia/blob/main/config.json#L36-L60).

As compared to Python, where we have [`wip`, but also one that's deprecated](https://github.com/exercism/python/blob/main/config.json#L46-L99).  The deprecated exercise (`Electric Bill`) doesn't show for anyone, and is noted as deprecated on the [track build status page](https://exercism.org/tracks/python/build).  The WIP exercise (`Pretty Leaflet`) shows to maintainers, and is not shown on the track build status page under active exercises/concepts.

colinleach:
> that's what the website is picking up, and what you have to label as deprecated

Apologies for maybe not explaining it clearly enough, but that block _isn't_ the problem. Lasagna and Annelyn's Infiltration are the result of recent/ongoing PRs.

What we're trying to get rid of are the old drafts such as `dnd-char`, `emoji-times` and the ever-annoying concept variant of `leap`, which are mentioned nowhere in `config.json`.

colinleach:
I should have included a link to the [relevant directory](https://github.com/colinleach/julia_ex/tree/master/exercises/concept.wip).

I'm quite surprised this even exists, more surprised that the website treats it as special (presumably because it's under `exercises`, alongside `concept` and `practice`). We're just trying to empty it, to reduce distractions as we start adding new stuff.

Issue #787, opened 2024-09-17
GHA workflows - are we doing too much CI testing?


colinleach:
The Julia track currently tests 3 versions of Julia against 3 OS's, which seems unusual compared with similar tracks. This is the relevant part of `exercise-tests.yml`:

```yaml
  test:
    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix:
        julia-version: ["1.6", "1", nightly]
        os: [ubuntu-22.04, windows-2022, macos-14]
        exclude:
          - julia-version: 1.6
            os: macos-14
```

Python and R both seem fine with just using ubuntu-22.04. Even C# uses only ubuntu-20.04, despite its Windows heritage.

I'm particularly keen to get rid of the Windows/Julia 1.6 combination, which is glacially slow (15-20 minutes, compared with <5 min for all the others). That should go away in any case later this year, when 1.6 is due to be dropped as the LTS version, but maybe Windows and Mac are now inappropriate?

I don't know the history of this choice, but perhaps it made more sense when Julia was less mature and stable, so there could be important differences between platforms.

Thoughts?

depial:
I'm unaware of the reasons for testing all the OSs, but did notice the issue with Windows a while ago (if I remember correctly it hangs longest when loading dependencies).

We did have an issue with MacOS and Julia 1.6 a few months ago which @cmcaine believed was due to 1.6 not supporting ARM versions of MacOS. There's a bit more detail in [PR #733](https://github.com/exercism/julia/pull/733) if you want to have a look.

colinleach:
> We did have an issue with MacOS and Julia 1.6 a few months ago

Presumably, that explains the `exclude:` in the testing. It still leaves me wondering why we test Mac and Windows, when other tracks don't.

I asked about this on Discord yesterday but with no replies so far.

BNAndras:
The likely answer to me is the boring one. Most tracks don’t bother testing multiple platforms since they don’t have to. Exercism exercises generally don’t go in directions where platform differences may become significant like I/O or networking. Linux is the default platform to test since most if not all test runners run Linux.

colinleach:
#793 agreed we should revisit this when Julia 1.6 was deprecated (Nov 2024). We still test 9 combinations of Julia version and OS, which isn't as bad since 1.6 went away, but is still quite slow.

[...]

PR #793, opened 2024-09-25
Simplify CI-workflow to run on Ubuntu only

colinleach:
Two relatively small changes to a GHA workflow:

- Test only on Ubuntu, as proposed in Issue #787.
- In response to an advice message that popped up while I was testing this on GH, change Julia "1.6" to "lts". Those are currently the same, but "lts" should automatically update to 1.10 in the next few months.

I should warn that I have no previous experience of GHA, so getting someone from @exercism/cross-track-maintainers to cast a skeptical eye over it would be good. I see that @ErikSchierboom is on the reviewers list, and it won't let me remove him - apologies for any disturbance!

I ran this manually on my fork, and everything passed (admittedly, not at the first attempt).

### DIFF
--- a/.github/workflows/exercise-tests.yml
+++ b/.github/workflows/exercise-tests.yml
@@ -17,16 +17,12 @@ on:
 
 jobs:
   test:
-    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-22.04
+    name: Julia ${{ matrix.julia-version }} - ${{ runs-on }}
     strategy:
       fail-fast: false
       matrix:
-        julia-version: ["1.6", "1", nightly]
-        os: [ubuntu-22.04, windows-2022, macos-14]
-        exclude:
-          - julia-version: 1.6
-            os: macos-14
+        julia-version: ["lts", "1", nightly]
 
     steps:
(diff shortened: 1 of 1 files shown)


BNAndras:
From experience, Exercism requires (or at least strongly recommends) pinning to specific versions whenever possible. That helps with reproducibility between runs. We can manually upgrade to 1.10 when it’s available. For a point release, it shouldn’t be a big deal hopefully.

ErikSchierboom:
I wonder if we should go the other way around: testing on three OS-es but only with Julia 1.6?

colinleach:
> testing on three OS-es but only with Julia 1.6?

We're caught between two separate problems here:

- The 1.6/macos combination caused problems (#733, an issue with ARM support), so is currently specifically excluded from the workflow. Erik was involved in that decision, back in May.
- The 1.6/Windows combination is spectacularly slow to run CI tests (later Julia versions are still slower on Windows than on Ubuntu, but not nearly as bad).

One tempting option is to close this PR for now, and revisit the question when 1.6 reaches end-of-life. That is expected some time around November, depending on the release of 1.11 (which is currently at rc4). Then 1.10 becomes the LTS version to test.

depial:
> One tempting option is to close this PR for now, and revisit the question when 1.6 reaches end-of-life. That is expected some time around November, depending on the release of 1.11 (which is currently at rc4). Then 1.10 becomes the LTS version to test.

I think this is a good plan since it's not too far off anyway

PR #799, opened 2024-10-02
[Conditionals] - new concept

colinleach:
This is an unfinished draft, NOT ready for formal review.

Suitable exercises to go with this might be `vehicle-purchase` or `meltdown-mitigation`, still to be decided.

As with #792, I'm putting unfinished drafts on GH while trying to get a feel for the scope of each concept.

There will be a couple more later today, but NOT the Arrays and Ranges concept. The current draft of that is getting very long, so I'll need to split it into multiple concepts: at least 2, maybe 3 (despite already excluding multidimensional arrays at this stage).


colinleach:
I think this is now ready for review. I copied the `introduction.md` from `cars-assemble` and moved the directory from `concepts.wip` to `concepts`. Slight glitch with `config.sys` conflicts (often the case when submitting multiple PRs), but I think that is now sorted.

depial (approved):
Looks good to me

colinleach:
Great, thank you!

PR #800, opened 2024-10-02
WIP draft of rational-numbers concept

colinleach:
Another incomplete draft.

The main problem with this is that I don't have an exercise to go with it. @BethanyG suggested some general ideas when we wrote the Python version earlier this year, but attention switched to stuff in support of 48in24 before we decided on an actual exercise.


colinleach:
Replaced by #895 after a merge mishap.

PR #801, opened 2024-10-02
complex-numbers concept

colinleach:
A rather self-indulgent concept: not _quite_ up there with arrays and loops as a priority concept, but I wrote the Python version with @BethanyG a few months ago, and it translated pretty smoothly to Julia.

As with rational-numbers, we discussed some vague ideas for exercises but didn't reach a decision.

It would be great if we could think of something, so we could use it for both Julia and Python. Sadly, my usual plan of stealing exercises from Elixir or Go doesn't work for the more mathematical stuff.


colinleach:
Preparing for #873 (which I'm about to start reviewing). I took your `introduction.md` and removed references, then copied the bit on `cis` and `cispi` to `about.md`. Though as I type this, I realize that I need to add links to the manual (that will be the next push).

concepts/complex-numbers/about.md
```diff
@@ -0,0 +1,275 @@
...
+
+julia> round(imag(euler), digits=15)  # round to 15 decimal places
+0.0
+```
+
+So a simple expression with three of the most important constants in nature `e`, `i` (or `j`) and `pi` gives the result `-1`.
+Some people believe this is the most beautiful result in all of mathematics.
+It dates back to around 1740.
+
+The polar `(r, θ)` notation is so useful, that there are built-in functions `cis` and `cispi` for constructing it more efficiently.
```
depial, line 196: We could add a little extra info here to help with the names of the functions. Something like:

> The polar `(r, θ)` notation is so useful, that there are built-in functions `cis` (short for `cos(x) + isin(x)`) and `cispi` (short for `cos(πx) + isin(πx)`) which can help in constructing it more efficiently.

concepts/complex-numbers/introduction.md
```diff
@@ -0,0 +1,187 @@
...
+```julia-repl
+julia> 1im * 1im == -1
+true
+```
+
+This is a simple idea, but it leads to interesting consequences.
+
+## Arithmetic
+
+All of the standard mathematical `operators`] and elementary functions used with floats and integers also work with complex numbers. A small sample:
```
depial, line 92: There is an extra `]` at the end of `operators` which needs to be deleted.

concepts/complex-numbers/introduction.md
```diff
@@ -0,0 +1,187 @@
...
+-1.0 + 1.2246467991473532e-16im
+
+julia> real(euler)
+-1.0
+
+julia> round(imag(euler), digits=15)  # round to 15 decimal places
+0.0
+```
+
+The polar `(r, θ)` notation is so useful, that there are built-in functions `cis` and `cispi` for constructing it more efficiently.
```
depial, line 162: Again, I kinda like the rewording:

> The polar (r, θ) notation is so useful, that there are built-in functions cis (short for cos(x) + isin(x)) and cispi (short for cos(πx) + isin(πx)) which can help in constructing it more efficiently.

depial (approved):
- One sync we should likely make between `about.md` and `introduction.md`
- One typo in `introduction.md`
- One suggested rephrasing in both

colinleach:
I think I addressed all your comments - thanks, they were useful.

It's 4:20 pm and my headache is proving ibuprofen-resistant. I can't see the #873 review getting finished today, though I'll add some preliminary comments.

PR #803, opened 2024-10-02
WIP draft ranges concept

colinleach:
I split this out of the earlier Arrays draft.

The best exercise I can find for it is `elixir/chessboard`. We might want to expand that a bit, but it seems like a useful starting point.

Edit: Because `go/chessboard` is quite unlike the Elixir version, we could maybe take the most suitable features of each.


colinleach:
I think this + the Chessboard exercise are now good to review. My local Git installation got corrupted somehow, so I hope I caught all the issues.

colinleach:
The issues with `config.json` are getting silly. I've moved this back to draft. Once some of the earlier concepts are merged, I'll rebase onto main and have another attempt. At the moment, there are a lot of concept PRs open, and some of them are reusing old UUIDs - all very confusing.

colinleach:
With so many earlier concepts now merged, adding this one to `config.json` is now easier. Not that that stopped me messing it up a couple of time!

concepts/ranges/introduction.md
```diff
@@ -0,0 +1,90 @@
...
+julia> typeof(0:1000)
+UnitRange{Int64}
+```
+
+Ranges are very common: not just to save you typing, but also as return types from functions.
+
+Note that ranges are _not_ vectors.
+They are just a set of instructions to generate a sequence ("lazy" evaluation, or an "iterator").
+
+If you need an vector, use the `collect()` function for conversion:
```
depial, line 21: This sentence could be made to match that in `about.md`. Here the word "vector" is used, while "array" is used there. (Previous comment about clarification still applies)

depial (approved):
Only minor nits, which need not be addressed (however "an vector" should probably at least be made "a vector"

PR #804, opened 2024-10-03
WIP draft loops concept

colinleach:
A relatively simple concept.

Possible exercises might include mixed-juices (JS + an R draft) or making-the-grade (Python). The latter may need some changes, if this concept comes before strings.

This string issue is starting to look quite common. We may need to find a way introduce string interpolation quite early in the syllabus, then go back later and deal with other the other string stuff (especially Unicode complexities).

### DIFF
--- a/concepts.wip/loops/.meta/config.json
+++ b/concepts.wip/loops/.meta/config.json
@@ -1,0 +1,7 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "contributors": [],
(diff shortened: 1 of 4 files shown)


colinleach:
I'll keep this as a WIP draft for now, to try and minimize merge conflicts. It should be possible to finish it quickly once some other concepts are merged.

depial (approved):
Short and sweet. I didn't catch any issues here

PR #805, opened 2024-10-04
WIP draft of chars concept

colinleach:
I decided there is enough in this to justify separating it from the Strings concept.

No decision yet on the corresponding exercise. I'll draft Strings first, then consider them both together.

PR #806, opened 2024-10-06
WIP draft of squeaky-clean concept exercise

colinleach:
The `introduction.md` is blank for now. Once the `chars` concept is finalized, we can copy that introduction here.

This exercise pairs with the `chars` concept, #805.

### DIFF
--- a/exercises/concept/squeaky-clean/.docs/introduction.md
+++ b/exercises/concept/squeaky-clean/.docs/introduction.md
@@ -1,0 +1,3 @@
+# Introduction
+
+_Left blank for now. Contents of `concepts/chars/introduction.md` to be copied here when finalized._
--- a/config.json
+++ b/config.json
@@ -57,6 +57,13 @@
           "basics"
         ],
         "status": "wip"
+      },
+      {
+        "slug": "squeaky-clean",
+        "name": "squeaky-clean",
+        "uuid": "7c06c4e8-4bc1-42c3-982a-f973f45193bf",
+        "concepts": [],
+        "prerequisites": []
       }
     ],
     "practice": [
--- a/exercises/concept/squeaky-clean/.docs/instructions.md
+++ b/exercises/concept/squeaky-clean/.docs/instructions.md
@@ -1,0 +1,60 @@
+# Instructions
+
+In this exercise you will implement a partial set of utility routines to help a developer clean up identifier names.
+
+In the 6 tasks you will gradually build up the functions `transform` to convert single characters and `clean` to convert strings.
+
+A valid identifier comprises zero or more letters, underscores, hyphens, question marks and emojis.
+
+If an empty string is passed to the `clean` function, an empty string should be returned.
+
+## 1. Replace any hyphens encountered with underscores
+
+Implement the `transform` function to replace any hyphens with underscores.
+
+```julia
+transform('-')  # => "_"
+```
+
+## 2. Remove all whitespace
+
+Remove all whitespace characters.
+This will include leading and trailing whitespace.
+
+```julia
+transform(' ')  # => ""
+```
+
+## 3. Convert camelCase to kebab-case
+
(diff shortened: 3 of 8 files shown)


colinleach:
Also, `.meta/design.md` is missing, mainly because I only just remembered that this is required (according to the documentation, though the linter doesn't check for it).

I'll add this later, once we get a bit clearer about the scope and limits of each concept exercise. I'm still feeling my way through working out how everything fits together in the syllabus.

depial:
I'm thinking it would be better to get rid of test 3 (camelcase to kebab) and then keep all the outputs of the functions as characters. If you wanted to replace that test with another test, we could check that the outputs are indeed characters, since I think it's worthwhile making an explicit test to explain to the student that `'a' != "a"` (for those coming from Python or Javascript).

colinleach:
Thanks for the comment. I'll go back and look at it later today.

I've been relatively absent the last couple of weeks, because of more health problems. In fact, I was waiting at Urgent Care  when your comment pinged up on my phone. Nothing life-threatening, but stuff that gets in the way of clear thinking. I have another doctor appointment upcoming on Tuesday, so I'll hope to get this sorted.

depial:
Sorry to hear about your health problems! I hope all comes well soon. Don't worry about pressing to get stuff done in the meantime. 

I was thinking that if you wanted to keep Test 3, we could make it more Julian by making camel case into lowercase (instead of kebab case), and that could satisfy the character only output without any significant change to the testset.

colinleach:
I just read through this again, months later. My first thought was that I like your suggestion about making `transform` output a `char`, and I was about to make changes. Then I realized that task 4 is incompatible with this: an empty string `""` is valid, `''` will give an error.

FWIW, 5 other tracks use this exercise, of which 4 are strongly typed with `" " != ' '` (I think Elm is an exception?). Could we argue that juggling `char` and `string` in the same function is a common real-world situation?

depial:
We could leave `transform` outputting a string.

If I remember correctly, my suggestion to output a character was potentially related to the subtleties of methods which iterate through a string character by character. For example:

```julia-repl
julia> filter(==('a'), "abcdabcda")
"aaa"

julia> filter(==("a"), "abcdabcda")
""
```
Another example:
```julia-repl
julia> findfirst("a", "xabcahliazlnkiw")
2:2

julia> findfirst('a', "xabcahliazlnkiw")
2
```
Things like this have caught me out a few times in my learning journey, so I think I wanted to hint at them, but I don't see them as directly relevant to this exercise.

exercises/concept/squeaky-clean/.docs/hints.md
```diff
@@ -0,0 +1,32 @@
...
+- [Reference documentation][char-docs] for `Char`s is here.
+- `Char` literals are enclosed in 'single quotes', `String` literals in "double quotes".
+
+## 2. Remove all whitespace
+
+- See [this method][isspace] for detecting whitespace characters and [this method][isdigit] for digits.
+
+## 3. Convert camel-case to kebab-case
+
+- See [this method][isuppercase] to check if a character is case.
```
depial, line 14: Typo: ...check if a character is *upper* case.

exercises/concept/squeaky-clean/.docs/introduction.md
```diff
@@ -0,0 +1,3 @@
+# Introduction
+
+_Left blank for now. Contents of `concepts/chars/introduction.md` to be copied here when finalized._
```
depial, line 3: Just a reminder that this is still blank

depial (approved):
Just the two comments. I'll go ahead and take care of the typo, and you can decide if you want to proceed without an intro

exercises/concept/squeaky-clean/.docs/introduction.md
colinleach: Oof ... That wasn't deliberate!  I'll add an intro by copying it from `chars`.

PR #810, opened 2024-10-09
WIP draft of nothingness concept

colinleach:
Still needs some work, but indicates the scope of this concept.

I'm probably looking at `name-badges` for the exercise, which I previously used in R -though it's quite contrived. 

Python has `restaurant-rozalynn`, but that has narrower scope (only `None`, equivalent to Julia `nothing`).


colinleach:
Oops, I didn't mean to close this, but accidentally crunched it when merging main. I'll open a new one at some point.

PR #812, opened 2024-10-11
WIP draft of pairs-and-dicts concept

colinleach:
The `inventory-management` exercise from Python might work with this.


concepts.wip/pairs-and-dicts/about.md
```diff
@@ -0,0 +1,147 @@
...
+ 'a' => 1
+ 'b' => 2
+ 'c' => 3
+
+# Each pair is a single entry
+julia> length(pv)
+3
+```
+
+A [`Dict`][dict] is superficially similar, but storage is now implemented in a way that allows fast retrieval by key, even when the number of entries grows large.
```
depial, line 51: Since you talk about something being "hashable" below, you could add, as an aside here, that the fast way is called a hash table since this would make it easier for the curious to google it. Maybe something like:

> ...in a way that allows for a fast retrieval by key, known as a "hash table", even when the number...

depial:
I took a quick look at the `inventory-management` exercise, and I think it's a great fit for the topic. Everything else also looks good!

Just a side note: I might have written `inventory-management` a little differently so people could get the chance to use the `get` method (I don't know why I like that method so much).

BethanyG:
@depial  -- as the author of both `Inventory Management` (dictionaries) and [`Mecha Munch Management`](https://github.com/exercism/python/tree/main/exercises/concept/mecha-munch-management) (_dictionary methods_) in Python, I am going to chime in here.

I, too think `dict.get()` is fabulous, but there was so much to go over in dictionaries that we moved `get()` to the follow-on exercise that covered methods.  Both exercises are very very similar, but `Mecha Much` requires the student to do a bit more thinking around nested dictionaries and other data structures within a dictionary, as well as working with creation and access methods.

**Edit:**  Just saw your note.  ;-)  These decisions are all difficult when planning a syllabus!  It is really hard not to have enormous scope creep -- especially in early concepts and data structures.

colinleach:
I don't currently have plans for `Dict` methods. I'll need to look into it, but I suspect there may be fewer in Julia than in Python. Or maybe it's just the functions vs object-method difference that makes them less visibly in Julia (plus Python being at least a decade ahead on documentation!).

depial:
@BethanyG Thanks! You've cleared up the only lingering question (about scope) I'd had about this one.

@colinleach I wouldn't say it's a high priority to do the dictionary methods concept, but it sounds interesting if not for only thinking Julia's `get!` and `pop!` add some fun functionality.

colinleach:
Let's remember that we should maybe approve/merge #809 before submitting more exercises. That would make our job easier (or at least quicker).

depial:
@colinleach I started a review on [PR #809](https://github.com/exercism/julia/pull/809) with a comment. I'm not sure if you had any other input on that though

colinleach:
> I started a review on https://github.com/exercism/julia/pull/809 with a comment

Sorry, I'm not sure what you mean. The only comment I can see in that PR is from Bethany.

No other input, so I assume silence implies consent in this context.

depial:
I've made another comment with a picture of the first review. Can you see that one?

colinleach:
I've recycled the UUID from the conditionals duplicate. The introduction may need some tweaking once inventory-management is (finally!) finished. Meanwhile, I'm interested to see if this makes any difference to the website.

I haven't yet deleted `concepts.wip/conditionals` but I think that could be a useful later step.

concepts/pairs-and-dicts/about.md
```diff
@@ -0,0 +1,147 @@
...
+julia> pd['a'] = 42
+42
+
+julia> pd
+Dict{Char, Int64} with 4 entries:
+  'a' => 42
+  'c' => 3
+  'd' => 4
+  'b' => 2
+```
```
depial, line 107: Since there was an intervening example between the definition of `pd` and here, I was initially lost with the `overwrite` example (given. I also have a tendency not to pay close attention to variable names when reading examples in literature). Maybe we could add a reminder of what `pd` is at the top of the example?

```julia-repl
julia> pd
Dict{Char, Int64} with 3 entries:
  'a' => 1
  'c' => 3
  'b' => 2

# Add
julia> pd['d'] = 4
4

# Overwrite
julia> pd['a'] = 42
42

julia> pd = Dict(pv)
Dict{Char, Int64} with 3 entries:
  'a' => 42
  'c' => 3
  'd' => 4
  'b' => 2
```

concepts/pairs-and-dicts/about.md
```diff
@@ -0,0 +1,147 @@
...
+
+To be valid in a `Dict`, the `Pair` must be a `key => value` pair, where the `key` is "hashable".
+Most importantly, this means the `key` must be _immutable_, so `Char`, `Int`, `String`, `Symbol`, and `Tuple` are all fine, but `Vector` is not allowed.
+
+If mutable keys are important to you, there is a separate but much less common [`IdDict`][iddict] type that can allow this.
+See the [manual][dict] for several other variants on the `Dict` type.
+
+### Modifying a Dict
+
+Entries can be added with a new key or overwritten with an existing key.
```
depial, line 90: I feel like this could use some commas?

Entries can be added, with a new key, or overwritten, with an existing key.

concepts/pairs-and-dicts/introduction.md
```diff
@@ -0,0 +1,137 @@
...
+julia> pd['a'] = 42
+42
+
+julia> pd
+Dict{Char, Int64} with 4 entries:
+  'a' => 42
+  'c' => 3
+  'd' => 4
+  'b' => 2
+```
```
depial, line 104: Ditto

depial (approved):
Looks good as is, but just a couple possible suggestions

colinleach:
I made the changes you suggested, plus one that was niggling me: showing the constructor `Dict('a' => 1, 'b' => 2, 'c' => 3)` instead of using a pre-existing vector. Best mention the simplest case.

colinleach:
No obvious change on the website, despite reusing the UUID. Let's see how it looks once `inventory-management` is merged.

PR #815, opened 2024-10-12
sets concept

colinleach:
There are several choices of exercise, with not much agreement among other tracks. For R, I went with [`ozans-playlist`](https://github.com/colinleach/learning-track-doodles/tree/master/r/concept-exercises/ozans-playlist). Python uses `cater-waiter`, Gleam and Java have `gotta-snatch-em-all`.


depial:
`about.md` looks great! Every time I thought of a potential question, it was taken care of shortly after. 

The only potential formatting nit is that I'm not sure the _and_ in the description of `intersect` needs italics, since no italics are used elsewhere in similar circumstances (or maybe more italics could be used elsewhere?). Either way, definitely not an issue.

As for the exercises, I think I would lean towards `cater-water`, since it looks more comprehensive and encourages the use of a wider range of the functionality listed in `about.md`.

colinleach:
The Introduction is pretty much the same as the About, as both are relatively short.

colinleach:
I'll try to pick up on `cater-waiter` next, aiming to try and simplify your current version. I _think_ I can see some potential win's, but need experimental proof.

depial (approved):
Still looks good to me

PR #816, opened 2024-10-19
Randomness concept

colinleach:
I started this because I thought porting it from Python would be a quick win. Spectacularly wrong! It took more than a week, and still isn't great.

To be fair, the way relevant functionality is spread around Julia packages is kind of a mess and needs improvement.

Pushing `.gitignore` was a mistake, but it shouldn't matter. It just leaves out any Jetbrains configuration.

I'll try to get back to more central concepts soon, though I may also submit a small Statistics concept. That overlaps with the random stuff, so I made notes on both in parallel.

PR #817, opened 2024-10-19
WIP draft of the Statistics concept

colinleach:
I have no idea how we would create an exercise to go with this. Too divisive: half the students (of the subset interested in Julia) already know this stuff and get bored with things that seem self-evident; the other half would need a semester-long course to understand it.

I'm not sure it matters much. If this gets merged, we can link to it from other Concepts, as I already did from `randomness` and will probably do from one I've not written yet (but may be called `vector-operations`).


keiravillekode:
The publicly launched
https://exercism.org/tracks/julia/concepts/vector-operations
"See the Statistics Concept for some examples."
has a broken link to
https://exercism.org/tracks/julia/concepts/statistics

colinleach:
The Introduction is identical to the About - both are short. It's now in the correct directory, with a `config.json` entry.

However, this introduces _another_ broken link, to `randomness`. Is it worth finishing #816 to fix this? That will eventually be paired with `captains-log`, when I have the time, energy and brain cells to port it. I already did an R draft, which gives me a starting point.

depial (approved):

depial:
> Is it worth finishing https://github.com/exercism/julia/pull/816 to fix this?

It looks like the `about.md` is pretty well developed, so I guess if it is not too far off, it would be good to get it out.

PR #818, opened 2024-10-21
Add cater waiter concept exercise

depial:
I've translated the python version of `cater-waiter`. It was quite involved, so please have a close look. I also translated the `introduction.md`, which I'm not sure if that could be used for in [WIP draft of sets concept](https://github.com/exercism/julia/pull/815/files)

I did stray from using proper REPL output, in favor of what gets output if it is wrapped in `println`. This was because longer collections get unwieldy otherwise.

Another point is that the doc strings for the functions may be a bit Pythonic, since I just did a quick translation. 

I'm not sure if we want to merge this or not, since it would publish the exercise to the site.

**Edit:** Ugh... I'll have a look at why all the checks are failing tomorrow


colinleach:
I'll look at this properly tomorrow. One detail I noticed is the missing `hints.md`, which has been [discussed on the forum](https://forum.exercism.org/t/concept-exercise-files/13289/7) recently. There is quite a lot or pressure to include both hints (to help mentors as well as students) and a `meta/design.md` (to prevent feature creep).

To generate the `config.sys` entry, it is easiest to use `configlet create --concept-exercise cater-waiter -a depial`, though you still need to fill in some fields. Status is "wip", concepts is whatever, prereqs are pretty much unknown at this stage and will need to be added later.

depial:
I think I've go most of this straightened out. Let me push those changes before making any new ones

colinleach:
Oops, I was trying to add a suggested change, but accidentally committed it - sorry! In any case, as you say, leaving the author blank is not an option. GH already lists you as the author, so does having this show up on Exercism make things so much worse?

I was working on this locally while you were writing your comment, so I just saw that. Some tentative comments:
- I think there are too many file imports. If `sets_categories_data.jl` is imported in `cater-waiter.jl`, also importing it in `runtests.jl` and `sets_test-data.jl` confuses things. Locally, I just got a bunch of redefinition warnings, but maybe the CI is getting more deeply confused?
- We can work on `hints.md` and `.meta/design.md` later, once we've sorted the more urgent problems.
- I was a bit surprised that you translated `introduction.md` from Python. This has to be a subset of the concept's `about.md`, so anything new you add here needs to be copied to `about.md`.

colinleach:
On file imports: I meant to point out that Python doesn't do this, it just imports specific items from the files. Also, the language seems more tolerant of overwrites than Julia (perhaps because it has no real constants).

colinleach:
I'm off to breakfast. I've not yet had my first cup of coffee, and I'm feeling it...

depial:
> To generate the config.sys entry, it is easiest to use configlet create --concept-exercise cater-waiter -a depial, though you still need to fill in some fields.

This is how I initially created it, but it seems to have broken when I added `sets` to concepts array.

>  Also, the language seems more tolerant of overwrites than Julia (perhaps because it has no real constants)

I'm pretty sure this is due to me using `const` for the constants in `sets_categories_data.jl`, so I've tried dropping them.

> I was a bit surprised that you translated introduction.md from Python.

I got a bit confused since there appears to be a nearly empty `introduction.md` in the [WIP draft of sets concept](https://github.com/exercism/julia/pull/815/files) changed files, so I went ahead and translated this.

colinleach:
> This is how I initially created it, but it seems to have broken when I added sets to concepts array.

That was my mistake yesterday, when I didn't read the error carefully enough. It's not your `config.sys`, the problem is that the linter can't find a `sets` concept.  Some options:
- Leave the "concepts" field empty for now (though the linter may also complain about that).
- Finish and merge the `sets` concept.
- Ignore the linter, and merge your exercise.

> I'm pretty sure this is due to me using const for the constants in sets_categories_data.jl, so I've tried dropping them.

Well, your choice, but I think all the includes might remain a problem. Let's find out empirically.

> here appears to be a nearly empty introduction.md

A miscommunication, though we had some earlier discussion about leaving the introduction till last, then chopping down the about to just what is needed for the exercise. That's Bethany's recommendation, and she _tends_ to be right!

colinleach:
Clearly, not enough coffee yet - I hit the wrong button...

depial:
Configlet passes now. The other error does seem to be associated with trying to include `sets_categories_data.jl` in `exemplar.jl`. The only other exercise which uses a similar setup is [alphametics](https://github.com/exercism/julia/tree/main/exercises/practice/alphametics), since it has an extra file for `permutations.jl`. However, the `example.jl` simply wrote the contents of that file directly in the the example, so it doesn't need to be included. I did this and it seems to have cleared up one problem, but now there are others. 

> A miscommunication, though we had some earlier discussion about leaving the introduction till last, then chopping down the about to just what is needed for the exercise.

Your right, I forgot about this.

Lunchtime for me, so I'll be back to look at this later

depial:
It looks like I've been on autopilot for the past day and haven't been thinking... I realized that I tried to use an outside file with helper functions for test generation with my first attempt at that [Perceptron](https://github.com/exercism/julia/pull/759) exercise, and it failed in a similar manner. The solution was to include everything in the `runtests.jl` file and wrap that in a function, so I could put the potentially spoiler code at the end.

Since the files `sets_categories_data.jl` and `sets_test_data.jl` are so large, I don't think that adding them directly into `runtests.jl` is a great option. For that reason, we might have to abandon the use of this exercise, and go with something less proprietary, or maybe I can find a way to restructure this one to have less detail.

Any thoughts?

colinleach:
Let me play with it later. I'm pretty confident we can fix this, but I need to prove it.

I'm close to getting `array-operations` completed and it would be good to get that sorted before I lose my train of thought (the risk increases with ageing, unfortunately!).

depial:
I've got it tentatively "working", presuming that the `include("sets_categories_data.jl")`  that is in the slug will provide access to that file for `runtests.jl` when a student submits their code. Effectively, I put the data from `sets_test_data.jl` at the bottom of `sets_categories_data.jl` and put both of those explicitly in the `exemplar.jl`.

If you want to play around a bit more to see if you can clean things up (as I find this to be a messy solution), feel free (at a minimum  `sets_test_data.jl` is now superfluous). In case you want to continue tinkering, I'll give you the heads-up that the issue that I couldn't overcome before (when working with other Colin) is the use of other files in `runtests.jl` (which would be the use of `sets_test_data.jl` in this case). It appears as though the inclusion of anything other than the slug is ignored.

> I'm close to getting array-operations completed and it would be good to get that sorted before I lose my train of thought (the risk increases with ageing, unfortunately!).

I've also got some other things on my plate for the next few days of this week, so I'll be more sporadic in my responses here, but I'll keep thinking to see if there's a better solution.

colinleach:
Looking back through the PRs, I realize I am 4 months overdue in responding to this. Apologies, Oct 21 was just about the time my health abruptly fell apart.

Cater Waiter is complicated, so I still can't give a quick response. I'll take a look at it, but I've come to think I need to focus on:
- Concept/Exercise pairs near the top of the tree.
- Key nodes that unlock a lot of further concepts.

In practice, I think that means:
- numbers + currency-exchange
- conditionals + something (probably vehicle-purchase?)
- arrays + elyses-enchantments
- array-operations + something
- strings + something

Playing with stuff like complex-numbers and function-composition was fun, but I have to admit they are not immediately useful in getting the syllabus launched.

As a reminder, I maintain (sporadically) a Mermaid diagram at https://github.com/colinleach/learning-track-doodles/blob/master/julia/flowchart.md. You will probably need to expand it (button with the double-ended arrow, top right)

colinleach:
I made several attempts to simplify this, with zero success - see #850. Having 677 lines of duplicate data stuffed in the top of `.meta/exemplar.jl` is less than ideal, but maybe all we can manage for now.

Two other things to discuss, both coming from the Python direction:
- Do we want all the verbose docstrings? Personally, I find them distracting, but others disagree. Students quite likely haven't reached that part of the syllabus yet.
- What do we want in `introduction.md`? You translated the Python version, I wrote one from scratch for the `sets` Concept. They are not obliged to be the same, but we should at least review the situation.

colinleach:
We'll also need to work out where this fits in the syllabus. I was wildly optimistic in putting it under `vectors`. The exemplar uses a loop and (if I'm reading it correctly) a tuple of pairs, so that's several extra prereqs. No problem, we just need to take these into account.

colinleach:
> I'm pretty sure I'll open a new PR

OK, I'd probably do the same. Note my recent comment about the editor files.

PR #819, opened 2024-10-21
Vector Operations

colinleach:
A few days ago, there was some discussion about when/how to explain broadcasting. This is one attempt to answer that question.

I suspect there are gaps (especially in talking about useful Julia functions). I'll add more content once I translate that vague feeling into something more specific.


depial (approved):
I just cleaned up two typos in `introduction.md` to match with that of the exercise. Everything else looks good!

colinleach:
Great, thank you!

PR #831, opened 2025-02-26
[Booleans] another attempt

colinleach:
Continues from the damaged #781.  I _think_ I've got it back to the state I intended, without dragging in a lot of irrelevant stuff.

Once this is merged, it should appear on the website with `annelyns-infiltration`.


concepts/booleans/about.md
```diff
@@ -0,0 +1,120 @@
...
+
+julia> false
+false
+
+julia> typeof(true)
+Bool
+```
+
+In contrast to several other languages, Julia deliberately has no concept of "truthiness".
+Only expressions which evaluate to `true` or `false` will be treated as a `Bool`.
```
depial, line 20: Very minor, but, personally, I would combine these sentences into one with an "and", since that helped me to link the idea of truthiness more directly to the following sentences about empty arrays, etc.

depial (approved):
Looks good

depial:
You can merge this at your leisure. I should mention that I'll be out of town, starting tomorrow, through the weekend, so I'm not sure I'll be able to respond until Monday

colinleach:
Great, it shows on the website.

I'll try to stack up a few PRs for early-stage syllabus components by the time you get back.

PR #832, opened 2025-02-27
Cars assemble

colinleach:
Pairs with the `conditionals` Concept.

I had planned to use `vehicle-purchase`, but then found that relies heavily on string interpolation - not introduced until later in the syllabus.


exercises/concept/cars-assemble/.docs/instructions.md
```diff
@@ -0,0 +1,50 @@
...
+The production increases linearly with the speed. 
+So with the speed set to `4`, it should produce `4 * 221 = 884` cars per hour. 
+However, higher speeds increase the likelihood that faulty cars are produced, which then have to be discarded. 
+
+You have three tasks.
+Each of the required functions takes a single integer parameter, the speed of the assembly line.
+
+## 1. Calculate the success rate
+
+Implement the `success_rate()` method to calculate the ratio of an item being created without error for a given speed.
```
depial, line 16: Maybe "probability" would be more fitting in this sentence instead of "ratio".

exercises/concept/cars-assemble/cars-assemble.jl
```diff
@@ -0,0 +1,8 @@
+function success_rate(speed)
+end
+
+function production_rate_per_hour(speed)
+end
+
+function working_items_per_minute(speed)
+end
```
depial, line 8: I've been including a space between the function name and the `end` in the slugs so students can start coding without needing to create a new line. For example:

```julia
function success_rate(speed)

end
```

This choice is arbitrary, but I figure we should pick a standard. What do you prefer?

depial (approved):

exercises/concept/cars-assemble/cars-assemble.jl
colinleach: Yes, I was starting to get uncomfortable with this inconsistency. There is no agreed standard, but we can create our own. I'll go though this and other open exercise PRs to add a blank line.

I've heard of other maintainers trying to insist that it's more educational to leave the stubs out, presenting the student with a blank page to start coding in. Jeremy and Erik were vocally unhappy about this.

depial: I figure it might be good to be consistent for the concept exercises.

That said, I do like having a variety of starting points for practice exercises (e.g. blank pages, function names, etc), which I think can help build a student's flexibility / reduce complacency.

PR #840, opened 2025-03-06
Update config.json

depial:
Change concept exercises names to capitals and spaces. This follows the naming convention seen in practice exercises as well as other tracks (eg. Python).


colinleach (approved):
Oof, I really messed that up. Thank you for fixing it!

PR #841, opened 2025-03-07
Update config.json

depial:
Add `vector-operations` concept. I'm not sure if there is an automatic way to do this, so I used configlet to create a new UUID, then I filled in the three fields (`"uuid"`, `"slug"` and `"name"`) manually.

If this is a legitimate way to edit this file, then I think we can try this to get the concept onto the syllabus on the website.


depial:
Never mind. I've just seen the comments on the other PR, so I'll be closing this one.

PR #842, opened 2025-03-07
Vector operations (rescue attempt)

colinleach:
After an unfortunate merge mishap with #819, this PR attempts to remedy the situation.

### DIFF
--- a/concepts/vector-operations/.meta/config.json
+++ b/concepts/vector-operations/.meta/config.json
@@ -1,0 +1,8 @@
+{
+  "authors": [
+    "colinleach",
+    "depial"
+  ],
+  "contributors": [],
+  "blurb": "Julia provides many ways to operate on an array, either as a unit or element-wise."
+}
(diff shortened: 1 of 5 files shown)


depial (approved):
Looks good!

depial:
It looks like `Vector operations` appeared on the syllabus on the website :)

colinleach:
I am VERY relieved! Thank you for spotting the problem with the previous attempt (though I see you still don't want official recognition of this - OK, I'll respect your choice).

I'll aim to get `function-composition` out later today. I looked more closely at this, and it appears that the code is already merged into main. My mistake was failing to move it from `concepts.wip/` to `concepts/` and add it to the top-level config.

Embarrassing, but it should be fixable.

colinleach:
Trying to work out the next priorities, my current thinking is:
- `strings`, which is already ~80% written. I expect at least 10 further concepts to have this as a direct or indirect prereq.
- 'loops`, which is currently relegated to the bottom of my draft plan but I keep tripping over other exercises that need it.

PR #843, opened 2025-03-07
Function comp draft

colinleach:
I think/hope this should fix it. Fingers crossed...


colinleach:
Forget it. I unintentionally pushed to an old branch (not yet sure why).

PR #844, opened 2025-03-07
function-composition (rescue attempt)

colinleach:
OK, this is what I intended to do a few minutes ago. _Soooo_ many branches...


depial (approved):

depial:
At first, when I tried to merge this, it threw an error about the `json` finding an unexpected character. I tried to abort the merge, but it looks like it went through anyway...

colinleach:
That's very strange. I didn't get any linter issues.

Unfortunately, it's not showing up on the website.

colinleach:
Exercism doesn't seem to be seeing it at all. As well as the syllabus, my rep is showing nothing new sine `vector-operations`.

One step forward, one step back...

depial:
I'm wondering if it could be because the prerequisites for the exercise include `strings`... Should we try modifying that json?

colinleach:
Worth a try, I haven't got any better ideas. I assume `strings` really should be a prereq, so we can add it back in later.

PR #845, opened 2025-03-07
Update config.json

depial:
Delete strings from prerequisites of high-school-sweetheart

### DIFF
--- a/config.json
+++ b/config.json
@@ -113,8 +113,7 @@
           "function-composition"
         ],
         "prerequisites": [
-          "functions",
-          "strings"
+          "functions"
         ],
         "status": "wip"
       },


colinleach (approved):

colinleach:
I didn't see any issues with stray characters.

colinleach:
And it's on the website! Floating free, in splendid isolation.

depial:
That's good!

I suppose it'll get better anchored when we get strings up and that prereq back in there

depial:
It also says I've mastered the topic, so everything seems to be checking out.

We just have to figure out what the issue is with the other ones...

colinleach:
We probably also need to get `functions` in there. I looked at whether I could get that done quickly, but I'm hitting all sorts of other missing prereqs with the exercises. Classic dependency hell, and I'm not ruling out cyclic dependencies. All fixable, but it slows things down.

PR #847, opened 2025-03-21
Strings concept

colinleach:
Finally, something to PR (I started this several months ago).

This is a big concept, and many other parts of the syllabus have it as a prereq. Please look at it with a skeptical eye, and feel uninhibited in making criticisms.

I'll aim to get `log-levels` submitted later today, to go with this. I'll probably try to update `chars` (#805) and `squeaky-clean` (#806) over the weekend, because strings and chars are so closely related. Also, I find this part of the syllabus more of a slog to write than the mathematical stuff, so I'd like to get it finished then move on.

PR #848, opened 2025-03-21
log-levels concept exercise

colinleach:
Pairs with `strings` concept #847.

It's a popular exercise for this concept (7 other tracks use it), though maybe it is rather small and simple for the topic? Anyway, it's the best I've come up with at this stage.

### DIFF
--- a/exercises/concept/log-levels/.docs/instructions.md
+++ b/exercises/concept/log-levels/.docs/instructions.md
@@ -1,0 +1,52 @@
+# Instructions
+
+In this exercise you'll be processing log-lines.
+
+Each log line is a string formatted as follows: `"[<LEVEL>]: <MESSAGE>"`.
+
+There are three different log levels:
+
+- `INFO`
+- `WARNING`
+- `ERROR`
+
+You have three tasks, each of which will take a log line and ask you to do something with it.
+
+## 1. Get message from a log line
+
+Implement the `message` function to return a log line's message:
+
+```julia-repl
+message("[ERROR]: Invalid operation")
+# => "Invalid operation"
+```
+
+Any leading or trailing white space should be removed:
+
+```julia-repl
+message("[WARNING]:  Disk almost full\r\n")
+# => "Disk almost full"
+```
+
+## 2. Get log level from a log line
+
+Implement the `log_level` function to return a log line's log level, which should be returned in lowercase:
+
+```julia-repl
+log_level("[ERROR]: Invalid operation")
+# => "error"
+```
+
+## 3. Reformat a log line
+
+Implement the `reformat` function that reformats the log line, putting the message first and the log level after it in parentheses:
+
+```julia-repl
+reformat("[INFO]: Operation completed")
+# => "Operation completed (info)"
+```
+
+## Note
+
+All strings in this exercise are in English and limited to the ASCII character set.
+Later Concepts will give a chance to work with Unicode characters.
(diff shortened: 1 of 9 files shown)


colinleach:
I just noticed: if we get this merged, we can set it as a prereq for `high-school-sweetheart` and connect that to the rest of the syllabus (without waiting for `dicts-and-pairs` plus `functions`).

depial (approved):
Looks good to me!

depial:
> It's a popular exercise for this concept (7 other tracks use it), though maybe it is rather small and simple for the topic? Anyway, it's the best I've come up with at this stage.

It does seem pretty simple, but it's pretty early in the syllabus, and it still manages to touch on the most common methods, so I don't see a problem with that. That said, if you have any thoughts on expanding it, I wouldn't be adverse to doing that.

depial:
If you would like to link it now by changing the prereqs, it would clean up the syllabus, but I don't think it matters too much and, if it's just a few days/weeks, waiting for when `functions` goes up doesn't seem unreasonable. I'm fine either way, so it's up to you :)

exercises/concept/log-levels/.docs/instructions.md
```diff
@@ -0,0 +1,52 @@
...
+You have three tasks, each of which will take a log line and ask you to do something with it.
+
+## 1. Get message from a log line
+
+Implement the `message` function to return a log line's message:
+
+```julia-repl
+message("[ERROR]: Invalid operation")
+# => "Invalid operation"
+```
```
depial, line 22: I noticed this needed the Julia repl formatting.

exercises/concept/log-levels/.docs/instructions.md
```diff
@@ -0,0 +1,52 @@
...
+## 3. Reformat a log line
+
+Implement the `reformat` function that reformats the log line, putting the message first and the log level after it in parentheses:
+
+```julia-repl
+reformat("[INFO]: Operation completed")
+# => "Operation completed (info)"
+```
+
+## Note
```
depial, line 49: This note shows up as `Task 4` on the website, so it's a bit confusing. Maybe it could be switched to:

```
### Note
```
to avoid the auto-format?

depial (commented):
After doing the exercise on the website I noticed that a note was formatted as a task (and the repl formatting is R). I just wanted to leave these notes for when we get around to touching this up. Is there a better space for notes of this kind?

depial:
I also agree that the REPL output should be made more concise when possible. I get the following example in VSCode:

```julia-repl
julia> collect(1:4)[2]
2
```
Where the output does not have any indicator (such as =>). As you mentioned, with a more complex output, we get something like:
```julia-repl
julia> collect(1:4)
4-element Vector{Int64}:
 1
 2
 3
 4
```
Which I think can be simplified to something like:
```julia-repl
julia> collect(1:4)
[1, 2, 3, 4]
```
I guess the question is to whether we should have a cursor indicate output or not. I don't mind having nothing there, while I'm also slightly adverse to using `=>`, since this is meaningful in Julia in other contexts. If you would like to include something, though, how about something like `>>>`?
```julia-repl
julia> collect(1:4)
>>>  [1, 2, 3, 4]
```
I'm not sure if that is appropriate or not though. Any thoughts?

As to the note, I get it when doing the exercise under the instructions tab:
image

I think the horizontal rule with a note below you've suggested sounds like a good first option to try to remedy this.

PR #850, opened 2025-03-22
Cater waiter 2 - FAILING

colinleach:
This is pretty much a disaster. I added the "editor" list to `.meta/config.json` to tell it about the extra files, I turned to data files into modules and exported all the things `runtests.jl` needs (the same lists that the Python version imports), but CI still refuses to find the module.

I'll PR this for documentation purposes, but then close it.

I'm starting to think there must be something wrong with the way our GitHub Actions are set up, but I haven't found the problem. I looked at the equivalent Python GHA but that does everything very differently. I'll compare some other tracks at some point, but not today (too fed up with it!).


depial:
I also can't figure it out, so I think this is beyond my current level of understanding of the website. The curious thing is that we can include extra files that can be used by the student if we want to offer functionality beyond Base (e.g. [Alphametics](https://github.com/exercism/julia/tree/main/exercises/practice/alphametics) makes `permutations.jl` available to the student).

PR #853, opened 2025-03-24
add cater-waiter concept exercise

depial:
A second attempt at adding cater-waiter. Previous is [PR #818](https://github.com/exercism/julia/pull/818/files)

I haven't included `sets_test_data.jl` since I believe it was superfluous with the current setup putting everything in `exemplar.jl`, but we'll see if the checks pass.

### DIFF
--- a/config.json
+++ b/config.json
@@ -155,10 +155,11 @@
       },
       {
         "slug": "cater-waiter",
-        "name": "cater-waiter",
+        "name": "Cater Waiter",
         "uuid": "61e3bb88-a9a5-4978-bc55-1d87de9f2b70",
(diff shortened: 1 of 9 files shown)


colinleach (approved):

colinleach:
I think the prereq will be loops, currently awaiting review. We can update the `config.json` once that is merged.

depial:
I've realized that I haven't added a `design.md` to this. That's needed now, right?

colinleach:
Technically, it's required - mainly in an attempt to limit feature creep in future. Worth adding at some point, but I don't think it's urgent.

PR #854, opened 2025-03-25
Fix various formatting glitches

colinleach:
Addresses comments made in #848, plus a few errors I found during the read-through.

I also ran `configlet fmt -u -y` to get rid of some warning messages.


depial (approved):
I updated one file, but I'm not sure if it was left that way intentionally. If so, let me know and we can revert the changes. Otherwise, this looks good to me.

colinleach:
I like your changes.

depial:
I've had a look at the online editor and it looks like the note has been placed at the bottom of `Task 3`, rather than outside of it... I don't think that's a bad location, but I also assume it is not the intended one.

colinleach:
Not intended, but I'm out of ideas on how to fix it. Another non-urgent issue to think about later?

depial:
Agreed. I'll keep my eyes open to see if I stumble on a solution in the meantime.

Issue #846, opened 2025-03-11
Learning Track Snags


depial:
There are a couple issues with the syllabus on the learning track, so I think it would be good to catalog them here.

- [ ] Booleans: I was able to master the concept, however, the exercise is still listed as `Available` instead of `Published`.
- [ ] Vectors: I can't unlock this concept, and when hovering over the concept on the syllabus, it reads `Learn to unlock this concept.` when it should say `Learn Conditionals to unlock this concept`. Also, the concept page shows it has 0 exercises, even while there is an exercise listed as `Published`.

One thing I've noticed is that `Annalys Infiltration` (exercise for Booleans) is a preexisting exercise, and when I looked at it in [concept.wip](https://github.com/exercism/julia/tree/main/exercises/concept.wip/annalyns-infiltration), the stub name is `game.jl` instead of `annalyns-infiltration.jl`, as it is in [concept](https://github.com/exercism/julia/tree/main/exercises/concept/annalyns-infiltration). It is `game.jl` which shows up on the website when trying to complete the exercise, so this may be why my exercise is still showing as `Available`.

colinleach:
Thanks for documenting that. I haven't had chance to look at it today (picking up a guest from Phoenix airport, so I spent quite a few hours on I-10). I'll get back to you...

depial:
No worries! Have a fun time with your guest! 

It looks like my guests will be here next week, so I won't have much time then either.

colinleach:
> One thing I've noticed is that Annalyns Infiltration (exercise for Booleans) is a preexisting exercise, and when I looked at it in [concept.wip](https://github.com/exercism/julia/tree/main/exercises/concept.wip/annalyns-infiltration), the slug name is game.jl instead of annalyns-infiltration.jl

Also, the Practice page lists both `Annalyn's Infiltration` and `Annalyns Infiltration` (no apostrophe).

I think we're going to have to delete `julia/exercises/concept.wip/annalyns-infiltration` to untangle this. It makes everyone nervous (the GH-Website connection is fairly brittle), but things are so broken already that I believe it's now worth the risk. We can revert if necessary. Should I PR the change?

> Vectors: I can't unlock this concept, and when hovering over the concept on the syllabus, it reads Learn to unlock this concept

I can reproduce this.

> the concept page shows it has 0 exercises

This problem seems to have gone away, as mysteriously as it arrived.

colinleach:
[...]

PR #859, opened 2025-04-02
Update config.json

depial:
"status" for Elyse's Enchantmens, Annalyn's Infiltration and Lasagna set to "deprecated" in `concepts.wip` in an effort to address [Issue #846](https://github.com/exercism/julia/issues/846)

### DIFF
--- a/concepts.wip/config.json
+++ b/concepts.wip/config.json
@@ -11,7 +11,7 @@
           "comments"
         ],
         "prerequisites": [],
-        "status": "wip"
+        "status": "deprecated"
       },
       {
         "uuid": "84480c18-327e-47ea-94e7-9e14f2146643",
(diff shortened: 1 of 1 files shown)


colinleach (approved):

colinleach:
Let's try this for starters. My guess is that we will also need to change the slugs (e.g. from `x` to `old-x`), but it's probably safest to do one step at a time.

depial:
I've seen that the stub in the online editor for the exercise linked to the syllabus now reads `elyses-enchantments.jl` which is forward progress :)

depial:
I was really expecting to have those duplicates drop from the site... If deprecating them doesn't get rid of them, I'm not sure what else we are going to do. Maybe the site needs more time to update? I've put in the PR for the simple evaluation order switch in the meantime though.

depial:
> so long as we never, ever mess with the UUIDs

And just before you wrote that, I was considering what would happen if we swapped them between the exercises... :D

PR #860, opened 2025-04-02
Update eachexercise.jl

depial:
Changing the order of evaluation of "concept.wip" exercises and "concept" exercises in `eachexercise.jl` as an attempted fix for Issue #846


colinleach (approved):

colinleach:
🤞

colinleach:
Certainly not an instant change on the website!

depial:
I guess this change is more to do with the test runner rather than the site, so the way things are connected on the website might be deeper. We can try to give it some time to update anyway.

PR #861, opened 2025-04-02
Update mixed-juices.jl

depial:
Added missing `function`s to first two function definitions.


colinleach (approved):
Oops, I messed that up.

PR #862, opened 2025-04-02
emptied concepts and prerequisites fields

colinleach:
This is step 1. We can set status to deprecated in a later step (assuming this goes through without breaking anything).


depial (approved):

colinleach:
No change, which I'm OK with. At least nothing broke!

PR #863, opened 2025-04-02
Deprecate concept exercises

colinleach:
set everything to "status": "deprecated". Also cleaned up Lasagna, which I missed earlier.

### DIFF
--- a/concepts.wip/config.json
+++ b/concepts.wip/config.json
@@ -5,7 +5,11 @@
         "uuid": "39ebdd04-4f84-4817-bf9d-f1f9e066c283",
         "name": "Lasagna",
         "slug": "lasagna",
-        "concepts": [],
+        "concepts": [
+          "functions",
+          "integer-introduction",
+          "comments"
+        ],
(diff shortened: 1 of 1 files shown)


depial (approved):

depial:
Hopefully we see some progress tomorrow!

colinleach:
I'm trying to feel that optimism, though the website seems quite resistant to change.

depial:
> I'm trying to feel that optimism, though the website seems quite resistant to change.

I saw that you were able to repurpose a UUID [here](https://forum.exercism.org/t/is-there-a-way-to-deprecate-old-julia-concepts/15999/2). If the site won't budge, do you think it would it be possible to swap the two UUIDs between the duplicate exercises?

colinleach:
Good idea! I actually did more repurposing than that, but things got tangled up (and _very_ confusing) with a few exercises, so they were unintentionally duplicated.

depial:
I'm not seeing any changes on the website...

PR #864, opened 2025-04-03
Swap uuids

depial:
Swapping the UUIDs between duplicate the duplicate exercise `annalyns-infiltration` and the concept `conditionals`

### DIFF
--- a/concepts.wip/config.json
+++ b/concepts.wip/config.json
@@ -26,7 +26,7 @@
         "status": "deprecated"
       },
       {
-        "uuid": "05cfc609-3a14-479a-b6fa-6d2a53ad111c",
+        "uuid": "603c7f88-5250-427e-b5d8-f8306fdd9b03",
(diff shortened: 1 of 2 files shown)


colinleach (approved):

depial:
I think we should revert the changes. It created a new `Booleans` concept which is free floating (both of which now show two Annalyns Infiltrations exercises available, but both of which still point to the old one). Also, nothing happened with `Vectors`

depial:
Strangely, it says that Annalyns Infiltration has been updated, and when I update it, all the tests fail, I assume because the `game.jl` stub is still the same, but the testrunner is now looking for the `annalyns-infiltration.jl` stub.

depial:
Let's revert these changes and try de-duping the slugs, to see how that goes

colinleach:
To change the slug, I think we'll need to change the `config.json` entry, and also the directory name. Am I missing anything?

PR #865, opened 2025-04-03
Revert "Swap uuids"

depial:
Reverts exercism/julia#864

### DIFF
--- a/concepts.wip/config.json
+++ b/concepts.wip/config.json
@@ -26,7 +26,7 @@
         "status": "deprecated"
       },
       {
-        "uuid": "603c7f88-5250-427e-b5d8-f8306fdd9b03",
+        "uuid": "05cfc609-3a14-479a-b6fa-6d2a53ad111c",
         "name": "Annalyn's Infiltration",
         "slug": "annalyns-infiltration",
         "concepts": [],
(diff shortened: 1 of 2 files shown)


depial:
Actually, don't accept this yet, I'll also change the slugs.

colinleach (approved):

depial:
No worries, I don't think I could have made all the changes here anyway. Feel free to do the next PR if you'd like. I do think we'd have to change the slug and the directory.

I'm not sure if it would be better to prefix the `concept.wip` stuff with `old-` or to prefix the `concept` stuff with `new-`. I'll leave that up to you if you would like to make the PR

depial:
One thing to consider beforehand is how changing the UUIDs got the runner to see the right exercise, but the website still used the wrong stub. So maybe we should modify the old slugs with a prefix, which might force the site to look elsewhere?

I'm unsure what will happen

depial:
Also, bad news. The website hasn't reverted as far as I can see, so these exercises are now broken for the time being :(

colinleach:
I'm leaning towards changing the old exercise slugs. Give me half an hour or so, I have some non-Exercism stuff to sort out.

PR #866, opened 2025-04-03
Change slug

colinleach:
added "old" to Annalyn's Infiltration


depial (approved):
Let's hope this works

colinleach:
Not looking great so far

depial:
No, not good. The stub must be cached somewhere on the site, so maybe we need to wait for an update?

colinleach:
If nothing improves on the website in the next few hours, what's left?

- Maybe remove `concept.wip` from `eachexercise.jl`
- The nuclear option: delete the `old-annalyns-infiltration` directory completely
- Decide it's above our pay grade and focus on new content while somebody else fixes the legacy stuff

colinleach:
One (small) bit of good news. In the Annalyn's Infiltration online editor, I did a revert to start. That removed my old entry in `game.jl`, and on refresh I was given an `annalyns-infiltration.jl` stub. Added the solution, all tests passed.

I'm curious what a non-maintainer would see. Probably nothing, at this wip stage, but beyond that probably less visual noise.

depial:
I was able to reproduce that. I wonder if that would have gone through earlier with just the UUID swap? 

I also see that before the slug name change, there were two exercises listed, but now it lists `1 exercise` under `Booleans`  (although there are still two links to the exercise)

image

PR #867, opened 2025-04-03
Move deprecation

colinleach:
I'm not sure whether we should merge this, and certainly not today. I experimented with deprecating old-annalyns-infiltration and its UUID at the top level. It didn't go entirely smoothly: the linter insists on the full exercise files being in `exercises/concept` even for _deprecated_ exercises.

Maybe worth a gamble to untangle things, if we have no better ideas?

Along the way, I found out why Lasagna works: both old and new versions have the same UUID. I _think_ that means we could delete the one in `concept.wip/`????


depial:
I'm okay on waiting on this for now.

> Along the way, I found out why Lasagna works: both old and new versions have the same UUID. 

This is interesting. I can't see why deleting how deleting one would break things, but I'm not really in a good position to say ;) From now on, should we try just duplicating the UUIDs when we use a preexisting concept instead of creating new ones? Maybe this is the expected procedure?

In other news, the website doesn't seem to have changed overnight...

colinleach:
On a related subject, these are the UUIDs associated with `elyses-enchantments`:

"185683d8-96f7-4d51-8726-56e18b6ef0c7"  (top level)
"234466a0-08ce-45a1-aebc-5ec0b0cde24f"  (in concept.wip)

If getting rid of a `conditionals` duplicate works as we hope, I'd like to do the same with the "234466a0-08ce-45a1-aebc-5ec0b0cde24f" UUID.

> From now on, should we try just duplicating the UUIDs when we use a preexisting concept instead of creating new ones?

Mostly, this is what I've been doing (plus removing the original exercise files). A couple of the early ones got away from me, when I started working on the legacy stuff last September. We're still cleaning up the mess that caused.

depial:
> Landing page of the website just now

At least it doesn't show up in the syllabus, which is more than I can say for the two `Booleans`... :D

> Does that make sense?

Yep! (at least I hope so)

depial:
If we have to move it to deprecate it, I suppose we should try.

I'm only worried that we'll somehow end up with another duplicate somehow...

Another possible action: Could we try deprecating an existing exercise in `exercises\concept`, then return it to wip? Or is deprecation a point of no return? If it's a switch we can throw, we could at least test the effects beforehand, but, at the same time, I hesitate to mess with something that's not broken :D

colinleach:
> is deprecation a point of no return?

I don't know, but I think it might be.

> I hesitate to mess with something that's not broken

Agreed! This would make me very nervous.

> we'll somehow end up with another duplicate

Perfectly understandable paranoia (given recent experiences), but I don't think this is very likely. Either way, we end up having to ask somebody more senior (probably Jeremy) to sort it out for us.

depial (approved):

depial:
> Success!

I think it just reset the exercise, so we'll have to solve it again, but it looks good so far!

colinleach:
Yes, I just solved it again and it's looking good.

That just leaves Vectors. I don't have any good suggestions for that. Is it time for another forum post, to get the other maintainers involved?

depial:
> Is it time for another forum post, to get the other maintainers involved?

Possibly :)

I know there is a preexisting Conditionals concept, but was there something else strange with Elyse's Enchantments? I feel like the former is what is keeping the concept from linking, while the latter could be what is resulting `0 exercises` appearing. Of course, it could all be down to the Conditionals confusion.

Do we have any methods at our disposal?

depial:
I'm fairly convince it's got to be the Conditionals duplicate. Could we reuse the UUID of the old concept with a new concept, and delete the old conditionals entry in the `concepts.wip\config.json`?

colinleach:
I've no easy answers and need time (and energy) to think about this. Right now, I'm confused about which UUIDs are already recycled, which are duplicated (e,g, Lasagna). I need to sort this out in my head.

> Could we reuse the UUID of the old concept with a new concept, and delete the old conditionals entry in the concepts.wip\config.json?

We'll probably end up doing this.

depial:
> We'll probably end up doing this.

I figure this is our best shot at fixing things. Maybe we could try renaming the dupe? I figure if we try to stick with what just worked for `Annalyn's Infiltration`, we might have a shot.

colinleach:
After a bit more thought (probably not enough):

- I'm in favor of reusing the UUID of the conditionals dupe. I don't think renaming the slug will be enough - we hit success with annelyns-infiltration after moving to `exercises/config` and marking deprecated, which (I think) only works for exercises.
- We could repeat that deprecation trick with the `elyses-enchantments` dupe. That still has its own UUID.
- In `concept.wip`, both `lasagna` and `old-annalyns-infiltration` are now redundant. We already reused those UUIDs at the top level, so the dupe should be deleted. Should I PR that?
- I'll check the unmerged PRs to see if there are any we can use to recycle UUIDs. My first thought was `functions`, which is the next one on my list, but that's already committed to a `functions` dupe.
- I'll do a more systematic search for what UUIDs are used where. Old habits die hard, and I still like having these sort of lists in PostgreSQL for searching (overkill, obviously, but it's harmless and makes an old man happy).

colinleach:
Also, I'll post an update on the forum. Some people might be interested, and (other than Jeremy and his international peregrinations) it may be easier to get attention before Bootcamp pert 2 starts April 26.

depial:
> We could repeat that deprecation trick with the `elyses-enchantments` dupe. 

That seems like it could be beneficial (and I assume neutral a worst).

> Should I PR that?

At present, I don't see a problem with deleting these if they are using the same UUIDs as the top-level exercises.

> Also, I'll post an update on the forum.

Sounds good. Try to see if it's possible to deprecate a concept. If not, I'm starting to think we'll need help sorting out the `Vectors` / `Conditionals` connection issue.

PR #868, opened 2025-04-04
Strange stopwatch

depial:
This is a preliminary look at an exercise for complex numbers, and should probably be labeled a draft. A few points:

- `introduction.md` is left blank to be able to align it later with the concept intro. For that reason, I'm unsure on how much further detail (mathematical or functional) will be necessary in `instructions.md`.
- I've generally made the functions accept real arguments instead of complex ones. This allows more inbuilt methods to be used (e.g. `complex(a, b)`), while keeping touch with the "real world" example of the 2D plane. This might also make porting to other languages easier if desired.
- I've left the `timearg` function returning radians rather than seconds for generality. This could be made more specific with the demand of conversion to seconds.
- This feels fairly math intense for those who are not inclined, but much of the exercise is focused on using inbuilt functionality rather than focusing on the mathematics of complex numbers themselves.
- Prerequisites include: Numbers, strings and function composition

I'm looking for any suggestions on how to improve this. It turned out a bit differently than I had initially imagined it due to hiccups along the way, but the basic idea is still there.


colinleach:
> This feels fairly math intense for those who are not inclined

Pretty much inevitable for this sort of concept on this sort of track.

> Prerequisites include: Numbers, strings and function composition

We can put this well down the concept tree as a relatively advanced topic. My main concern is that it should be a leaf node, with nothing else depending on it. That way, students aren't penalized much if they want to skip it. 

I'll go through it in more detail later today. Meanwhile, non-urgent question for @BethanyG: how would you feel about this as something to port to the Python syllabus?

colinleach:
> I'll go through it in more detail later today

OK, definitely tomorrow at the latest. I've had inventory-management partly done for so long that I wanted to finally PR something. Quietly determination to make this syllabus join up, with nothing free-floating....

depial:
Some added comments on the rationality behind the choice of functions and structure of the exercise:

- A main constraint was avoiding the introduction of modular arithmetic associated with the addition/subtraction of time, since this seems an unnecessary complication that is beyond the scope of the topic.
- For the above reason, `rotate` was left general, in the sense that it is tested on vectors of non-integer size. This was to leave the concept easily generalizable to real world applications.
- The same can be said of `timearg` which I mentioned in the initial comment.
- `readtime` seems necessary for the backstory while giving an opening for seeing the relationship between the magnitude and argument of a complex number. I used a string output so `Strings` could use that as a prereq, thereby allowing this to appear lower in the syllabus. However, this could easily be changed to output a tuple as with most other functions.
- `getcoords` is effectively the inverse of `readtime`, so it helps close the loop while encouraging students to employ the polar form of complex numbers.
- The first two functions can, and should, be used in the last two (as seen in the `exemplar.jl`). 
- I've left the functions in `exemplar.jl` as one-liners in an effort to show the power and elegance of using complex numbers + Julia functionality for this task. However, these could be expanded out if they are too dense.

BethanyG:
> I'll go through it in more detail later today. Meanwhile, non-urgent question for @BethanyG: how would you feel about this as something to port to the Python syllabus?

I have to be really honest here.  I am not feeling this.  I think my "practicality brain" is resisting.  It feels contrived to me, and also has overlaps with things like [split second stopwatch](https://github.com/exercism/csharp/pull/2403) which confuses things.  I know this is about dealing more cleanly with rotations, but it doesn't quite have the motivation that a "Big Circle" navigation or "Translate-Transform Image" problem might.  Because complex numbers are "mathy" and "weird", I feel like (_at least for Python_) folx need to have both a strong reason to engage in solving the problem, and also a sense that their "mathy" and "weird" intuition might be off.  Otherwise, they'll just move on.

This could be because there isn't a story yet -- and that a story about animating a clock might help here.  But I also think we're adding a bunch of time and datetime exercises to problem specs right now, and so should avoid time-themed problems that **_aren't_** related, to well _time_ and _dates_.

It also lacks a bunch of things that I'd cover for Python (_at least in a concept exercise_) -- since in Python this is a built-in type, and is subject to all the "width" rules/conversion rules, and many of the basic mathematical operations.  My overriding goal in Python would be to get students "used to" treating this an any other numeric type - granted, its a bit special (_but so are Fractions and Decimals - in fact, they are much more special because they aren't core types_) and a bit off the beaten path -- but it is in no way super-exotic.

So something that encouraged/forced using addition, subtraction, multiplication, et., al. -- and even mixed math with other types.

I think I would also cover some select things in [cmath](https://docs.python.org/3/library/cmath.html#module-cmath) vs [math](https://docs.python.org/3/library/math.html), to help folx understand when you can "simply" use a complex number in a calculation, and when you have to deviate or apply alternate methods.

Sorry to be so long-winded!

colinleach:
Thanks for the detailed comments. I have some of the same feelings (somewhat attenuated for Julia, given the user base), but I don't have anything better.

I've not been following the (interminable) discussions are new cross-track problem specs, so it's useful to know what we might overlap with.

This is hard!!

depial:
@BethanyG Thanks for your views! 

I agree the story is contrived, so I'll be looking into other possible ideas. A couple questions on your thoughts:
- I had avoided an image rotation exercise because I believe there are cropping/filling issues at the corners for arbitrary rotations. Is there a way around this? 
- When you mentioned "Big Circle" navigation, does this refer to the stereographic projection of a great circles from a sphere onto circles on the complex plane? If so, that could be an interesting avenue to explore :)

@colinleach Judging from the rest of Bethany's comments, I'm not sure a Julia exercise in the mold I'm currently considering would port well to Python since the user base is generally more mathematical and: 

> "Julia includes predefined types for both complex and rational numbers, and supports all the standard [Mathematical Operations and Elementary Functions](https://docs.julialang.org/en/v1/manual/mathematical-operations/#Mathematical-Operations-and-Elementary-Functions) on them. [Conversion and Promotion](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#conversion-and-promotion) are defined so that operations on any combination of predefined numeric types, whether primitive or composite, behave as expected."

Which why I focused more on exposing users to the inbuilt methods to advance the concept. Of course, I'm new to this, so I'm not sure if my idea really constitutes a concept exercise or not. I was just running with the idea of vector rotation since it seems to be the simplest possible example. Let me know if you think I'm off base here and we should change tack.

colinleach:
Yes, we have two not-quite-identical problems here. Our first priority is to get something on the Julia syllabus. However, it would be a huge bonus if it could be ported to Python (quickly) and R (at some future point). There may also be other tracks, I'm not sure.

As I said, this is hard.

colinleach:
A question which I can't currently decide on the answer to: would this maybe work better as a track-specific practice exercise? That presupposes that we can find another concept exercise that also works for Python. Incidentally, if I seem to be pushing that too hard: I wrote Python's complex concept (and R's), so I have some personal interest in finding an exercise that works with all of them.

It's very unfortunate that I am ***much*** worse at thinking of new exercises than I am at drafting concept docs! We all have different strengths and weaknesses.

depial:
> would this maybe work better as a track-specific practice exercise?

I could certainly see it as such. I will get away from the time aspect since, as Bethany noted, it would help separate it from other exercises. It could also complement the [`complex-numbers`](https://github.com/exercism/julia/tree/main/exercises/practice/complex-numbers) practice exercise.

> I wrote Python's complex concept (and R's), so I have some personal interest in finding an exercise that works with all of them.

Would you be interested in getting an `about.md` together? That might help me brainstorm what kinds of elements should be included in an exercise. That said, I feel that it could be easier to go from a Python exercise and whittle it down to better suit Julia than it would be creating a Julia exercise and expanding it for Python, so, unfortunately, I'm not sure how useful I could be to the cause of creating cross-track exercise.

What I feel is the biggest roadblock here is the lack of real world examples which are simple to explain and/or understand and/or implement in an exercise. For that reason, the target audience seems to be a bigger factor in deciding an exercise for this topic than others. However, as I said, I'm new to creating concept exercises, so I'm sure there are tricks of the trade of which I'm unaware, and there very well could be a good way to bridge this gap which is currently beyond me, and I'm looking forward to further input from Bethany or others with more experience in the matter.

depial:
**Afterthought**: I hope the `complex-numbers` practice exercise doesn't clash with the concept of the same name. I know you shouldn't name a concept exercise the same as the concept, but I don't remember seeing anything about practice exercises.

colinleach:
The `about.md` trio:
- [Python](https://exercism.org/tracks/python/concepts/complex-numbers) (the most polished one - Bethany did a _lot_ of polishing of my draft)
- [Julia](https://github.com/exercism/julia/blob/f8eb631d3f36b9d9f09da25280a3387667b2c5c7/concepts.wip/complex-numbers/about.md)
- [R](https://github.com/exercism/r/blob/80e4ba81f8c1e6f9299c78e263fff07df789fb40/concepts/complex-numbers/about.md)

You will notice a strong family resemblance!

> I hope the complex-numbers practice exercise doesn't clash with the concept of the same name

No, the website seems to have no trouble separating concepts from exercises.

depial:
Thanks! That looks familiar. I must have forgotten the Julia one was sitting in `concepts.wip`. I now see a couple of things I would mention to more specialize it to Julia.

I love that all three languages chose a different convention (`i`, `j`, `im`).

Anyway, from the `about.md`, I still get the same feeling for the target of the exercise, so, without further input, I doubt I would change too much as far as adding or removing language features. I'll mull this exercise over for a bit, and if nothing else, maybe we can turn it into a practice exercise and try to adapt whatever goes into the Python concept exercise, since that sounds like it'll be the heaviest.

colinleach:
I feel (my head feels!) that I'm about done for today. I started porting `locomotive-engineer` to go with `functions`, but it's early-stage.

After that:
- `tisbury-treasure-hunt` (tuples, from Python)
- `captains-log` (randomness, from my R draft)
- `name-badges` (nothingness, also my R draft))

You probably noticed my posts on the forum. Not much response yet, but that's normal for Saturday evening.

I've been vaguely thinking about what concepts to write next, not backed by any action. 
- `types` is really important, but likely to be difficult (that's why I didn't do it earlier)
- `bitwise-operations`, `errors` and `regular-expressions` should be easier, with some plagiarism from other tracks.

depial:
> I feel (my head feels!) that I'm about done for today.

I hear you. My brain checked out a while ago :D

> You probably noticed my posts on the forum. Not much response yet, but that's normal for Saturday evening.

Yep! Luckily, there's less urgency now that the syllabus is flowing. I would still like to hear other's ideas/experiences though.

> `types` is really important, but likely to be difficult (that's why I didn't do it earlier)

Indeed this topic seems complicated. However, there are a few old concepts in `concepts.wip` such as [`abstract-types`](https://github.com/exercism/julia/tree/main/concepts.wip/abstract-types) and [`composite-types`](https://github.com/exercism/julia/tree/main/concepts.wip/composite-types) which we might make for good source content.

One thing I see in there that could be a good topic would be [`multiple-dispatch`](https://github.com/exercism/julia/tree/main/concepts.wip/multiple-dispatch) since it's the primary paradigm of the language.

colinleach:
> One thing I see in there that could be a good topic would be [multiple-dispatch](https://github.com/exercism/julia/tree/main/concepts.wip/multiple-dispatch)

Totally agree, but it absolutely depends on `types` - so not a quick win.

depial:
> Totally agree, but it absolutely depends on `types`.

Ha! I didn't even put that together. Is there an emoji for "brain fart"? :brain: :dash: ?

In any event, my day of pretending to be intelligent is officially over!

BethanyG:
> @BethanyG Thanks for your views!

As noted in the `turntable` exercise -- please take everything I say with a grain of salt -- I can be super-picky, and that's not always a good thing.  **I think you and Colin are doing great work, and I don't want my hangups to be any sort of delay or blocker for you.**

> I agree the story is contrived, so I'll be looking into other possible ideas. A couple questions on your thoughts:


> * I had avoided an image rotation exercise because I believe there are cropping/filling issues at the corners for arbitrary rotations. Is there a way around this?

I have some other details in those comments on `turntable` -- but I will repeat them here.  Apologies for not getting back to you yesterday my time - I just ran out of time in the day.  I do think [we could finagle our way out of this](https://stackoverflow.com/questions/74546776/cropping-an-image-after-rotation-scaling-and-translation-with-python-transform) (apologies for the half-assed SO post) by having a very wide margin around the image to be rotated, so that fill and crop were less of an issue (and then the tests crop it off or only check the foreground as opposed to background).  We could (maybe) also make the image a circle equivalent, so that corners were less of a problem.  Or make the image exactly square, so that aspect issues were reduced.  We might also only do specific set rotations (90, 180, 270, 360) to make the edges work.  I am (most likely naively) optimistic. 

> * When you mentioned "Big Circle" navigation, does this refer to the stereographic projection of a great circles from a sphere onto circles on the complex plane? If so, that could be an interesting avenue to explore :)

Yes - my addled brain went to "Big" but I meant "Great Circle".  Specifically the [Riemann sphere](https://en.wikipedia.org/wiki/Riemann_sphere), and [mapping it to the complex plane](https://navlist.net/img/139485.great_circles_and_rhumb_lines_on_the_complex_plane.pdf).  Some other resources I've come across:  [a cookbook!](https://github.com/ProjectPythia/great-circle-cookbook), an ancient [python recipe](https://code.activestate.com/recipes/393241-calculating-the-distance-between-zip-codes/) for calculating distances between zip codes we could possibly adapt, and the [Aviation Formulary](https://edwilliams.org/avform147.htm) - which even has worked examples in JavaScript, of all things.
 
> @colinleach Judging from the rest of BethanyG's comments, I'm not sure a Julia exercise in the mold I'm currently considering would port well to Python since the user base is generally more mathematical and:
> 
> > "Julia includes predefined types for both complex and rational numbers, and supports all the standard [Mathematical Operations and Elementary Functions](https://docs.julialang.org/en/v1/manual/mathematical-operations/#Mathematical-Operations-and-Elementary-Functions) on them. [Conversion and Promotion](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#conversion-and-promotion) are defined so that operations on any combination of predefined numeric types, whether primitive or composite, behave as expected."
> 
> Which why I focused more on exposing users to the inbuilt methods to advance the concept. Of course, I'm new to this, so I'm not sure if my idea really constitutes a concept exercise or not. I was just running with the idea of vector rotation since it seems to be the simplest possible example. Let me know if you think I'm off base here and we should change tack.

Python has everything Julia has - and there are many mathematically inclined Pythonistas.  I think the difference is that there is a very very strong streak of _"how does this apply in the real world?"_ running through things, whereas Julia users are already convinced of utility/functionality.

depial:
>  I think the difference is that there is a very very strong streak of "how does this apply in the real world?" running through things

This helps me understand more of where you're coming from! I've been trying to figure out why you would like to use, what I feel are, such conceptually difficult examples for an exercise which is meant to include "simple" things like complex addition, since these seem to be contradictory aims (i.e. audiences) to me. However, if real applications are a dominant concern, I believe I understand better.

In cases like this, I like to assume I'm missing something from the big picture, so I'm looking forward to finding out more on how one would reconcile these competing forces. Are there any current Exercism exercises I could read which have this level of conceptual sophistication?

> We might also only do specific set rotations (90, 180, 270, 360) to make the edges work.

One other potential snag that came to mind: I believe those are the four rotations you could do without any complications. Outside of those, I think you'll have to deal with complications resulting from the [lattice points on a circle](https://en.wikipedia.org/wiki/Gauss_circle_problem) not being one-to-one with the lattice (which raises questions about how to perform the mapping consistently).

PR #871, opened 2025-04-05
Deprecating old-elyses-enchantments

colinleach:
Submitting this with some trepidation. I _think_ it repeats what worked for Annalyn...

### DIFF
--- a/config.json
+++ b/config.json
@@ -196,6 +196,14 @@
         "concepts": [],
         "prerequisites": [],
         "status": "deprecated"
+      },
+      {
+        "slug": "old-elyses-enchantments",
+        "name": "Old Elyse's Enchantments",
+        "uuid": "234466a0-08ce-45a1-aebc-5ec0b0cde24f",
+        "concepts": [],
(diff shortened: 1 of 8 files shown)


depial (approved):

colinleach:
Wow - have you seen the website! Apparently, that's what we needed all along.

depial:
Great! That's a relief!

I'm glad it worked out. I hope the worst is now behind us :)

depial:
Should these still be there?

image

I did Old Elyse's Enchantments and it said I had failed the tests on the updated exercise. Oddly enough, the stub is called `elyses-enchantments` and not `old-elysesenchantments`, even after I reverted to exercise start. Annalyn's Infiltration is the same, but it didn't say there was a new version of the exercise...

Things still seem to be a bit mixed up, but at least I feel like we're making two steps forward and one step back, rather than the other way around :D

colinleach:
My guess (I put it no more strongly!) is that the two of us see these because we already published solutions. I remain hopeful that they will be invisible to students.

PR #872, opened 2025-04-06
Add practices

colinleach:
This is a small experiment, to see if this change does what I think it should. I'm hoping to see if these 2 easy exercises are added to the concepts on the learning syllabus, unlocking when the relevant concept exercises are completed.

At some point in the future, we'll have a _lot_ more of these to add.

I'm borrowing from the Python track, as so often. 71 of the 82 Julia practice exercises have a Python equivalent with the same slug and a non-empty "practices" field. I'll also look at some others later (C#, F#, Crystal...)


depial (approved):

colinleach:
Thanks. That did what I hoped.

PR #873, opened 2025-04-07
add concept exercise tracking-turntable

depial:
Here's another go at a potential concept exercise for `complex-numbers` which is quite simple, and which I hope doesn't sound quite as contrived. I ended up sticking with the idea of rotations in 2D since I couldn't think of how to simplify other "real-world" applications. A couple of points:

- This is fairly simple mathematically speaking, with no messy, convention-specific stuff.
- The first function teaches how to do rotations.
- The second function teaches how to do radial displacements.
- The last function is simply a composition of the first two.
- I'm not sure if `instructions.md` need more or less detail (in either backstory or math) and I intentionally left a few things somewhat vague (e.g. I don't talk about working with vectors to avoid confusion with the data structure).

Overall, I feel better about this one due to its simplicity and relatively low levels of abstractness. I'm not sure if there are other tests with different functionality that I could include to make it more expansive, but we could brainstorm if so desired.


depial:
I've tentatively included an `introduction.md` based on this [`about.md`](https://github.com/exercism/julia/blob/f8eb631d3f36b9d9f09da25280a3387667b2c5c7/concepts.wip/complex-numbers/about.md?plain=1). There are a few standard deletions for conciseness, but I've also added a couple of items I feel are relevant to the exercise.

depial:
> First impressions are that I like this a lot better

I feel much better about it. I think the first one was me trying to fit a concept into an exercise, and this time it was the other way round.

> [I'll] also update the complex-numbers concept to match (cispi, etc).

Give that a sec before updating. I might still throw a phrase or two in while I'm trying to better sync `introduction.md` with `instructions.md` (I feel the former is basically good to go, but the latter still needs polishing)

colinleach:
> since this is a pretty light exercise, is that we could add extra tasks

Yes, I'd be in favor of things that give the gentlest possible introduction to the exercise. In the Julia world, alot of people may already be highly familiar with complex numbers. Exercism aims for a broader audience, where possible.

colinleach:
> You're good to go!

I'll add this to the TODO list. Currently I'm working on `locomotive-engineer` but having to change the tasks in several places: partly to make it more about functions, less about dictionaries, partly because because it highlights a clash of idioms between Julia and Python.

Also, this exercise depends heavily on tuples, so I need to move `tisbury-treasure-hunt` up the priorities to satisfy the prereq.

BethanyG:
>  First impressions are that I like this a lot better. I'll need to go through it in more detail later, and also update the `complex-numbers` concept to match (`cispi`, etc).
> 
> @BethanyG is this getting closer to what you would want for Python?

I'm sorry -- No - I don't think it is getting closer.  It feels like we are rushing to get something - anything done for this, and I sorta feel like we could do other concepts and exercises both better and easier, while we think about this one more. If this is a priority for Julia, then maybe we let Python lag.  This currently feels like I could use trig functions - and the fact that complex numbers are "fun" just wouldn't push me.  I know that's a personal bias, but I'm craving a problem that would be _simplified_ or _clarified_ by complex_numbers.  

I did some research yesterday, and I think I am (personally) moving toward one of these:

1.  A complex-number based formula for Great Circle navigation/segmentation (_find the distance and waypoints between two points on the globe_).  This would exercise the polar coordinate forms, as well as demonstrating how complex-numbers simplify the system of equations needed for a solution.  I've got some code samples, and some really great resources for this...just not yet a clear outline.

2. A (very) lite Fourier Transform.  I really like this article from [Tom Rocks Maths](https://tomrocksmaths.com/2021/10/19/air-traffic-control-and-radar/), about Air Traffic Control and RADAR.  I think it really shows how complex numbers can simplify things.  It might be too much -- but it also feels very "real world" and very motivating to me.

3. A series of image rotations/translations.   So a variation on something like/similar to the WIP [exercism-matrix](https://github.com/exercism/julia/blob/main/exercises/concept.wip/exercism-matrix/.docs/instructions.md) Padding would take care of the crop and fill issues there ... but we could also make it so that rotations were a multiple of 90, or only check the portion of the image that contained the data we wanted, or make the image margin an approximate circle, so that rotation and translation didn't have corners.  This would likely exercise/motivate addition/subtraction etc.  Again, not yet a clear outline, but there are really good resources out there.

4. A very small [Gimbal Lock](https://en.wikipedia.org/wiki/Gimbal_lock) problem, using [quaternions](https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation).  I haven't thought this one through entirely, but it feels like it might really motivate at least the gaming-minded.  It gives a clear reason why you can't use traditional formulas and numbers when rotating in 3D.

5. [Robot Simulator](https://github.com/exercism/problem-specifications/blob/main/exercises/robot-simulator/description.md) goes global (_navigate on a sphere_) or airborne (_navigate in 3D_)).  This one is quite vague in my mind right now, but it might be fun - it would be a variant on either the Gimbal Lock above, or Great Circle navigation.

BethanyG:
I am sorry to disappoint you.  I am just really fussy about this one, which is why I don't want Python considerations to be in any way a blocker on the good work you two are doing for Julia.  I do not want to get in anyones way, and I certainly don't want to stifle creativity -- so please don't let me be a roadblock.  If you two feel that this works for you and your goals for the track -- then **it works**.

I just feel like (for Python) if this concept lays an egg, no one will bother even attempting the exercise.  And since everything takes so much effort, that would be a big bummer.  I know I am super-duper slow ... but I need more time to think on this.

Part of the issue here is that I am still working on things like Walrus and other concepts in the queue for Python.  I am also trying to upgrade Python versions and tooling, as well as investigate moving to using WASM as opposed to Docker.  So I am ... not exactly in the right mindset.  I apologize.

colinleach:
> I am still working on things like Walrus 

Nice response! I know who has been pressuring you to merge this...

BethanyG:
I also keep falling down rabbit holes like [this](https://www.reedbeta.com/blog/why-quaternions-double-cover/) - which is excellent, but doesn't get the WASM compiling done, nor the Walrus examples updated.  :-|

depial:
After thinking about everything, since I'm more inclined toward theory, I believe I would be happier with this exercise as a starting point, since the math and concepts are relatively simple. Afterwards we could use one of the topics mentioned by BethanyG as a practice exercise.

I understand better now that for practically minded people, the other way around is likely more appropriate.

@colinleach BethanyG mentioned that practicality is an important undercurrent in the Pythonista community (if I understood correctly!), but perhaps not so much in the Julian one. If you think we can focus on theory first (i.e. Here's the machinery. This is *how* you use it.) and temporarily forego practicality aspect (i.e. This is *where* you use it.). I think we could split these up rather than trying to combine them.

Any thoughts?

depial:
> I also keep falling down rabbit holes like [this](https://www.reedbeta.com/blog/why-quaternions-double-cover/) and [this](https://ntrs.nasa.gov/citations/19770019231)

@BethanyG In case you would like another [rabbit hole](https://www.youtube.com/watch?v=0w5QnhS_55U&list=PLJHszsWbB6hoOo_wMb0b6T44KM_ABZtBs&index=8) ;)

This series has a lot of gems in it if topics like quaternions are interesting to you. Like this one on [double covers](https://www.youtube.com/watch?v=Ay482u6b3Xo&list=PLJHszsWbB6hoOo_wMb0b6T44KM_ABZtBs&index=11)

colinleach:
@depial, I'll support your decision. It would have been nice to have something more broadly applicable, but trying to smoosh together 2 tracks with different user bases and different priorities isn't working out well.

config.json
colinleach, line 1: My habitual mistakes are proving infectious. The name needs updating from the slug.

exercises/concept/tracking-turntable/.docs/introduction.md
colinleach, line 1: The [docs](https://exercism.org/docs/building/tracks/concept-exercises#h-file-docs-introduction-md) discourage links in Concept Exercise introductions. I removed them from `complex-numbers` (give or take a stray ']'), so it might be easiest to copy that across.

colinleach:
To Be Continued, tomorrow. My head has given up for today.

depial:
I hope your headache feels better soon!

I'll make the initial changes you've suggested here, and you can feel free to continue when you feel up to it.

exercises/concept/tracking-turntable/.docs/instructions.md
colinleach, line 1: The [convention](https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions) for paragraphs recommending how we want students to approach an exercise  is to wrap it in an "admonition".

```markdown
~~~~exercism/note
These operations can be done through trigonometric functions and/or rotation matrices.

However, they can be made simpler (and more fun, I assure you!) with the use of complex numbers via rotations and radial displacements. 

In fact, each function in this exercise can be written in assignment form (i.e. a one-liner) using complex numbers and built-in Julia methods/functionality, should you so desire.
~~~~

This seems to mess with line wrapping, so I added some breaks in the above suggestion. The online editor may do it better, so this is an experiment.

Comment (colinleach) [reply exercises/concept/tracking-turntable/.docs/instructions.md:1]: > This seems to mess with line wrapping, so I added some breaks in the above suggestion. The online editor may do it better, so this is an experiment.

Also, I seem to have not closed the note properly! I have much to learn about this...

**Comment by colinleach** [comment]
The last 3 paragraphs of the instruction header seem like too much technical detail for this document, which works best if it short and punchy (even then, getting students to read it is an uphill struggle!).

> This ease results from Euler's elegant formula, `ℯ^(iθ) = cos(θ) + isin(θ) = x + iy`, where `i = √-1` is the imaginary unit and `|x + iy| = 1`.
With `r = |x + iy|`, we have the more general polar form of `r * ℯ^(iθ) = r * (cos(θ) + isin(θ)) = x + iy`.

> For rotations, the complex number `z = x + iy`, can be rotated an angle `θ` about the origin with a simple multiplication: `z * ℯ^(iθ)`.
Note that the `x` and `y` here are just the usual coordinates on the real 2D Cartesian plane, and a positive angle results in a *counterclockwise* rotation, while a negative angle results in a *clockwise* one.

> Likewise simply, a radial displacement `Δr` can be made by adding it to the magnitude `r` of a complex number in the polar form (eg. `z = r * ℯ^(iθ)` -> `z' = (r + Δr) * ℯ^(iθ)`).
Note how the angular part stays the same and only the magnitude, `r`, is varied, as expected.

I'm not sure what to recommend instead, but possibilities include:
- Move these to the top of `hints.md`
- Use them to expand the `introduction.md` discussion of Euler's equation (then copy it to the concept docs).

File: exercises/concept/tracking-turntable/.docs/instructions.md
Comment (colinleach) [reply exercises/concept/tracking-turntable/.docs/instructions.md:1]: Looking around at the way other tracks use admonitions, the Exercism website seems to handle them well. It's GH that gets confused.

Comment (depial) [reply exercises/concept/tracking-turntable/.docs/instructions.md:1]: Thanks! It's good to know where to use that. I've added it as suggested.

**Comment by depial** [comment]
I prefer the idea of adding it to `introduction.md` since I was already feeling like something of the sort should be in there with the change I suggested before:

> (short for `cos(x) + isin(x)`)

So I've put it in there where I thought it would fit best. Let me know if you don't like it and we could move it to the hints, which I also thought was a good idea.

If you think it's fine in `instructions.md` we'll have to modify the files from the concept since I left those be until I got your feedback.

File: exercises/concept/tracking-turntable/.docs/introduction.md
Comment (colinleach) [inline exercises/concept/tracking-turntable/.docs/introduction.md:1]: Thinking about it since my previous comment, I realized this was the way to go. I'm glad we both came to the same decision independently.

One detail: does the start of line 144 need a bit of rewording? It's no longer in the same context as before.

```markdown
This ease results from Euler's elegant formula ...
```

colinleach (approved):

colinleach:
Once you decide on the final wording of the Introduction, I'll copy the changes to the concept docs.

exercises/concept/tracking-turntable/.docs/introduction.md
depial: I've rephrased and added one extra line. Let me know if you don't think it's appropriate.

depial:
Seems to be working just fine on the website.

Oddly enough, after I finished the exercise, I was awarded the "Fundamental" trophy for completing learning mode. Strange because I think this is still `"status": "wip"`

colinleach:
Maintainers are not as other students!

Should it be changed to "beta", or do you want to wait a bit?

depial:
We can switch it to beta now. I just wanted to make sure it was looking alright on the website

keiravillekode:
Currently there only stubs for 3 of the 5 methods.

`z(x, y)` and `euler(r, θ)` are mentioned in the instructions and tests. They should also be  included in `tracking-turntable.jl`

depial:
Thanks! Addressed in #901

Issue #874, opened 2025-04-08
Moving the syllabus to public visibility?


colinleach:
My feeling is that we are within a week or two of being able to open the learning track to wider visibility, at least in preliminary form (but no promises!). Then we can get feedback on what works and what needs fixing.

I don't want to get into a full critical path analysis (Microsoft Project! GANTT charts! I did more than enough of this when I worked in industry!!). More roughly, these are the key steps I can think of:

- Implement the concepts that others have as a prereq. As far as I can see, this means `functions` (`locomotive-engineer`, in progress) and `tuples` (`tisbury-treasure-hunt` or something equivalent). Others are nice to have, but I think they can be added later.
- Check all the prereqs are correct, once all the relevant concepts are in place. For example, `high-school-sweetheart` will need to change, perhaps to ["functions", "strings"].
- Add "practices" entries for some more Practice Exercises. I did a couple of these a few days ago, just to test the process. For inspiration, I made a [sheet](https://docs.google.com/spreadsheets/d/1OprA7-ch1_Jub2Zze62otK1-xemjx2D3FUTEFkVsJFM/edit?usp=sharing) with tabs for various other tracks, showing what they chose. Of course, idioms differ, so these may not translate directly to Julia.

For visibility, we need `"concept_exercises": true` near the top of the `config.json`, as Glenn pointed out on the forum. I think we also need to change to `"status": "beta"` for each exercise, but that needs to be confirmed.

Anything else?

colinleach:
> the deprecated exercises are still on the site

Are you referring to them showing up on the Practice tab, rather than the syllabus? I think that's only for maintainers, so students can't see them. We can check. I'm not seeing them on the Learn tab.

depial:
> Are you referring to them showing up on the Practice tab

Yep, my fear would be that they are still entangled somehow, so when we make the syllabus public, those go public on the Practice tab too.

I'm unsure about the level of contortion that we've got going on now, so this is one thing I could imagine happening, although I have no reason to believe it would, and I have no idea on its likelihood. Just something that came to mind.

BethanyG:
👋🏽 

Just to chime in here.  Deprecated exercises will always show for anyone who:

1.  Is a maintainer
2.  Has started the exercise sometime in the past
3.  Has completed the exercise sometime in the past

[...]

PR #876, opened 2025-04-10
Move concept.wip to reference

colinleach:
See #874 for discussion.

Should we do this?


depial:
I've provisionally approved this. I think I'll be near the computer all day today, so just let me know when you would like to try merging this.

colinleach:
Unusually, I'm awake early this morning (not sure how long it will last!). Should we press the button?

colinleach:
I've had a couple of messages from you in the last 5 minutes, so I took the risk and merged. Wait and hope...

colinleach:
So far, the change on the website is ... absolutely nothing. I'll give it an hour.

If this fails, I guess the next step would be to move them to exercises/concept, create config.json entries and deprecate.

depial:
Sounds good!

PR #811, opened 2024-10-10
WIP draft of tuples concept

colinleach:
A popular exercise to go with this is `tisbury-treasure-hunt` (Python and several other languages), but we have multiple choices.

PR #879, opened 2025-04-12
Locomotive Engineer concept exercise

colinleach:
The `exemplar.jl` runs the tests (after some hassle!), but I think it could be improved. Have fun seeing what you can suggest.

The `hints.md` is probably grossly inadequate in its current form. Comments/suggestions welcome!

This is one I found difficult to write, and I suspect students will find it confusing and challenging to solve. Any improvements in the documentation could be very useful. Though we can only do so much - this really is quite complicated, and will be unfamiliar to many people.

I think this is the penultimate big step towards getting a syllabus we can release. We still need tuples (#811 draft) and an exercise to go with it. Because locomotive-engineer uses tuples and named tuples so much, I think I now have a clearer idea how to update the tuple concept.

I promise I'll get back to your complex number PRs! It's just that syllabus release now feels _so_ close.


colinleach:
FYI: The Python version of this has a 5th task, but I didn't port that to Julia. It's basically a matrix transpose, done via a tricky bit of list unpacking.

```python
def fix_wagon_depot(wagons_rows):
    [*row_one], [*row_two], [*row_three] = zip(*wagons_rows)
    return [row_one, row_two, row_three]

>>> fix_wagon_depot([
                    [(2, "red"), (4, "red"), (8, "red")],
                    [(5, "blue"), (9, "blue"), (13,"blue")],
                    [(3, "orange"), (7, "orange"), (11, "orange")],
                    ])

[
[(2, "red"), (5, "blue"), (3, "orange")],
[(4, "red"), (9, "blue"), (7, "orange")],
[(8, "red"), (13,"blue"), (11, "orange")]
]
```

I'm not sure how you would even do that in Julia. Probably treat it as a matrix and transpose it? It's certainly off-topic for a functions concept.

depial:
I'll try to look through this carefully tomorrow or Monday. Brain's not with me at the moment ;)

colinleach:
> Brain's not with me at the moment

You and me both!

Luckily, `tisbury-treasure-hunt` is proving a lot simpler to port, so I hope to get that PR'd tomorrow. `tuples` are an important prereq for `functions`, so we need it.

exercises/concept/locomotive-engineer/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
+# Hints
+
+## 1. Create a vector of all wagons
+
+- The [`collect()`][collect] function is useful here.
```
depial, line 5: We could mention the splatting option as well: `[args...]`

depial (approved):
Soft suggestions rather than hard ones. I don't see any issue with how things are already, so feel free to keep it as is.

depial:
> The `hints.md` is probably grossly inadequate in its current form. Comments/suggestions welcome!

I didn't see a problem with `hints.md`, since it seems to mention the main concepts needed in `exemplar.jl`. If you feel you want to expand, I could see that being useful, but I'm unsure what further info you would like to include other than perhaps more detail in what is already there?

> This is one I found difficult to write, and I suspect students will find it confusing and challenging to solve.

I agree that it's a bit to wrap my mind around, and especially if I were learning splatting/slurping for the first time, so again, I don't see anything wrong with maybe a bit more description in `hints.md` if you see a good place for it.

> FYI: The Python version of this has a 5th task, but I didn't port that to Julia.

I'm okay with leaving this task out. It seems overly complicated and potentially frustrating for what can be gained from doing it.

colinleach:
Thanks for your comments. I'll think about this some more tomorrow, when I hope to be more awake.

depial:
One other minor note. I'm used to the vocabulary of: *engine* and *(train) cars* instead of *locomotive* and *wagons* (where I've hear *locomotive*, but *wagons* is new for me). I'm not sure if this is American vs British English thing or a regional American thing, though. I assume the instructions were lifted from another version, so I doubt there's any need for any adjustment. It's just something that popped out at me.

colinleach:
OK, I hadn't spotted that - perhaps because I'm a US/UK dual citizen who split my life between the two countries.

It's a bit surprising, because I lifted this from the Python track, which is firmly under Bethany's control. She is Californian to her core, I think it's fair to say.

exercises/concept/locomotive-engineer/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
...
+
+## 2. Fix the vector of wagons
+
+- It is useful to split the front wagons and the engine from everything else, so that they can be reassembled in the correct order.
+- Multiple assignment with splatting may help you.
+
+## 3. Add missing stops
+
+- Only the stop name is needed, not the stop number, so each pair must be split.
+- There are multiple ways to assemple the vector of names: list comprehension, broadcasting of an anonymous function, map...
```
depial, line 15: Maybe one way to add a bit more here is to expand this into separate points or a sub-list?

- [Array comprehensions](https://docs.julialang.org/en/v1/manual/arrays/#man-comprehensions)
- [Map](https://docs.julialang.org/en/v1/base/collections/#Base.map)
- [Broadcasting](https://docs.julialang.org/en/v1/manual/arrays/#Broadcasting) of an [anonymous function](https://docs.julialang.org/en/v1/manual/functions/#man-anonymous-functions) (e.g. `(x -> x^4).([1,2,3]) == [1, 16, 81]` evaluates true)

I think an example of broadcasting an anonymous function is likely a good idea since I didn't readily find how to do this in the documentation (aside: I remember being quite surprised the first time I found this was possible).

Of course, you would likely want to add the links at the bottom of the file as you did with `collect` and `merge`

exercises/concept/locomotive-engineer/.docs/introduction.md
```diff
@@ -0,0 +1,224 @@
...
+ 10
+ 18
+```
+
+In this case, `x -> 2x` is an "anonymous function".
+This is equivalent to what some other languages call a "lambda function".
+
+Note that multiple arguments need parentheses, as in `(x, y) -> x * y`.
+
+Anonymous functions are common in Julia code, especially when combined with higher-order functions such as `map()` and `filter()`.
```
depial, line 39: I think a quick example here would be good as it may help encourage students to use it. Something like:

```julia-repl
julia> map(x -> x^4, [1, 2, 3])
3-element Vector{Int64}:
  1
 16
 81
```

This could line up with any example in `hints.md` for broadcasted anonymous functions.

exercises/concept/locomotive-engineer/.meta/design.md
```diff
@@ -0,0 +1,31 @@
...
+- Higher Order Funtions are only briefly mentioned, and will be the subject of a later concept.
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `function-composition`
+- `higher-order-functions`
+- `multiple-dispatch`
+- `complex-numbers` (probably?)
```
depial, line 26: I agree with this unlocking `complex-numbers` since then I could add a note that the exercise can be solved with one-liners when using complex numbers (since I used a splat in `exemplar.jl`). I had been hoping to point this out as another way to show the convenience of complex numbers.

depial:
I realized that I didn't give very detailed options/suggestions, so I've added a couple more things that I had initially noticed but didn't make clear. Again, feel free to incorporate or not as you feel is best for the exercise.

colinleach:
This is an very helpful set of improvements, thank you!  Addressing them is a bit more than a quick one-line edit, so (despite my natural impatience) it might be better to delay addressing them until I've had some coffee. I'll probably introduce more errors if I do it now.

exercises/concept/locomotive-engineer/.meta/exemplar.jl
```diff
@@ -0,0 +1,14 @@
+get_vector_of_wagons(args...) = collect(args)
```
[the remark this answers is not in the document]
colinleach: changed

exercises/concept/locomotive-engineer/.docs/hints.md
colinleach: added

exercises/concept/locomotive-engineer/.docs/hints.md
colinleach: all changes added

exercises/concept/locomotive-engineer/.docs/instructions.md
```diff
@@ -0,0 +1,107 @@
...
+Your friend Linus is a Locomotive Engineer who drives cargo trains between cities.
+Although they are amazing at handling trains, they are not amazing at handling logistics or computers.
+They would like to enlist your programming help organizing train details and correcting mistakes in route data.
+
+~~~~exercism/note
+This exercise could easily be solved using slicing, indexing, and various `Dict` methods.
+However, we would like you to practice packing, unpacking, and multiple assignment in solving each of the tasks below.
+~~~~
+
+## 1. Create a vector of all wagons
```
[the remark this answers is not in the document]
colinleach: I'm a bit undecided about this, but leaning against. It seems to be an Exercism cultural thing to name things for the application, not the implementation.

exercises/concept/locomotive-engineer/.docs/introduction.md
colinleach: added, along with a note that we'll do more on higher-order functions in another concept

depial:
Everything looks good on the website!

> Stuff on naming (both trains and functions) is unresolved for now.

I think we can consider these resolved. I don't think we need to deal with the train terminology since the current vocab is technically correct. Also, it's perfectly fine with me if you want to continue with the specific function name, especially since there is no real connection between the tasks, so throwing a connection in doesn't really help anything.

PR #880, opened 2025-04-13
Tuples concept

colinleach:
Replaces the old draft #811.

The Introduction is synced with `tisbury-treasure-hunt`, which will be PR'd soon.


colinleach:
I just looked back at the old draft, where you said this:

> I'm not sure if it's too advanced of a topic, but we could mention that tuples can still be broadcast over (requiring the . syntax to do so).

```julia
1 .+ (3, 2) = (4, 3)
1 +  (3, 2) = MethodError
```
Since then, we've put all the stuff about broadcasting into `vector-operations`, which isn't currently a prereq of `tuples`. I'm a bit reluctant to change that, but feel free to persuade me.

depial (approved):
lg2m

depial:
> I'm a bit reluctant to change that, but feel free to persuade me.

I'm fine with leaving this out. It's fairly intuitive and unrelated to the task at hand.

depial:
Before merging, please see my review of `Tisbury Treasure Hunt`

concepts/tuples/introduction.md
```diff
@@ -0,0 +1,105 @@
...
+ERROR: MethodError: no method matching setindex!(::Tuple{Int64, Float64, String})
+```
+
+There is also a `Tuple()` constructor, which will turn any iterable collection into a tuple.
+Usefully, this treats a string as an iterable collection of characters.
+
+```julia-repl
+julia> Tuple("Julia")
+('J', 'u', 'l', 'i', 'a')
+```
```
depial, line 37: This example is not present in `about.md`. Should we add it?

PR #881, opened 2025-04-13
Tisbury treasure hunt

colinleach:
Pairs with `tuples`.

This one was relatively quick to create, and I _hope_ you will also find it easier than `locomotive-engineer` to review.

I think/hope this is the final big piece needed to make the syllabus join up, with sensible prereqs. I still need to add a short section on vector comprehensions to `vector-operations` (as previously discussed).


colinleach:
In case anyone wonders: I borrowed from both the Python and F# versions. They are slightly different, and I tried to get the most appropriate features of each.

config.json
```diff
@@ -129,6 +129,18 @@
         ],
         "status": "wip"
       },
+      {
+        "slug": "tisbury-treasure-hunt",
+        "name": "tisbury-treasure-hunt",
```
depial, line 134: I think the name should be:

```
"name": "Tisbury Treasure hunt",
```

exercises/concept/tisbury-treasure-hunt/.docs/introduction.md
```diff
@@ -0,0 +1,105 @@
...
+ERROR: MethodError: no method matching setindex!(::Tuple{Int64, Float64, String})
+```
+
+There is also a `Tuple()` constructor, which will turn any iterable collection into a tuple.
+Usefully, this treats a string as an iterable collection of characters.
+
+```julia-repl
+julia> Tuple("Julia")
+('J', 'u', 'l', 'i', 'a')
+```
```
depial, line 37: I notice that this example is absent from `about.md` in the concept.

exercises/concept/tisbury-treasure-hunt/.docs/instructions.md
```diff
@@ -0,0 +1,96 @@
...
+Implement the `get_cooordinate()` function that takes a `(treasure, coordinate)` pair from Azara's list and returns only the extracted map coordinate.
+
+```julia-repl
+julia> get_cooordinate( ("Scrimshaw Whale's Tooth", "2A") )
+"2A"
+```
+
+## 2. Format coordinates
+
+Implement the `convert_coordinate()` function that takes a coordinate in the format "2A" and returns a tuple in the format `(2, 'A')`.
```
depial, line 62: When doing the exercise online, I noticed the tuple format should be `('2', 'A')` (two characters), to line up with the tests, instead of `(2, 'A')` (an integer and a character). This issue perpetuates through the instructions.

colinleach:
Careless copying from the F# version. I caught and fixed most of these, but this one slipped through.

PR #882, opened 2025-04-14
Update instructions.md

depial:
Convert integers to characters in tuples


colinleach (approved):

colinleach:
Thanks for fixing this.

PR #883, opened 2025-04-14
Update instructions.md

depial:
Correct the tuples in the table to be `Tuple{Char, Char}`


depial:
Oops! I overlooked the tuples in the table...

PR #886, opened 2025-04-14
Activate syllabus

colinleach:
`"concept_exercises": true` but exercises still "wip"


colinleach:
I don't know if you monitor Discord, but @BethanyG recommended making this change first. It should become visible to  cross-track maintainers, but not the wider public yet.

depial (approved):

depial:
> I don't know if you monitor Discord, but @BethanyG recommended making this change first.

Seems like a good first step then :)

BethanyG:
Whelp, apologies are in order @colinleach and @depial.

As a maintainer, I can see the `learn` tab — but sadly, no exercises.  So **`wip`** status only works for the track maintainers.  **D'oh!**.  Sincere apologies.  Looks like its **`beta`** FTW.

So we'll have to deal with students as well as maintainers — but with the heads up, maintainers can get in there and give feedback quickly, so there is that as a silver lining.....

colinleach:
OK, we know what to do next. PR in prep...

PR #887, opened 2025-04-14
Exercises to beta

colinleach:
Now we're _really_ pressing the Big Red Button. Hold your breath...


depial (approved):
Wheeee!

depial:
I can see the syllabus on an incognito look at the website. So that's looking good!

depial:
Somewhat awkwardly, the number of exercises doesn't often seem to align with what's shown. For example:

image

colinleach:
Odd, I'm not seeing that. I wonder if it's related to you still seeing the deprecated exercises, because you solved them?

depial:
> I wonder if it's related to you still seeing the deprecated exercises.

I don't think it should, because that's what I'm seeing from the incognito view without signing into the site. Maybe there's something else that ups that number when not signed in, because from my account, everything looks normal:

image

depial:
Another thing that I'm hoping is just related to my account's exercises being mixed up:

image

I believe it should have awarded me this considering I've mastered all of the concepts.

There are a couple of other minor things, but, again, I'm hoping they will only show on my account.

colinleach:
My incognito view is signed in (to a novice account), and the number of exercises looks OK. Odd, but I'll be very happy if that's the worst problem we find!

I still need to publish a couple more exercises before I can check the trophy. The website certainly hasn't caught up with the 15 beta concepts yet: the [track build status](https://exercism.org/tracks/julia/build) still shows zero.

colinleach:
Update: I got the trophy, and the build status now shows 15 concept exercise. Just a slow sync.

depial:
That's good to hear! At least it looks to be working properly for regular students. My trophy still doesn't show (among some other minor strangish behavior), but I'm assuming that's because of some residual crossed wires from the previous syllabus attempt. As long as that doesn't affect anyone else, we're good to go! 

I guess the only real downside is it'll be more difficult for me to separate general site issues from issues specific to my account. I wonder if I should reset my track? Seems overkill for now, but I'll consider it.

PR #893, opened 2025-04-16
Minor corrections to descriptions

blakelewis:
A few missing words or letters, and a spurious reference to writing Python scripts.


colinleach (approved):
Thank you! We suspected that getting more eyeballs on the syllabus would uncover problems that we had repeatedly read past.

PR #895, opened 2025-04-18
Rational Numbers concept

colinleach:
Replaces #800, after a merge mishap.

It may be some time before we have an exercise to go with this. Reasons to merge regardless:
- It gives other concepts something to link to.
- It helps clear out the open PR list.

Agreed??

Currently, the Introduction just copies the About, so we may want to revisit that in future.


depial (approved):

depial:
> Agreed??

Agreed :+1:

PR #896, opened 2025-04-18
Nothingness concept

colinleach:
Replaces the (now-demised) #810.

Corresponding exercise `name-badges` will follow, probably tomorrow.


concepts/nothingness/about.md
```diff
@@ -0,0 +1,149 @@
...
+If a value really does not exist, it is represented by [`nothing`][nothing]. This is probably closest to what C (`NULL`) or Python (`None`) might do.
+
+```julia-repl
+julia> n = nothing
+
+julia> isnothing(n)
+true
+
+julia> typeof(n)
+Nothing
```
depial, line 19: Possibly could include an example using `notnothing` if you think it would be useful.

concepts/nothingness/about.md
```diff
@@ -0,0 +1,149 @@
...
+```julia-repl
+julia> filter(!isnan, [1, 2, NaN])
+2-element Vector{Float64}:
+ 1.0
+ 2.0
+```
+
+## Undefined data
+
+This is not something you are ever likely to want, but there are ways to create composite types (such as [`struct`][struct]) with uninitialized fields.
```
depial, line 132: This section feels a bit sparse.

Also, I think I've used this for something like initializing an empty array (for performance). I'll try to find a definite example if you want, because it was useful (or even needed) at the time.

depial (approved):

concepts/nothingness/about.md
depial: [Here's an example](https://exercism.org/tracks/julia/exercises/circular-buffer/solutions/depial) if you want to have a look.

concepts/nothingness/about.md
```diff
@@ -0,0 +1,149 @@
...
+Nothing
+```
+
+So `nothing` is a singleton value of type `Nothing`, and we can test for it.
+
+One common use of `nothing` is as a return (non-)value for functions which are used only for their side effects (printing, network configuration, or whatever).
+
+## `missing`
+
+For situations where a value exists in theory but we don't know what it is, [`missing`][missing] is used. For example, when counting vehicles traveling on a road, human observers might go off sick or automatic sensors break down, but the traffic continues to flow.
```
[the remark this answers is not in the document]
colinleach: Nearly 22 years since I moved to America, the Brit in me still breaks through. Americans who used to work for me joked that my emails made more sense if they tried reading them out in a British accent.

concepts/nothingness/about.md
colinleach: I found the documentation, but when I just tried this I got "UndefVarError: `notnothing` not defined in `Main`".

Very odd, the manual says it's in `Base`.

depial: Maybe it was deprecated?

Let's just leave things as is then.

concepts/nothingness/about.md
colinleach: I made some changes based on your suggestion. Better, or does it still need more?

colinleach: If felt a bit thin because I didn't know what I was talking about, and just quoted something I'd read...

depial: I think that looks good. I doubt an example would add much since it appears to be fairly advanced.

colinleach:
I'll merge for now so that #899 passes linting. Let me know if you'd like more changes (or just push them, if you prefer).

PR #900, opened 2025-04-19
Nothingness fix

colinleach:
I made the change in two places, and _of course_ I missed the third.


colinleach:
Other than this, it looks good on the website.

I started on `captains-log` for `randomness`, but that will take a day or two. It's also overdue to review your FFT practice exercise!

depial (approved):

depial:
I was able to do the exercise after resetting it (since I had done the previous version), but all went smoothly after that.

Just a heads-up, that next week will likely be the last full week of normal availability for me before I start traveling. I'll do my best to participate when I can, but my response time will likely drop, especially for the first three weeks of May.

colinleach:
OK. Once we get `captains-log` and FFT out, we'll have pretty much cleared the PR queue. Then I need to go back to writing concept docs, which means quite a lot of textbook- and manual-reading. The timing works out pretty well.

PR #906, opened 2025-04-22
Update hints.md

depial:
Updating to more closely follow the format from the Python [`darts`](https://github.com/exercism/python/blob/main/exercises/practice/darts/.docs/hints.md?plain=1) practice exercise.

The only thing that would differ here is there is a third level of list nesting, rather than just two.

### DIFF
--- a/exercises/practice/fft/.docs/hints.md
+++ b/exercises/practice/fft/.docs/hints.md
@@ -1,3 +1,5 @@
+# Hints
+
 ## General
 
 - DFT


colinleach (approved):

colinleach:
Worth trying. I certainly have no better ideas.

colinleach:
Looking good - there is now a "Get help" button

depial:
Yep, seems to be working now! :)

PR #907, opened 2025-04-23
Captains Log concept exercise

colinleach:
Pairs with `randomness`.

In the spirit of de-cluttering, I recycled the UUID from `annalyns-infiltration` and removed those files and config entry.

Note: I took much of this from the Elixir version, but modified Task 4 and added Task 5. I'm slightly handicapped by the fact that I never watched Star Trek (extremely unusual for nerds of my age). If I messed up the story, I may get death threats from enraged fans.


config.json
```diff
@@ -242,12 +242,16 @@
         "status": "beta"
       },
       {
-        "slug": "old-annalyns-infiltration",
-        "name": "Old Annalyn's Infiltration",
+        "slug": "captains-log",
+        "name": "captains-log",
```
depial, line 246: `"Captains Log"` or similar.

exercises/concept/captains-log/.docs/introduction.md
```diff
@@ -0,0 +1,264 @@
...
+
+`Random` is part of the standard library and likely to be pre-installed, but you will need to add `using Random` at the top of your program to bring its contents into the namespace.
+
+## The `rand()` function
+
+What this function does depends on the arguments you give it.
+There are _many_ options.
+
+With no arguments, it generates a float between 0 and 1.
+This is a `uniform` with all values equally likely, as discussed in the Working with Distributions section, below.
```
depial, line 30: Reading again, would this be better as: "This is a `uniform` distribution..."

depial (approved):
Interesting test suite! I would naturally want to set a seed and test deterministically, but I can see how that could fail if people choose different implementations.

depial:
> If I messed up the story, I may get death threats from enraged fans.

Unfortunately, I can't be of much assistance here since I was never a big fan. However, the stuff in the story aligns with what I remember, so that's a plus. Maybe?

depial:
On the website, I'm still getting the `introduction` from `Booleans` and the `instructions` from `Annalyn's Infiltration`

colinleach:
> Interesting test suite!

Yes, it needed some thought!  I'm not just going off-piste here, this is a recognized way to write Exercism tests. Erik uses it several times in the C# track.

The chances of getting a false-negative on a solution that should pass is not literally zero for Task 1, but near enough. I tried evaluating `0.9 ^ 1e4` (the chance of a particular planet of the 10 _not_ coming up, 10k times in a row, and Julia just returned 0.0. It's lucky for us that calling the same function lots of times is quick in this language, so on my machine all the tests ran in 0.9 s.

colinleach:
The website prompted me to update, and it looks OK now (except my old Annalyn solution fails the new tests. It should be fine for anyone who didn't solve it before.

colinleach:
That clears all my drafts from last October. Now I need to write new concepts, while you're away.

colinleach:
I just checked the build status page.
image

If only 3 people completed the exercise, I assume that means us plus the original author.

depial:
>The website prompted me to update, and it looks OK now

Yep, that did the trick! I thought I just had to revert to the start of the exercise, but I need to update as well.

Also, I noticed that there may be issues with white space which I'll take a closer look at and maybe open a PR to correct.

colinleach:
Also, I need to revise the `randomness` concept docs to sync with changes here - especially the new bit on the Float version of `rand(range)`.

PR #908, opened 2025-04-23
Update captains-log.jl

depial:
Fix white spacing issues. Two of the functions did not include a four spaced indent.


colinleach (approved):
Sorry, I'll need to watch out for that when porting from an R draft. The R linter enforces their 2-space standard.

Issue #910, opened 2025-04-28
Syllabus sequencing


colinleach:
See the [forum](https://forum.exercism.org/t/ordering-keys-in-dictionaries/17144/7) for background.

The short solutions to the last task of Inventory Management use one of the standard trio of
- comprehensions
- higher-order functions
- broadcasting an anonymous function

None of these is covered in the prereqs for `pairs-and-dicts`.

I think the simplest fix is to move the short section on comprehensions from `vector-operations` to `loops`.

Alternatively, we could move `pairs-and-dicts` down the syllabus. The problem is that we can't do that with everything - _something_ has to come high up!

depial:
> I think the simplest fix is to move the short section on comprehensions from `vector-operations` to `loops`.

This sounds like a reasonable idea to me. It seems to fit in the flow of the document and is already mentioned at the end. 

The only thing that does come to mind with this option is that I think the section would need a bit more detail to parallel the sections on `The while loop` and `Looping over a collection`. I don't know if it would need as much, since it is effectively a loop over a collection, but we may want to make basic concepts explicit (e.g. that these are generally used when a array is needed) since `Loops` is such a basic concept. One thought is to write it as though the student hasn't heard of comprehensions before.

> There is some divergence for multiple variables and/or multi-dimensional arrays, so this topic will be covered in more detail in a later Concept.

I assume that this would still be the case, so we wouldn't need to be comprehensive in detail in `loops`, but it might be a good place to lay a solid foundation.

colinleach:
Closed with #911

PR #911, opened 2025-04-29
Move Comprehensions section

colinleach:
Removed from `vector-operations` and added to `loops`, with some expansion. Addresses #910.


depial (approved):
lg2m!

PR #913, opened 2025-05-02
Types draft concept

colinleach:
I managed to write something, and it was no easier than I expected.

I'm a bit concerned about the length: not quite _War and Peace_, but nudging close to _Anna Karenina_.

I also have no idea how to create an exercise to go with this. That gets a bit easier for later concepts like `composite-types`, which I'll PR later, but first we need to get this introductory one in front of students' eyes. Merging them is tempting, but then the length gets even more crazy.


concepts/types/about.md
```diff
@@ -0,0 +1,268 @@
...
+ERROR: TypeError: in typeassert, expected Number, got a value of type String
+```
+
+Commonly, this might be used with the return value of a function, as a simple last check that the function behaved as expected.
+
+Note that there is an [`@assert`][assert] macro for other forms of assertion.
+
+## The Type Hierarchy
+
+`int64`, `Int16`, `String`, `Char`: where do these types "come from".
```
depial, line 88: Typo: `int64` => `Int64`

concepts/types/about.md
```diff
@@ -0,0 +1,268 @@
...
+```julia-repl
+julia> supertypes(String)
+(String, AbstractString, Any)
+
+julia> supertype(Any)
+Any
+```
+
+So, Julia has no `class` hierarchy, but it _does_ have a `type` hierarchy.
+Trying to show the entire hierarchy give a [huge tree][hierarchy-diagram], impossible to really view.
```
depial, line 137: Typo: "give" => "gives"

concepts/types/introduction.md
```diff
@@ -0,0 +1,246 @@
...
+ERROR: TypeError: in typeassert, expected Number, got a value of type String
+```
+
+Commonly, this might be used with the return value of a function, as a simple last check that the function behaved as expected.
+
+Note that there is an `@assert` macro for other forms of assertion.
+
+## The Type Hierarchy
+
+`int64`, `Int16`, `String`, `Char`: where do these types "come from".
```
depial, line 88: Typo: `int64` => `Int64`

depial:
I think this looks good so far! I don't feel it's too long as this is quite a complex and important topic. 

If you're still uncomfortable with the length, I'm sure we could whittle things down a bit when we come up with an exercise, but I quite like the `about.md` length/detail.

Also, as always, I appreciate the sprinklings of humor :)

> I also have no idea how to create an exercise to go with this. 

I feel this sentiment. I guess the first thing to think of is what we are going to test and how complicated we could make the functions... For example, if we want to keep things really simple, we could aim at the student simply instantiating certain types (e.g. `Int32` instead of `Int64`), but I think that would leave very little to do. Another thing we could do, is use types as a way to control flow in a function. For example:
```julia
if typeof(x) == Int64
    return x * 5
elseif x isa String
    return uppercase(x)
end
```
It's possible that an exercise of this type could also have the added bonus of being a prelude to the concept of multiple dispatch (i.e. at least hinting at the idea).

Any thoughts?

colinleach:
Thanks for picking up the typos, I'll fix those tomorrow.

It also strikes me that I probably didn't say enough about type promotion, and the `promote()` function. More length!

> the first thing to think of is what we are going to test 

I'd be happy with a short and simple exercise. This is mainly a reading assignment before getting into the next concepts: `composite-types`, `parametric-types`, `type-unions`, and _finally!_ `multiple-dispatch`. I _hope_ those will be easier to match to exercises.

> I appreciate the sprinklings of humor :)

In grad school, our [department head](https://en.wikipedia.org/wiki/Alan_R._Battersby) always insisted that we should approach science in a way that was "serious but not solemn". I still think that's pretty good advice.

colinleach:
I fixed the typos, and added a short section on type promotion.

> Another thing we could do, is use types as a way to control flow in a function.

I like that idea. We could present it sort and straight in the introduction, then add a forward-reference to multiple dispatch in the about.

colinleach:
The introduction is still quite long, but I can't find much I want to remove from such an important concept.

The introduction within the exercise does not need to be identical to these, so feel free to delete stuff.

PR #914, opened 2025-05-02
Composite Types concept

colinleach:
This will pair with `land-grab-in-space`, which is pretty much drafted.

Unfortunately, it follows on from `types`, which has no exercise currently (and I've no good ideas for that).


concepts/composite-types/about.md
```diff
@@ -0,0 +1,198 @@
...
+julia> isconcretetype(Person)
+true
+
+julia> fred.age
+46
+```
+
+A few points are worth noting in the above example:
+
+- No type is specified for `id`, so the compiler interprets this as `id::Any`.
```
depial, line 40: We may want to mention that [it's best not to leave a field with an abstract type](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type). Maybe this could be included as a note or parenthetical comment?

> No type is specified for `id`, so the compiler interprets this as `id::Any`. Note: in practice, it is best to help the compiler by specifying the type of a field with a concrete type or by using parameters.

This phrasing points at the `Parametric Types` concept, so we would likely have to provide a link to that or to the docs. 

I'm not sure what you think about including this information here, but I think we should try to sneak it in in one of the Type concepts, since it can be mentioned in passing while pointing at a best practice.

concepts/composite-types/about.md
```diff
@@ -0,0 +1,198 @@
...
+           ConstrainedPoint(x, y) =
+               y > x ? new(x, y) : @error("require y > x")
+       end
+
+julia> ConstrainedPoint(3, 2)
+Error: require y > x
+
+julia> ConstrainedPoint(3, 5)
+ConstrainedPoint(3, 5)
+```
```
depial, line 184: Just a thought here... Maybe it could be instructive to give a bit more detail here which could expand on the idea of hierarchical structure while still showing the main point of the inner constructor. For example:

```julia-repl
julia> abstract type AbstractPoint end;
julia> struct Point <: AbstractPoint
           x::Integer
           y::Integer
       end;
julia> struct ConstrainedPoint <: AbstractPoint
           x::Integer
           y::Integer
           
           ConstrainedPoint(x, y) =
               y > x ? new(x, y) : @error("require y > x")
       end;
julia> Point(3, 2)
Point(3, 2)
julia> ConstrainedPoint(3, 2)
Error: require y > x
julia> ConstrainedPoint(3, 5)
ConstrainedPoint(3, 5)
```

Of course "For example, we have an abstract type `Point`..." => "For example, we have an abstract type `AbstractPoint`..." in this case, but could be further revised to include something about `Point` also.

I'm not sure if you find this overkill or not though.

colinleach:
I've tried to incorporate your comments, but see what you think.

concepts/composite-types/about.md
depial: I liked the sub-list idea with the look forward to parametric types

depial (approved):
lg2m

PR #915, opened 2025-05-02
Land grab in space

colinleach:
Pairs with `composite-types`.

Based on the C# version with the same name, but I ended up making lots of changes.


colinleach:
It looks like the previous test failures were with the result vector being (randomly) out of order. As an easy fix, I switched to returning a set instead of a vector.

exercises/concept/land-grab-in-space/.docs/instructions.md
```diff
@@ -0,0 +1,92 @@
...
+julia> plot = Plot(bottom_left=Coord(10, 1), top_right=Coord(20, 2))
+Plot(Coord(0x000a, 0x0001), Coord(0x0014, 0x0002))
+
+julia> plot.top_right
+Coord(0x0014, 0x0002)
+```
+
+## 3. Check whether your claim has already been filed
+
+There is competition for these claims, so before your claim is allowed it will be checked agains existing claims in the register.
```
depial, line 41: Typo: "agains" => "against"

exercises/concept/land-grab-in-space/.docs/instructions.md
```diff
@@ -0,0 +1,92 @@
...
+julia> plot1 = Plot(bottom_left=Coord(5, 3), top_right=Coord(10, 5));
+julia> plot2 = Plot(bottom_left=Coord(0, 0), top_right=Coord(3, 3));
+julia> register = Set{Plot}([plot1, plot2]);
+
+julia> get_claim_with_longest_side(register)
+Plot(Coord(0x0005, 0x0003), Coord(0x000a, 0x0005))
+```
+
+If there is a single longest plot, return the `Plot`.
+In the event of a tie, return a vector of longest plots.
```
depial, line 92: I think both options should be returned as a "register" (i.e. as a set of `Plot`s) for consistency's sake.

exercises/concept/land-grab-in-space/.meta/exemplar.jl
```diff
@@ -0,0 +1,30 @@
...
+    x::UInt16
+    y::UInt16
+end
+
+@kwdef struct Plot
+    bottom_left::Coord
+    top_right::Coord
+end
+
+is_claim_staked(claim::Plot, register::Set{Plot}) = claim ∈ register
```
depial, line 11: I believe this may be too simplistic given the story (see comments about tests).

exercises/concept/land-grab-in-space/runtests.jl
```diff
@@ -0,0 +1,84 @@
...
+            register = Set{Plot}([plot,])
+            @test is_claim_staked(plot, register)
+        end
+
+        @testset "does not exist" begin
+            plot1 = Plot(bottom_left=Coord(1, 1), top_right=Coord(2, 3))
+            plot2 = Plot(bottom_left=Coord(1, 0), top_right=Coord(2, 5))
+            register = Set{Plot}([plot1,])
+            @test !is_claim_staked(plot2, register)
+        end
```
depial, line 37: `plot1` and `plot2` are overlapping (indeed `plot1` is a strict subset of `plot2`), so I don't think `plot2`'s claim should be allowed. To fix this, I guess we should probably change `plot2` to something non-overlapping with `plot1`.

Originally, this test made me think that the claims could be arbitrary, rather than from a predefined set of plots, so I solved it for the former (hence my comment about the solution being "simplistic"). However, I see that the instructions specify the latter, so you can ignore that comment. However, we may want to revisit the wording of that part of the instructions to avoid confusion of this type.

exercises/concept/land-grab-in-space/.docs/instructions.md
```diff
@@ -0,0 +1,92 @@
+# Instructions
+
+You have been tasked by the claims department of Isaac's Asteroid Exploration Co. to improve the performance of their land claim system.
+
+Every time a asteroid is ready for exploitation speculators are invited to stake their claim to a plot of land. 
+The asteroid's land is divided into rectangular plots, with sides parallel to the coordinate axes.
+Speculators claim the land by specifying its coordinates.
```
depial, line 7: This could be:

"Speculators claim one of the plots of land by specifying its coordinates."

exercises/concept/land-grab-in-space/.docs/hints.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## 6. Find the plot claimed that has the longest side, for research purposes.
+
+- The register is a Set, unordered but iterable.
+- It may be convenient to convert the register to something ordered, for example with `collect()`.
+- You already wrote a function to calculate the longest side of a plot, and can apply this to the whole collection.
+- See Task 5 for a note on `max` versus `maximum`.
+- Find one or more plots with the maximum length.
+- Return either a `Plot` or a `Vector{Plot}` as appropriate.
```
depial, line 50: This will need to be brought in line with the end testing decision (currently `Vector{Plot}` => `Set{Plot}`)

exercises/concept/land-grab-in-space/runtests.jl
colinleach: Oops! Brain failure...

I've also made some changes at the top of the instructions to clarify that plots are non-overlapping (and by implication, the tests can be simplistic).

colinleach:
I think I've now addressed all your points, though not at the first attempt.

I'll change this to "Ready for review", though there's no rush to approve it if you're still traveling. It's still wip, so invisible to students. We don't want unconnected concepts floating around now the syllabus is live.

exercises/concept/land-grab-in-space/runtests.jl
depial: These changes look good!

exercises/concept/land-grab-in-space/.docs/hints.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## 6. Find the plot claimed that has the longest side, for research purposes.
+
+- The register is a Set, unordered but iterable.
+- It may be convenient to convert the register to something ordered, for example with `collect()`.
+- You already wrote a function to calculate the longest side of a plot, and can apply this to the whole collection.
+- See Task 5 for a note on `max` versus `maximum`.
+- Find one or more plots with the maximum length.
+- Return the results as a `Set{Plot}`.
```
depial, line 50: On first inspection, this struck me as being potentially overly specific to the `exemplar.jl` solution, but I'm not sure if that is intended/desired. To be clear, I quite like your `exemplar.jl` solution because it shows off a lot of idiomatic Julia, so I don't mind this pushing that direction. 

For reference, when I first solved this, my main aim was to only loop over the register once, so it turned out a bit more generic:

```julia
function get_claim_with_longest_side(register::Set{Plot})::Set{Plot}
    longestsides, longestside = Set(), 0
    for claim in register
        claimmax = get_longest_side(claim)
        claimmax == longestside && push!(longestsides, claim)
        claimmax > longestside && (longestsides = Set([claim]); longestside = claimmax)
    end
    longestsides
end
```

To be fair, if we were to make the hints more generic, I'm not sure exactly how to go about it... A first thought is that I could see putting point 2 and point 3 into a sub-list under point 1 (with slightly more open wording) possibly?

- The register is a Set, unordered but iterable.
  - It may be convenient to convert the register to something ordered, for example with `collect()`.
  - You already wrote a function to calculate the longest side of a plot and could apply this to the whole collection.
- See Task 5 for a note on `max` versus `maximum`.
- Find one or more plots with the maximum length.
- Return the results as a `Set{Plot}`.

Again, I'm fine with leaving the hint as is, but I thought I'd at least mention this.

depial (approved):
This is looking good! Just one more niggle to discuss, but it could be ignored if you'd like.

exercises/concept/land-grab-in-space/.docs/hints.md
colinleach: Thanks for the suggestions. I'll have a think about this tomorrow, when I hope to be thinking more clearly.

colinleach: Getting to the point of "thinking more clearly" has been slower than I hoped!

For now, I've made the changes you suggested and will merge as wip. We can change it later if we get any better ideas.

PR #916, opened 2025-05-03
Parametric types concept

colinleach:
It's fairly preliminary at this stage. I'm hoping to wrangle `elm/treasure-chest` into an exercise that can pair with it.


concepts/parametric-types/about.md
```diff
@@ -0,0 +1,104 @@
...
+
+- Types in Julia form a hierarchy.
+- Each type has a supertype.
+- `Any` is at the top of the hierarchy, and is its own supertype.
+- Types _may_ have subtypes, and those which do are called [`abstract types`][abstract].
+- Types without subtypes (leaf nodes on the tree) are called [`concrete types`][concrete].
+- Indivisible types, such as `Char` or `Int64`, are called [`primitive types`][primitive].
+
+Only `primitive` types follow the abstract versus concrete pattern in its simple form.
+`Vector` is a collection, separable into elements with their own type.
```
depial, line 13: I feel like this transition is a bit abrupt. A possible way to make more of a connection could be something like:

> Collections, such as `Vector` or `Set` (which are not primitive types), are separable into elements with their own type.

That said, I feel like the second part of this sentence ("...are separable into elements...) becomes redundant with the two sentences following the example, starting with:

> _A collection of what?_

If the second part of the sentence were omitted, I could see the sentence being rephrased to something like:

> One example of types which are not primitive are Collections, such as `Vector` or `Set`.

PR #922, opened 2025-05-16
`multiple-dispatch` concept

colinleach:
As requested (long ago)!

I'm not entirely happy with it it the moment, but I'll read it with fresh eyes tomorrow and see if I can think of improvements.

I aim to restore `old-encounters` to wip status and get that updated. I watched the https://youtu.be/kc9HwsxE1OY video again yesterday, and noted that the exercise is lifted directly from a talk we provide a link to (with solutions in Julia and C++). Not ideal, but I don't have anything better.

### DIFF
--- a/concepts/multiple-dispatch/about.md
+++ b/concepts/multiple-dispatch/about.md
@@ -1,0 +1,205 @@
+# About
+
+By this point in the syllabus, you already saw a lot of function definitions.
+Usually, these were presented without much comment about the output in the REPL:
+
+```julia-repl
+julia> f(x) = x^2
+f (generic function with 1 method)
+```
+
+Why _"generic function with 1 method"_ ?
+
+Suppose we want to be more specific about the argument types.
+
+```julia-repl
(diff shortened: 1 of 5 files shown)


concepts/multiple-dispatch/about.md
```diff
@@ -0,0 +1,250 @@
...
+same_type (generic function with 1 method)
+
+julia> same_type([1, 2], 3)
+true
+
+julia> same_type([1, 2], "three")
+ERROR: MethodError: no method matching same_type(::Vector{Int64}, ::String)
+```
+
+The above example requires both arguments to use the same type `T`, whatever that is.
```
depial, line 174: Phrasing: "whatever that is" => "whatever it may be"

Minor and probably just a personal feeling since "whatever that is" evokes thoughts of a passive aggressive teen for me :D

colinleach: I have no recent experience of teens, so I just made the change as you suggested!

PR #926, opened 2025-05-18
Grading guru

depial:
This is a rough draft of a concept exercise meant to pair with [`Types`](https://github.com/exercism/julia/pull/921). I'm open to any suggestions as I'm not sure how much more functionality we should try to fit in or if the story is sufficiently reasonable.

Some notes:

- An attempt to use both concrete and abstract comparisons.
- Type conversion is necessary, but promotion is not.
- At a minimum, the set of { `isa`, `==`, `typeof`, `Int8`, `UInt8`} are `Type` related functions necessary to complete the exercise.
- Function stories are a bit contrived, but I was trying to avoid solutions which work with both valid data types (e.g. easy to sort both a vector and a set with the same code => `sort!(collect(vector/set))`)
- Since error handling concept hasn't been broached, a string mimicking the form of a method error is expected instead.
- `hints.md` might still need some work.
- `introduction.md` is left blank until the concept version is finalized.

Any and all thoughts are welcome!

### DIFF
--- a/exercises/concept/grading-guru/.docs/design.md
+++ b/exercises/concept/grading-guru/.docs/design.md
@@ -1,0 +1,32 @@
+# Design
+
+## Goal
+
+The goal of this exercise is to teach the student the basics of the type system in Julia.
+
+## Learning objectives
+
+- Know what types are, focused on primitive types.
+- Know how to instantiate types.
+- Understand type comparisons.
+
+## Out of scope
+
+These will be covered in following Concepts:
+
+- Type Unions
+- Parametric Types
+- Multiple Dispatch
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `types`
+
+## Prerequisites
+
+- `function-composition`
+- `vector-operations`
+- `sets`
+- `strings`
--- a/exercises/concept/grading-guru/.docs/design.md
+++ b/exercises/concept/grading-guru/.docs/design.md
@@ -30,4 +30,3 @@ The Concepts this exercise unlocks are:
 - `function-composition`
 - `vector-operations`
 - `sets`
-- `strings`
--- a/exercises/concept/grading-guru/.docs/design.md
+++ b/exercises/concept/grading-guru/.docs/design.md
@@ -16,6 +16,7 @@ These will be covered in following Concepts:
 
 - Type Unions
 - Parametric Types
+- Composite Types
 - Multiple Dispatch
 
 ## Concepts
--- a/exercises/concept/grading-guru/.docs/instructions.md
+++ b/exercises/concept/grading-guru/.docs/instructions.md
@@ -1,0 +1,30 @@
+# Instructions
+
+In this exercise, you are going to do a bit of data engineering by preprocessing some grading data.
+A school with many, many, many (too many?) students is trying to do some data analysis on the grades received by students and is looking to do so efficiently, so the idea is to minimize and organize the data.
+
+The grading scale is usually from 0-10, so it is most efficient to use a data type with only eight bits per grade.
+With that in mind, some professors use real numbers (as they are sticklers for precision), and you will need to convert these to the `UInt8` data type.
+Other professors use integers with negative numbers (to punish especially unruly students), and you will need to convert these to the `Int8` data type in these situations.
+
+Once you have implemented the function to convert the grades, you will need to write another to handle the collections that they are held in by demoting the grades and returning them in descending order.
+
+Some professors use `Vector`s sorted in ascending order to store their grades because they are meticulous about their data.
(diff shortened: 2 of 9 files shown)


depial:
Currently failing due to:

> The Concept Exercise `grading-guru` has `types` in its `concepts`, which is not a `slug` in the top-level `concepts` array

colinleach:
Great, I think this will work (with a bit of polishing). Thank you for coming up with something!

> error handling concept hasn't been broached

I'll probably add errors next, after we get the 5 type-related concepts finished. However, I'm reluctant to add yet another prereq here - I'm sure neither of us thrilled that it already takes students so long to reach the explanation of types.

It's allowed, and maybe encouraged, to provide students with just enough information in the instructions to use part of a concept which they haven't yet learned properly. So we could use a real error in the exercise, and add a note that the syntax to raise a method error is... (with an example).

> introduction.md is left blank until the concept version is finalized

I'll finish that next, now that we have (a) an exercise and (b) a better idea of how it needs to mesh into later concepts.

> Currently failing 

No problem, that will go away as soon as we merge #913.

exercises/concept/grading-guru/.docs/instructions.md
colinleach, line 1: I think we need to break up the introductory text a bit. Students seem to struggle with big paragraphs, especially if their English is patchy.

For the 2nd para, how about something like

```mardown
The grading scale is usually from 0-10, so it is most efficient to use a data type with only eight bits per grade.
However:
- Some professors use real numbers (as they are sticklers for precision), and you will need to convert these to the `UInt8` data type.
- Other professors use integers with negative numbers (to punish especially unruly students), and you will need to convert these to the `Int8` data type in these situations.
```

Could some of the early text be moved to the numbered tasks?

We definitely need to add some example code to the tasks. I always have to remind myself that students working online can't see `runtests.jl` for concept exercises, so rely on the instructions to know what is expected. This is a major contrast to practice exercises, where the tests are the final source of truth.

Incidentally, did you see the recent discussion on Discord (the insiders channel) about CLI usage vs online editor? For Julia, only 25% of student submissions use the CLI, vs 75% online. This shocked me!

exercises/concept/grading-guru/.docs/design.md
colinleach, line 1: Composite Types are also on the out-of-scope list.

depial:
> It's allowed, and maybe encouraged, to provide students with just enough information in the instructions to use part of a concept which they haven't yet learned properly. 

I hope I've done this in a reasonable way. I had initially wanted the student to return the string because it forced the use of `typeof`, but now that function is highly encouraged in Task 1, so that concern should be irrelevant.

> I think we need to break up the introductory text a bit. 

I've deleted a line and gave it parallel structure with your suggestion which I hope breaks it up and provides consistency. Let me know if you would like further changes.

> We definitely need to add some example code to the tasks. 

I've added this, but let me know it's not up to snuff.

exercises/concept/grading-guru/.docs/instructions.md
depial: > For Julia, only 25% of student submissions use the CLI, vs 75% online.

Coming from a non-coding background, this sounds about right to me :D 

I don't think I've used the CLI more than a couple of times out of curiosity. I probably should familiarize myself with it more now that I'm helping here...

depial:
I've tried to whittle down the `introduction.md` from the concept to the most relevant ideas to the exercise, without taking out too much meat. Feel free to further adjust it (up or down) as you see fit

exercises/concept/grading-guru/.docs/instructions.md
colinleach: I like the changes you made, but after looking at it again, I think we could go a bit further (to make the instructions as clear as possible).

For each task, would it be better to convert the various cases to a bulleted list? They're already nicely formating on different lines in the source file, so just needs some "- " prefixes added.

```markdown
Implement the `demote(n)` method.
- With a `Float64` input, it should round up to the nearest whole number and return a `UInt8` data type.
- With an arbitrary `Integer`, it should return the same integer in a `Int8` data type.
- All other inputs should throw a MethodError.
```

depial:
I've added bullets to the tasks in `instructions.md`.

In other news... I'm finally going to have more reliable internet next week! So, I hope to get to properly review the `Multiple Dispatch` PRs then.

colinleach:
> I'm finally going to have more reliable internet next week!

Good to hear!

I'm close to PR'ing `treasure-chest`, to pair with `parametric-types`. After that, I'll look at `role-playing-game`, which I hope will pair with `type-unions`. That should wrap up the types concepts.

I started thinking about `errors`, and belatedly realized that this needs `composite-types` to define custom errors. I'll try to find a dependency tree which is non-circular.

PR #929, opened 2025-05-21
Treasure chest concept exercise

colinleach:
Pairs with `parametric-types`.

It is, of course, rather contrived. Also, there's a step up in difficulty for task 3, after some very easy tasks. I'm not sure if the instructions and the hints are enough?

Minor detail: while I was editing the top-level `config.json`, I moved `grading-guru` up to put the types concepts in sequence, above all the deprecated stuff.

And I just noticed that I didn't add links to the hints. That will be my next push!


depial (approved):

depial:
Related to my comment in the Concept, the third task can be solved with:

```julia
TreasureChest(chest.password, fill(chest.treasure, multiplier))
```

depial:
Should we find a way to show the difference between the two options? 

One appears to be like type inference (my solution) and the other is more like type conversion (your solution).

colinleach:
It's hard to create exercises that use these comments! (and `type-unions is even worse).

I hadn't realized that type inference is so effective here. Something I need to go away and think about...

Please let me know if you have any good suggestions.

depial:
I don't have an easy answer right now, so I'll have to think about it too...

colinleach:
Crossing my fingers as I merge...

depial:
Everything is looking good on the site!

If you are looking to push the type conversion aspect, I think another test may have to be contrived... For example, we could say that integer treasures should be converted to a type of our choosing `T`. Modifying testset 3 slightly, this could look like:

```julia
    @testset "3. Multiply integer treasure" begin
        @testset "Treasure is multiplied correctly" begin
            chest = TreasureChest{Int}("password", 1)
            multiplier = 3
            expected_treasure = Int8[1, 1, 1]
            multiplied_chest = multiply_integer_treasure(multiplier, chest, Int8)
            @test typeof(multiplied_chest) == TreasureChest{Vector{Int8}}
            @test multiplied_chest.password == "password"
            @test multiplied_chest.treasure == expected_treasure
        end
    end
```
This will force conversion of some sort with (at least) the two possibilities:

```julia
function multiply_integer_treasure(multiplier, chest, T)
    TreasureChest{Vector{T}}(chest.password, fill(chest.treasure, multiplier))
end
```
or
```julia
function multiply_integer_treasure(multiplier, chest, T)
    TreasureChest(chest.password, fill(T(chest.treasure), multiplier))
end
```

I'm not sure if this does much better, since the second one still feels like cheating (albeit increasingly more contrived cheating), so I'm ambivalent about the inclusion of something of this sort and am fine with just leaving the expanded hints.

As an aside, trying to modify the above test to do something like `TreasureChest{Char}` => `TreasureChest{Vector{String}}` results in the method errors .

PR #931, opened 2025-05-29
Simple Correction in Captains Log Exercise

matheuscastro43:
Both changes are related to Task 3 of the Captain's Log exercise.

**1) Change to** `instructions.md`

In `instructions.md`, Task 3 states that the generator should produce "_a floating point number between 41000.0 and 42000.0 (inclusive)_". However, `rand()` returns a value from a continuous uniform distribution over the open interval (0, 1). Because the distribution is continuous, the probability of generating exactly any single value (including the endpoints 0 and 1) is zero.

As a result, the expression `rand() * 1000 + 41000` produces values from the interval (41000, 42000)—excluding the endpoints. Including the word “inclusive” here could be misleading, unless we intentionally implement a method to ensure inclusion of 41000 or 42000 (e.g., via a mixture model), which does not appear to be the case.

Therefore, I suggest removing the word "(inclusive)" to better reflect the behavior of the function.

Note: In Task 4, a discrete uniform distribution is used, so the use of "(inclusive)" remains appropriate there.

**2) Change to** `exemplar.jl`

Since `rand()`—and by extension `random_stardate()`—returns a scalar, the broadcast operator seems unnecessary in this context. I've removed it for clarity and simplicity.

Best regards,
Matheus Castro

### DIFF
--- a/exercises/concept/captains-log/.docs/instructions.md
+++ b/exercises/concept/captains-log/.docs/instructions.md
@@ -34,7 +34,7 @@ What's the use of a log if it doesn't include dates?
 
 A stardate is a floating point number. The adventures of the _Starship Enterprise_ from the first season of _The Next Generation_ take place between the stardates 41000.0 and 42000.0. The "4" stands for the 24th century, the "1" for the first season.
 
-Implement the function `random_stardate` that returns a floating point number between 41000.0 and 42000.0 (inclusive).
+Implement the function `random_stardate` that returns a floating point number between 41000.0 and 42000.0.
 
 ```julia-repl
 julia> random_stardate()
--- a/exercises/concept/captains-log/.docs/instructions.md
+++ b/exercises/concept/captains-log/.docs/instructions.md
@@ -34,7 +34,7 @@ What's the use of a log if it doesn't include dates?
 
 A stardate is a floating point number. The adventures of the _Starship Enterprise_ from the first season of _The Next Generation_ take place between the stardates 41000.0 and 42000.0. The "4" stands for the 24th century, the "1" for the first season.
 
-Implement the function `random_stardate` that returns a floating point number between 41000.0 and 42000.0.
+Implement the function `random_stardate` that returns a floating point number between 41000.0 (inclusive) and 42000.0.
 
 ```julia-repl
 julia> random_stardate()
--- a/exercises/concept/captains-log/.meta/exemplar.jl
+++ b/exercises/concept/captains-log/.meta/exemplar.jl
@@ -4,7 +4,7 @@ random_planet() = rand(planetary_classes)
 
 random_ship_registry_number() = "NCC-$(rand(1000:9999))"
 
-random_stardate() = rand() .* 1000 .+ 41000
+random_stardate() = rand() * 1000 + 41000
 
 random_stardate_v2() = rand(41000.0:0.1:42000.0)
 


exercises/concept/captains-log/.docs/instructions.md
```diff
@@ -34,7 +34,7 @@ What's the use of a log if it doesn't include dates?
 
 A stardate is a floating point number. The adventures of the _Starship Enterprise_ from the first season of _The Next Generation_ take place between the stardates 41000.0 and 42000.0. The "4" stands for the 24th century, the "1" for the first season.
 
-Implement the function `random_stardate` that returns a floating point number between 41000.0 and 42000.0 (inclusive).
+Implement the function `random_stardate` that returns a floating point number between 41000.0 and 42000.0.
```
depial, line 37: I believe the [`rand()`](https://docs.julialang.org/en/v1/stdlib/Random/#Base.rand) function returns a float on the half-open interval `[0, 1)`, so the interval in question would be `[41000, 42000)`, therefore I agree the word inclusive at the end of the sentence is misleading. To better reflect the interval, maybe we could change the sentence to:

> Implement the function `random_stardate` that returns a floating point number between 41000.0 (inclusive) and 42000.0.

Side note: since `rand()` returns a machine floating point number (`Float64` by default), with finite precision, I believe the probability of choosing any particular number in the interval [0, 1) is technically non-zero.

depial (changes requested):
Thanks for finding these issues!

matheuscastro43:
Indeed, after looking more carefully at the manual, it states that rand() generates values in the [0, 1) interval. You are correct — keeping “(inclusive)” only in the first part makes perfect sense.

I see your point (in the side note). My earlier comment was mainly to emphasize that, in standard mathematical and statistical notation, if either endpoint has nonzero probability, the distribution is no longer considered a continuous uniform distribution. It is possible, as in the Wikipedia page, to include the endpoints in the support, but sampling exactly 0 or 1 still has zero probability.

Note that, from a probabilistic standpoint, Uniform([0, 1)) doesn't strictly make sense, since 0 cannot occur with the same frequency as the infinitely many other values in the interval, when P(X = 0) > 0.

The exercise instructions state that rand() generates values from a uniform distribution, with no further detail. A user might naturally assume this refers to Uniform(0, 1), as defined in probability textbooks — especially if they are coming from languages like R or Matlab, where this is indeed the behavior.

At the end of the day, sampling exactly 0 should be extremely rare, but I think it would be helpful to add a brief note in the instructions for clarity.

Apologies for the lengthy response, and thank you for your attention!

depial:
I thank you for your lengthy response! It helps clarify things and (hopefully) allows me to also respond better.

Personally, I feel adding `(inclusive)` to the correct part of the sentence sufficiently helps clarify the implementation details in Julia from other languages. However, we may be able to add this level of detail to `about.md` and `introduction.md` as well.

The remaining confusion I feel may be related to the difference between theory and practice, which is likely outside the scope of the exercise (i.e. that every distribution is discrete in CS). We always have to be careful of giving too much information in an exercise which can overload students. For example, many students may not have taken a formal course in probability, so "reminding" them of  details from there may be more confusing than helpful (e.g. singling out zero seems strange since it has as much probability of being chosen as any other number).

In the end, I think this is a very minor detail in terms of this introductory `concept exercise`. However, it could be added to a more advanced exercise which could be used as a `practice exercise` for the concept.

Just to let you know, I'll be busy for the next few days and won't be able to respond, but I'll get back as soon as I can.

matheuscastro43:
I totally agree with you.

Just to be clear, when I say "add a brief comment" I mean it could be as simple as including “(inclusive)” in introduction.md as well, so the explanation is consistent in both the introduction and the exercise.

introduction.md (line 29):
Change:
"With no arguments, it generates a float between 0 and 1"
To:
"With no arguments, it generates a float between 0 (inclusive) and 1".

Side note: I didn’t look at about.md.

Issue #934, opened 2025-06-01
name-badge misnamed exercises/functions: IF required to *return*, then don't label it "print", please..


libweirdness:
"1. Print a badge for an employee"

"Implement the print_name_badge function."

"It should return the badge label, with the department name in uppercase."

IF it should RETURN a label, & not print it, then it shouldn't be named "print_badge_label".

ALL the "print_badge_label" exercise items are misleadingly worded..

"2. Print a badge for a new employee."

"When the id is missing, it should print a badge without it."

but if you println, instead of return, you fail your tests!

3. also.

_ /\ _

colinleach:
Fair point. Unfortunately, changing the function names now would break all the existing solutions: a big step.

FYI, essentially the same exercise has existed on the Elixir track for 4 years already, causing no problems that I'm aware of.

This exercise is deep in the syllabus, so I assume that most students getting this far know that all tests within Exercism need a return value

depial:
Since this is so systematic, maybe we could change the wording in the instructions? For example, something like: 

`"1. Print a badge for an employee"` => `"1. Create a badge for an employee"`.

Also, I may be able to find a work around if we want to change the names of the functions without breaking current solutions (but I'm not sure it's worth it).

depial:
I'm inclined to agree with @colinleach about not renaming the functions, especially since they are referencing a physical printing process instead of an electronic one, so I've reverted those changes. 

Hopefully I've left things a bit clearer in the PR.

PR #935, opened 2025-06-02
create name badge

depial:
A possible fix for Issue #934 

- exchanged the more specific word `print` for the more general `create` throughout.
- tried to implement backwards compatibility with old solutions, with `exemplar.jl` left with the old function name `print_name_badge` as a test.

### DIFF
--- a/exercises/concept/name-badges/.docs/hints.md
+++ b/exercises/concept/name-badges/.docs/hints.md
@@ -1,20 +1,20 @@
 # Hints
 
-The first three tasks all call the same `print_name_badge()` function.
+The first three tasks all call the same `create_name_badge()` function.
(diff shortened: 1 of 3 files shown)


colinleach:
Thanks for the attempt, but...

As discussed in #934, I'm reluctant to change the code. Altering task titles is less disruptive, and I'm fine with that.

depial:
I've gone ahead and reverted the function name change and took out the backwards compatibility fix. 

I've tried to change the word `print` to `create` in all other contexts and added some clarity that the badges are worn in the first mention of them. If we decide to proceed, please give a good look through as I caught a couple of `print`s I missed on the first go around.

colinleach (approved):

PR #941, opened 2025-06-05
Errors Concept

colinleach:
Please don't spend much time on this yet, it still needs some proofreading and polishing.

I'm aiming to port `javascript/factory-sensors` to pair with it. That isn't yet well advanced.


colinleach:
I've hacked this about a bit since the original submission. Meanwhile, `factory-sensors` is taking shape, now I've accepted that the JS original is a loose inspiration rather than something I can port directly.

I've beefed up the discussion on logging (`@info`, etc), and  that will be part of a task in the exercise. Changing the concept to "Errors and Logging" probably isn't worthwhile.

colinleach:
The Julia manual seemed to suggest that testing for log entries is simple and well-supported. I'm not having much success, so I may back out of most of what I was attempting.

depial:
> I'm not having much success, so I may back out of most of what I was attempting.

I'm not familiar with logging, so I'm afraid I can't be of much help.

I'll give you some more time to decide on what you would like to include before I review, but I'll get to this tomorrow regardless since my availability may drop off starting Wednesday.

colinleach:
I'm fine with the concept being reviewed. My problems are limited to Task 3 of `factory-sensors`: a nice-to-have, but maybe not worth this much aggravation.

I'll try to PR an exercise by the end of today, but it may be a draft-for-comment.

concepts/errors/about.md
```diff
@@ -0,0 +1,208 @@
...
+
+Some errors are necessarily fatal, but often we expect the program to recover gracefully.
+
+By default, an error immediately terminates the current function, and the error (with any informative message) is passed to the calling function.
+
+This continues up the call stack, until the top-level code terminates with an error message.
+
+At any stage, the error can be intercepted with a [`try...catch`][try-catch] block which attempts to handle it.
+
+```julia
```
depial, line 141: Should this be `julia-repl`?

concepts/errors/about.md
```diff
@@ -0,0 +1,208 @@
...
+
+```julia-repl
+julia> @warn "Something looks not quite right"
+┌ Warning: Something looks not quite right
+└ @ Main REPL[55]:1
+
+julia> @error "Panic!"
+┌ Error: Panic!
+└ @ Main REPL[56]:1
+```
```
depial, line 189: Is there a reason the formatting makes all of this text blue? I can't see anything wrong with the previous code block, but it seems a bit odd to me.

Also, you use the `@warn` and `@info` macros in the previous example. Maybe it could be good to mention them somewhere in here?

concepts/errors/about.md
```diff
@@ -0,0 +1,208 @@
...
+
+More generally, the [`@assert`][assert] macro lets us test any expression that evaluates to a boolean:
+
+```julia-repl
+julia> n = 22;
+julia> @assert isodd(n) "n must be odd"
+ERROR: AssertionError: n must be odd
+```
+
+## `try`...`catch`
```
depial, line 131: It might be nice to mention somewhere that this is rather poor in terms of performance in comparison to using `if-else` logic. I'm not sure if that could be worked in smoothly or not though.

depial (approved):
Just small things. I think this summed up the ideas nicely. 

Aside: I'm debating whether I would have put `custom errors` before `throw`, since it seems to follow from `standard error types`, but I also see the logic in the current order. I could go either way, so I just thought I'd throw that out there.

concepts/errors/about.md
depial: I forgot to mention... Do you think it could be too much information to include something on `else` and `finally`? I think they are interesting features, but I wouldn't want `about.md` getting too bulky.

concepts/errors/about.md
colinleach: > Is there a reason the formatting makes all of this text blue?

I'm not seeing it as blue, so I don't know what's happening.

> The hazards of moving the order of sections around. I've changed it, but it may still be non-ideal.

colinleach:
I'll merge for now, but yell at me if I haven't addressed all your comments. I want to see how this looks on the website, given confusion about blue text.

depial:
It looks like the blue text is gone. Maybe changing that to `julia-repl` made the difference?

depial:
> I'll merge for now, but yell at me if I haven't addressed all your comments.

The only thing I would point at is that I feel we may want to mention [`else`](https://docs.julialang.org/en/v1/manual/control-flow/#else-Clauses) clauses if we also mention `finally` clauses.

colinleach:
> we may want to mention else clauses if we also mention finally clauses

I'll submit another PR a bit later. I'm close to getting a draft of `factory-sensors` out, so I'd better finish that before I lose the mental thread.

PR #942, opened 2025-06-05
add concept exercise role-playing-game

depial:
This is a first go at the Role Playing Game adapted for Julia. I think this give some good practice with composite types, as well as about as much as we can hope for with type unions. Some notes:

- There are seven tasks, but I've tried to keep them fairly simple. Most can be solved with a ternary operator.
- Task 7 is currently unnecessary for in regards to the aims for the concept. It was left in to add more context, but could be removed
- I used the `name` and `mana` as arguments for the `increment` methods to give another chance to use the type union definitions from Task 1.
- The annotations of all the functions have been deleted, so students can be encouraged to do that themselves (for the `increment` methods at a minimum).
- The `name` method for `increment` feels a bit contrived, but I tried to keep it simple.
- `composite-types` and `nothingness` are currently the prereqs, but we can change the type requirement to move this up or down the syllabus as desired.
- status is `wip`
- instructions feel a bit long, and maybe could be compressed by deleting white space in the repl examples

In general, I feel feedback is most needed on the `instructions.md` and whether the story is too contrived with the `increment` methods. I'll be happy to give it another look if this is falling short.

Final note: I'll try to proofread the code examples in `instructions.md` more later, but this is another area that could use attention in general.

### DIFF
--- a/exercises/concept/role-playing-game/.docs/hints.md
+++ b/exercises/concept/role-playing-game/.docs/hints.md
@@ -1,0 +1,42 @@
+# Hints
+
+This exercise is centered around a keyword composite type with type unions.
+
+## 1. Define type unions
+
+- See the example where a type was named.
+
+## 2. Implement the Player composite type
+
+- All fields should be annotated, two of which are type unions
+- All fields should have the default values in the instructions
+
+## 3. Introduce yourself
+
+- A [ternary operator][control-flow] can be used for a one line solution.
+- The `ismissing` function is useful here.
+
+## 4. Implement increment methods
+
+- The arguments of the methods need to be annotated with the correct types.
+- See example of argument annotation in the introduction.
+- [Ternary operators][control-flow] can also be used here for a one line solution.
+
+## 5. Implement the title function
+
+- Use your `increment` function.
+- A [ternary operator][control-flow] can be used for a one line solution.
+
+## 6. Implement the revive function
+
+- You can use your `increment` function here.
+- Return a modified player if they were dead, or an unmodified player if they were alive.
+
+## 7. Implement the spell casting function
+
+- When doing checks on type unions with `Nothing`, it's often best to start with the function `isnothing`.
+- When the type is found to be something, mathematical checks can be done.
+- Don't forget to modify the fields in the player.
+- Return the damage of the spell in all cases.
+
+[control-flow]: https://docs.julialang.org/en/v1/manual/control-flow/#man-conditional-evaluation
--- a/exercises/concept/role-playing-game/.docs/hints.md
+++ b/exercises/concept/role-playing-game/.docs/hints.md
@@ -8,8 +8,9 @@ This exercise is centered around a keyword composite type with type unions.
 
 ## 2. Implement the Player composite type
 
-- All fields should be *annotated with types* (see `Node` example in introduction), two of which with type unions
-- All fields should have the default values in the instructions
+- All fields should be *annotated with types*, two of which with type unions.
+- See the `Node` example in the introduction for a concrete example of fields annotated with types.
+- All fields should have the default values listed in the instructions.
 
 ## 3. Introduce yourself
 
@@ -19,8 +20,9 @@ This exercise is centered around a keyword composite type with type unions.
 ## 4. Implement increment methods
 
 - The arguments of the methods need to be annotated with the correct types.
-- See example of argument annotation in the introduction.
-- [Ternary operators][control-flow] can also be used here for a one line solution.
+- See further examples of argument annotation in the introduction.
+- [Ternary operators][control-flow] can also be used here for one line solutions.
+- It may be helpful to study/review the Concept on `Multiple Dispatch` if needed.
 
 ## 5. Implement the title function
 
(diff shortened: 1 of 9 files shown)


colinleach:
First impression is that this looks great, much better than anything I came up with.

I'll go through it in more detail, and provide full feedback later today.

colinleach:
> Task 7 is currently unnecessary for in regards to the aims for the concept. It was left in to add more context, but could be removed

I'm ambivalent about this. I find this the most confusing task (also in the Gleam version), so I won't fight to keep it.

> The name method for increment feels a bit contrived, but I tried to keep it simple

TBH, most concept exercises feel contrived. I've made my peace with that feeling! The bigger issue is whether this should be explicit about multiple dispatch (which becomes a prereq), or it's OK to fudge it. I'm currently undecided.

> status is wip

The concept is now merged, so the linter will still pass if you want to make it beta. Either way, I think we'll want to decide on prereqs and task 7 before we merge.

> instructions feel a bit long, and maybe could be compressed by deleting white space in the repl examples

I much prefer to keep the white space. The only way I can see to make it significantly shorter is to lose task 7, if that's what you decide.

I'll add a couple of small nit-picks in the relevant files.

exercises/concept/role-playing-game/.docs/hints.md
```diff
@@ -0,0 +1,42 @@
...
+
+This exercise is centered around a keyword composite type with type unions.
+
+## 1. Define type unions
+
+- See the example where a type was named.
+
+## 2. Implement the Player composite type
+
+- All fields should be annotated, two of which are type unions
```
colinleach, line 11: I'm not sure all students will understand "annotated". Maybe something like "annotated with types"? These are hints, so anyone reading them may already be struggling.

depial: Initially, I did feel a bit uncomfortable with the wording here, so thanks for catching this!

I've changed the sentence to point at the `Node` example in the introduction, and put quotes around "annotated" to give the feeling that this could be a new word. I'm not sure if this is enough or not, so further thoughts are welcomed.

depial:
> The bigger issue is whether this should be explicit about multiple dispatch

I hope I'm not too long-winded here!  But, I'll give you my pedagogical reasoning on this idea.

*My initial aim was to give this as an example of multiple dispatch without explicitly telling the student it is multiple dispatch.*

**Motive**: Along a learning journey, I like to sprinkle in an aspect of "self-discovery" (where examples are presented and the student infers structure from them) to provide some balance with the (more traditional/common?) explicit introduction of structure followed by examples. I feel it can sometimes help provide a soft landing for more complex topics because this gives the student the opportunity to have seen it before having to visualize it when it's explicitly explained.

**Audience**: I feel like this aspect of the exercise, as it is currently, has the potential to give lower-intermediate students the hardest time. That's because beginner students will follow the directions naively, and get the right solution, while more advanced/motivated students will likely already be somewhat familiar enough with multiple dispatch not to question things and get it right. A lower intermediate student may not be fully aware of multiple dispatch, but might be aware of other languages where this type of behavior is not allowed, then they can over think things (i.e. L1 interference).

For me, the question is: *Where are we expecting students to be at this point?* We've mentioned multiple dispatch in previous concepts, so I don't think we can expect students to be ignorant of the term, but neither do I expect them to know how to use the concepts. For that reason, I wanted to make this part more "plug-n-chug", and not leave a big gap from instruction to implementation. Off the top of my head, I can think of two ways to do this:

1. The current gap is the wider of the two, expecting the student to fill in the function signature, but we are attempting to be explicit with instruction on its necessity and how to do it.
2. Alternatively, we could provide the functions with the arguments already annotated. Which narrows the gap further and only expects the students not to mess with the function signatures and just write the body of the method.

The second option carries the downside (or upside, depends?) that it takes away another opportunity to employ the named types from the first part, as well as "giving away" more information on how to annotate the `struct` (a relatively minor concern considering the info in `introduction.md`). I currently feel this is a downside here because I sense the students could/should/would be at a level where it is beneficial to not have that extra information. That's why I deleted all the type annotations in the stub, but maybe that points at a third way:

3. Annotate all function arguments with the exception of the two methods.

That said, I know you have more experience with syllabi on Exercism than I do, so I'm curious to see where you think the student could/should/would be?

Other options that are still on the table and should be considered:

- Move this after multiple dispatch and make the directions vaguer on how to complete that part of the exercise.
- Remove the multiple dispatch aspect (e.g. potentially stick more closely to the original design).

**Aside**: Another minor niggle for me is that I feel this part of the syllabus (type system) has been very linear, so I would like to see a branch somewhere (and type unions seems a decent place for it), but I currently have no explicit reasoning for having that feeling :D

depial:
> I'm ambivalent about this. I find this the most confusing task (also in the Gleam version), so I won't fight to keep it.

I'm starting to lean towards dropping it, and slightly modifying the story to say the protagonist has only been assigned a portion of the game.

But let's wait until we resolve the question of multiple dispatch and see where we are.

depial:
> Probably my preference, of the various possibilities.

Sounds good! I've made the following changes (the first two being your expressed preference):
- Changed the prereq to `parametric-types`
- Added the hint to finish/review `multiple-dispatch` if students are having problems
- Annotated the arguments in the functions with the exception of the `increment` methods (Option 3 from above)
- Deleted Task 7 as it was irrelevant to the concept and overly complicated.

>  I'll be out at a medical appointment this afternoon, so my online availability is uncertain.

I hope all goes smoothly! I'll see if I can give a first look at `Errors Concept` today, but there's no need to rush on anything, so don't feel the need to respond to any.

colinleach:
Sounds good. Merge when you're happy.

colinleach:
> I hope all goes smoothly!

All went smoothly, this time. A big improvement on the experience a few weeks ago, when I had a significant allergic reaction to a drug they injected (first time in my life, despite spending so much of my life working with prescription drugs).

I'm tempted to write the `bitwise-operations` concept in celebration, because it looks _really easy_ compared with all the stuff around types!

PR #943, opened 2025-06-06
Update runtests.jl

depial:
Minor typos/clarifications in the test names.


colinleach:
While we-re fixing typos, I was just looking at the `introduction.md` on the website. There's a reference to g(z::Union{Integer, String}) = z^3, which would be more readable if it was in backticks. That also needs propagating to the concept.

Do you want to fix it in this PR, or should I submit a new one?

depial:
Good catch!

colinleach:
> Good catch!

Except, I messed it up in the first place...

PR #947, opened 2025-06-08
Secrets concept exercise

colinleach:
To go with `bitwise-operations`. It's a relatively simple one, though I got caught out by the Java original implicitly assuming Int32 values. I mentioned this in the hints, not sure if it needs to be added to the instructions?

If you notice a change to Role Playing Game, `configlet fmt` updated that JSON as well as Secrets.

### DIFF
--- a/config.json
+++ b/config.json
@@ -130,6 +130,18 @@
         "status": "beta"
       },
       {
+        "slug": "secrets",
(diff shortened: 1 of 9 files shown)


depial (approved):
Looks good to me!

Issue #948, opened 2025-06-08
Practice exercise sync


colinleach:
After reading some recent forum posts, I got curious and ran `configlet sync` on the Julia repo. Apparently, we have lots of unsynced docs and missing test cases.

The docs are easy, and I'll prepare a PR for those. Updating the tests means actually work for us, since auto-generate was never really implemented (just some bits).

colinleach:
Docs are addressed in #949 

For missing test cases, this seems to be the TODO list:

```
[warn] all-your-base: missing 21 test cases
[warn] allergies: missing 1 test case
[warn] anagram: missing 8 test cases
[warn] armstrong-numbers: missing 2 test cases
[warn] bob: missing 1 test case
[warn] collatz-conjecture: missing 2 test cases
[warn] complex-numbers: missing 9 test cases
[warn] custom-set: missing 2 test cases
[warn] gigasecond: missing 1 test case
[warn] hamming: missing 6 test cases
[warn] isbn-verifier: missing 2 test cases
[warn] largest-series-product: missing 2 test cases
[warn] luhn: missing 4 test cases
[warn] matching-brackets: missing 1 test case
[warn] phone-number: missing 4 test cases
[warn] pig-latin: missing 1 test case
[warn] resistor-color-trio: missing 5 test cases
[warn] reverse-string: missing 3 test cases
[warn] roman-numerals: missing 1 test case
[warn] triangle: missing 2 test cases
[warn] word-count: missing 2 test cases
```

Looks substantial! I'm unconvinced about how urgent it is, compared with all the syllabus stuff.

depial:
I would assume this is a lower priority than syllabus stuff as well. Luckily most of them are just a couple of test cases too.

depial:
There only remains `complex-numbers`, which I would like to discuss... 

Most of the missing tests are related to arithmetic between `ComplexNumber` and regular numbers, but the instructions explicitly list what will be tested and everything is between `ComplexNumber`s. 

I assume this is because there is an additional layer of dealing with promotion that the author of the exercise didn't want to deal with. I currently see four possibilities:

1. Leave the tests as they are (i.e. set the missing tests to `include = false` in `tests.toml` to fix the configlet issue)
2. Include the tests in an `@test_skip` group (which already exists for other tests in the testset)
3. Make them `bonus` tests, which the student can activate.
4. Include them as mandatory, changing the requirements of the exercise.

I'm more leaning towards either Option 1 or Option 3. What do you think?

[...]

PR #950, opened 2025-06-09
update tests

depial:
I've updated 11 exercises from Issue #948:

- `tests.toml` have been updated
- `runtests.jl` have been updated

I would suggest we mark this as `[no important files changed]` to leave it to users to decide if they want to run the updated tests since most of the new tests are just corner cases.


colinleach:
Wow, you've been busy! Not what I expected so soon.

There seems to be a problem in `resistor-color-trio`. I fixed `@tetset` to @testset` in the final test, but only managed to make things worse!

depial:
I replaced the `example.jl` with my solution, since the original was returning `NaN` for the `minimum value` test and I didn't really want to dig into why :D

colinleach (approved):

PR #951, opened 2025-06-09
Fix misspelling "occurences".

blakelewis:
I have found my mission in life!


depial (approved):
Thank you for your service! :D

colinleach:
We applaud your mission!

PR #952, opened 2025-06-09
tests.toml and runtests.jl sync

depial:
More test sync. I've synced most of the remaining exercises. Only the exercises with 8 or more missing exercises remain, which I'll handle separately.

On `armstrong-numbers` I chose to not put the `Big` number tests in, so they are listed as `include = false` in the `tests.toml`. We can always change this if you think it's best to include them in testing.


depial:
Missed a problem with `Bob`...

depial:
Looks like the checks are passing. Only 3 more exercises remain in Issue #948

colinleach (approved):

PR #955, opened 2025-06-10
Sync tests

depial:
This updates the tests of `anagram` and `all-your-base` as per Issue #948


colinleach:
I've approved, but recommend not merging yet. There are all sorts of problems with the website today (at the AWS server level). More details on the maintaining channel of Discord. Jeremy is partly aware, but more trouble has surfaced while he's been asleep in Japan.

PR #957, opened 2025-06-12
Fix misspelling "humiditycheckk"

blakelewis:
I found the Easter egg!


colinleach:
I wish I was confident that this is the only one! I still need to sync the concept docs with the latest `introduction.md` in the exercise.

PR #960, opened 2025-06-13
Practice entries in config.json

colinleach:
With 25 concepts on the board, maybe it's time to add some more practice exercises. - or at least think about it This starter list isn't at all definitive, just something to start a conversation.

I've tried to pick out exercises which are both easy and fairly well focused on one or two concepts, rather than a whole-language workout.


depial:
Is it expected to have a non-zero intersection between `practices` and ` prerequisites`?

depial:
I think I misunderstood the structure... The concept should be finished first, which is the prerequisite, then the exercise practices the concept, right?

depial (approved):

depial:
Another (somewhat related?) thing I could float here is to add `Tuples` to the prereqs for `Composite Types`. 

I feel that the topics are related, and `Tuples` is just kinda hanging freely out there as a leaf. I'm not sure how you'd feel about  that idea, though.

colinleach:
> add Tuples to the prereqs for Composite Types

I was going to agree with this, then I looked more closely at the `config.json` and realized we have a different problem. In my head, and on the [flowchart](https://github.com/colinleach/learning-track-doodles/blob/master/julia/flowchart.md), Tuples is a prereq for functions, which is in turn a prereq for Types (and all that follows below). Somehow, we/I missed this off the entry for Locomotive Engineer.

```json
{
        "slug": "locomotive-engineer",
        "name": "Locomotive Engineer",
        "uuid": "3341044c-6dd6-4a7b-a841-3a1215cd177b",
        "concepts": [
          "functions"
        ],
        "prerequisites": [
          "vector-operations",
          "pairs-and-dicts"
        ],
        "status": "beta"
 }
```

I think the better fix is to add `tuples` here, as it should long have been!

colinleach:
You approved these. Does that mean you want them merged as-is?

PR #965, opened 2025-06-20
Reduce OS variants

colinleach:
Removes Windows and Mac-OS, leaves only Ubuntu.

I'm sure this is not the most elegant GHA code! I'll be happy if it just works, for now.


depial:
I'm likely to have better availability Tue-Fri starting next week. 

> Catching up on what I promised 4 days ago.

I'm slowly trying to get an idea formed for a rational numbers exercise, so hopefully something will solidify in the not-too-distant future ;D

PR #968, opened 2025-06-23
Test runner update

depial:
Attempts to allow auxiliary files to be accessed directly by `runtests.jl`. Loosely modeled on the Python track. Some notes:

- Allows for helper functions which are not visible to the student (e.g. for testing or for boilerplate).
- Auxiliary files (likely!) no longer need be included by student using `include("auxiliary.jl")`
- This may break solutions which have done so (via infinite loop), but I'm unsure how this will play out on the site.
- The only affected exercises are: `alphametics` = 53 solutions and `cater-waiter` = 31 solutions.

This significantly cleans up `exemplar.jl` for `aplhametics` and `cater-waiter` since the info in `permutations.jl` and  `sets_categories_data.jl`, respectively, is now available through `runtests.jl` (rather than copying it explicitly into the file).

Also, this allows for more diverse testing options, including using a correct example solution to check against the students' solutions output. That, for example, makes it easier to use parametric tests, which are randomly generated, to create larger or more complex examples than hard coding can offer.

### DIFF
--- a/exercises/concept/problematic-probabilities/runtests.jl
+++ b/exercises/concept/problematic-probabilities/runtests.jl
@@ -13,25 +13,25 @@ include("problematic-probabilities.jl")
         @test probabilities([1, 2, 3, 4], [2, 3, 4, 5]) == [1/2, 2/3, 3/4, 4/5]
     end
 
-    @testset "average check" begin
-        successes, trials = [8, 1, 4, 4, 4], [13, 24, 16, 23, 18]
+    @testset "check mean" begin
+        successes, trials = [5, 8, 6, 1, 5], [11, 16, 17, 22, 18]
         @test checkmean(successes, trials)
 
-        successes, trials = [9, 8, 6, 8, 3], [18, 15, 23, 19, 14]
+        successes, trials = [9, 2, 3, 9, 6], [23, 10, 15, 23, 10]
         @test checkmean(successes, trials)
         
-        successes, trials = [1, 9, 7, 7, 8], [13, 16, 13, 12, 21]
-        @test checkmean(successes, trials) == 3119//7280
+        successes, trials = [7, 8, 1, 4, 1], [18, 25, 19, 12, 23]
+        @test checkmean(successes, trials) == 223853//983250
 
-        successes, trials = [1, 6, 8, 6, 5], [20, 15, 18, 11, 25]
-        @test checkmean(successes, trials) == 3247//9900
+        successes, trials = [6, 3, 1, 1, 1], [12, 10, 21, 11, 10]
+        @test checkmean(successes, trials) == 2399//11550
     end
 
-    @testset "probability check" begin
+    @testset "check probability" begin
         successes, trials = [4, 2, 4, 3, 2], [18, 20, 18, 25, 22]
         @test checkprob(successes, trials)
 
-        successes, trials = [9, 1, 8, 4, 3], [14, 22, 15, 15, 20]
+        successes, trials = [2, 3, 4, 6, 2], [14, 24, 24, 16, 23]
         @test checkprob(successes, trials)
 
         successes, trials = [6, 2, 2, 9, 8], [18, 23, 12, 18, 16]


depial:
@BethanyG Oops! That was a miscue... Thanks for the heads-up!

If you happen to have time would you be able to take a quick look? I'm not completely comfortable with how everything comes together with testing, and I saw you've done a lot on the Python runner...

colinleach:
FYI, Alphametics has had 183 starts, 85 completions; Cater Waiter has had 66 starts, 52 completions. These are not huge numbers, so I'm tempted to let them rerun and see what breaks. I certainly can't predict what will happen!

It would be great to get this working. I've not looked at Alphametics, but I remember the pain around Cater Waiter, and it's currently pretty ugly.

colinleach (approved):
Provisional approval, though I'm not 100% clear about what this will do. Worth a bit more discussion, perhaps?

depial:
> Worth a bit more discussion, perhaps?

This is probably a good idea :)

While working my way to a solution here, I tried to use `include("auxiliary.jl")` in `exemplar.jl`, but CI didn't terminate and said that there was a stack overflow, so I assumed there was some sort of issue with the examplar and the stub using `include("auxiliary.jl")`. After that, I only included it in `runtests.jl`, deleting it from `exemplar.jl` and it worked. I am assuming the same thing will happen when a student's solution is run, so I will not be surprised if those solutions break.

In any event, I feel like this could be quite a useful upgrade (I've been wanting to work on this for a long time), and I think it would be good to get it implemented while we have so few solutions potentially affected.

colinleach:
> I feel like this could be quite a useful upgrade 

100% agree!

colinleach:
> I think it would be good to get it implemented while we have so few solutions potentially affected

Sometimes there are advantages in being one of the less popular tracks! For comparison, Bethany is currently dealing with someone wanting a major rewrite on `python/sublist`. That has 8,892 starts, 6,723 completions. Nobody wants to break all those!

depial:
@BethanyG I'd hate to load you with more work as I know you're busy...

How about @colinleach and I go ahead and merge this, check how the site works, and if it's breaks stuff unacceptably, revert quickly?

Can we limit any disruption that way if we're quick? :)

BethanyG:
@depial  -- I'm fine!  I just didn't want you to expect an _immediate_ review.  I am happy to help, and it pushes me to get more involved (_which I want to do here and on the R track_).

It may take me a couple of hours.  😅   

But also -- I do think that if you and Colin are fairly confident, you can do a merge and a revert if things blow up.  But you're going to have to get on the site and spot check that yourselves.  Reqjeueing solutions for retest is a background task .. so you might not see problems for a bit.  It's best to go and test directly with code you think might be problematic.

depial:
I'll go ahead and merge this with `[no important files changed]`.

Be ready for a quick revert @colinleach  ;)

depial:
However, I see that `sets_categories_data.jl` now appears in the online editor in another tab...

I would prefer that not be visible to the student... Maybe it's because I used put these files under the field `editor` in the `config.json`?

I think this might be where @BethanyG's expertise could help...

@colinleach, should we revert and wait for @BethanyG's feedback or should we just wait?

colinleach:
I just reran my solution both with and without `include("sets_categories_data.jl")` at the top. Both succeeded.

> should we revert

I don't think it's that big a deal, for today. If we can hide it later, great, but I don't think it needs an emergency revert.

colinleach:
Playing seems like the least-bad answer. I don't know much about `.meta/config.json` and how the Exercism backend interacts with it.

BethanyG:
Here are the docs for [`config.json`](https://exercism.org/docs/building/tracks/config-json).  Two things to try:

1.  Remove from `editor` and see if it still works
2.  Remove from `editor` and add to `test`.  Dunno what that would do.  

Checked the python config and we have the data files included under `editor`.

depial:
Thank you!

I removed `editor`, and used an arbitrary `auxiliary`, while modifying the relevant test runner functions. So it looks like `auxiliary` might be ignored by the backend.

PR #977, opened 2025-07-05
Booking up for Beauty concept exercise (replaces #976)

colinleach:
Goes with Dates & Times concept.

This one is particularly vigorous in directing students to the manual for details. The topic is too big to include everything in the Introduction (e.g. all the `DateFormat` codes: look them up!)

I've followed the C# version of this exercise, but we could certainly think of ways to add more tasks. It might be worthwhile, but maybe I prefer to point keen students to the practice exercises (and I think there are some new ones in problem-specifications that we could implement).

I haven't yet tried playing with `TimeZones.jl`, but if we can add that to the test runner without bloating the Docker file even more, it could be a good addition. Erik likes this stuff (not just because it shows off the strengths of C#), so there are exercises for it in other tracks.


colinleach:
We need to merge the `dates-times` concept first, to make `configlet` happy.

exercises/concept/booking-up-for-beauty/runtests.jl
```diff
@@ -0,0 +1,45 @@
+using Test
+using Dates
+
+include("booking-up-for-beauty.jl")
+
+@testset verbose = true "tests" begin
+
+    @testset "1. Parse all-number appointment date" begin
+        @test schedule_appointment("7/25/2019 13:45:00") == DateTime(2019, 07, 25, 13, 45, 0)
```
depial, line 9: I think we could add one more example where both the month and day are less than 13 . Something like:

```julia
@test schedule_appointment("7/8/2021 8:00:00") == DateTime(2021, 07, 08, 8, 00, 0)
```

exercises/concept/booking-up-for-beauty/.meta/exemplar.jl
```diff
@@ -0,0 +1,12 @@
+using Dates
+
+schedule_appointment(appointment::String) = DateTime(appointment, dateformat"m/d/y H:M:S")
```
depial, line 3: I don't think it's terribly important for the exercise, but we might be nice to define the date formats as constants in `exemplar.jl`.

```julia
const dtfm_schedule = dateformat"m/d/y H:M:S"
schedule_appointment(appointment::String) = DateTime(appointment, dtfm_schedule)

...

const dtfm_desc = dateformat"E, U d, Y at HH:MM"
describe(appointment::DateTime) = "You have an appointment on " * Dates.format(appointment, dtfm_desc)
```
(also I went with concatenation, rather than interpolation, since `format` returns a string anyway)

depial (approved):
I found it clear and straightforward to solve. Again, I'm onboard to push users to use the docs here.

As an aside, at first I expected to see the function from task 1 used in later tasks, until I stepped back and put everything into context, then I realized this is how these would work in reality.

exercises/concept/booking-up-for-beauty/.docs/introduction.md
```diff
@@ -0,0 +1,277 @@
...
+
+The same `DateFormat` can also be used to output a date/time as any desired string format, using the [`Dates.format()`][format] function.
+
+Unfortunately, there seems to be a namespace conflict with `format` (at least as of Julia 1.11.5), so it is best to include the `Dates.` prefix even after `using Dates`.
+
+```julia-repl
+julia> dt1
+2025-06-23T14:57:25.142
+
+julia> Dates.format(dt1, dateformat"e, u d")
```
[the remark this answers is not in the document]
colinleach: I updated the docs in `dates-times`, hopefully fixing the mistake when I merged that. The new `introduction.md` is now copied to this exercise. I _really hope_ I've got it all straight now!

depial:
> After some (very) minor surgery on Monday, I've been taking an antibiotic

I hope your recovery is swift! Make sure to get some good rest!

FYI, I'll probably have quite patchy availability until Wednesday, when I'll finally be back home.

PR #978, opened 2025-07-05
Fix broken links in `strings/about.md`

colinleach:
I just read this on the Exercism website, as a warmup for starting `regular-expressions`. I was alarmed to discover 3 broken links - it's surprising that nobody complained.


colinleach:
Another day, another typo to fix in the same file.

depial (approved):

PR #979, opened 2025-07-11
A couple of gremlins in the intro

blakelewis:
- One of the examples of the Times constructor was a bit mangled
- Missing link for "wiki-format"


depial (approved):
Thanks for catching these!

PR #985, opened 2025-07-15
Regular Chatbot concept exercise

colinleach:
Incomplete at the moment: I still need to add `hints.md` and `design.md`. Those will need to wait to tomorrow.

It will fail configlet in any case, until `regular-expressions` is merged.


config.json
```diff
@@ -367,6 +367,18 @@
         ],
         "status": "beta"
       },
+      {
+        "slug": "regular-chatbot",
+        "name": "regular-chatbot",
```
depial, line 372: Typo: Name should be "Regular Chatbot"

exercises/concept/regular-chatbot/.docs/hints.md
```diff
@@ -0,0 +1,41 @@
...
+
+Remember to set the flavor to PCRE2.
+
+Pre-defining each regular expression as a `const` can be helpful style (for clarity and performance).
+
+## 1. Check Valid Command
+
+- The anchor character for start of string is `^`.
+- You only need a true/false from the regex search.
+- Using [`startswith()`][startswith] is probably more efficient than a Regex, but not _quite_ in the spirit of a Concept exercise.
```
depial, line 13: I used `startswith` but I used a regex with the info from the `introduction.md` (line 42), so, I'm not sure if this could be rephrased?

```julia
is_valid_command(msg) = startswith(msg, r"chatbot"i)
```
But I realize that, at the same time, it's certainly easier just to do `startswith(lowercase(msg), "chatbot")`, so many people may gravitate to that.

exercises/concept/regular-chatbot/regular-chatbot.jl
```diff
@@ -0,0 +1,19 @@
...
+    
+end
+
+function getURL(msg)
+    
+end
+
+function nice_to_meet_you(str)
+    
+end
```
depial, line 19: These function skeletons are missing the four spaces in the body.

config.json
depial: I do it all the time too, which is probably why I check it so carefully :D

exercises/concept/regular-chatbot/regular-chatbot.jl
colinleach: I don't think so. Spaces aren't very obvious to me, but I just checked locally in VS Code and also here on the GH website, and the spaces exist.

exercises/concept/regular-chatbot/.docs/hints.md
colinleach: I'm tempted to just remove that bullet. Thoughts?

colinleach:
> I could see making the tests a bit more stringent if wanted

Sure, if you prefer. I confess I copied the JS version without thinking deeply about it!

exercises/concept/regular-chatbot/regular-chatbot.jl
depial: Okay, good to know. I've just been assuming that when I copy and paste from here to do the exercise locally that it would stay faithful to the original, but it must lose the spaces in the process.

colinleach: I can't claim any credit, it's not something I ever think about! VS Code seems to insert them automatically.

depial:
I like the agnostic hint :)

PR #987, opened 2025-07-16
Fix link in Regex docs

colinleach:
I looked at the website, and found links don't work properly within `exercism/note` admonition blocks. I've edited all the docs affected by this. I don't like the new version much, but at least the link should display correctly.


depial (approved):

colinleach:
Marking this one Draft was a mistake, caused by brain-fade.

PR #992, opened 2025-08-08
Boutique Suggestions concept exercise [DRAFT]

colinleach:
This is written for `comprehensions`, but I could imagine a version of it also fitting `higher-order-functions`.

I copied Elixir in using Dicts, then I thought it would be better with a composite type and added that code, now I'm leaning back towards Dicts, to give flexibility in building up the instructions in stages. Something to think about! I can't do much with Hints until that is decided.

~~This is supposed to be a draft PR, but GH isn't showing me the option to change it at present.~~ Fixed, once I closed and reopened it.

PR #996, opened 2025-08-19
add concept exercise boutique-suggestions

depial:
This is a redesign of `boutique-suggestions` from PR #992. Some notes:

- There are four problems, with all being solvable by different variations of comprehensions (see `exemplar.jl`).
- There are also some added tests to emphasize the column major aspect of Julia (but this will go unnoticed if simple comprehension are used naively).
- Specific concepts from `introduction.md` have been included and referenced in `hints.md`.
- It would be possible to have a final test where all the student's functions are tested in conjunction, but I don't see this as necessary.
- A helper function was used to shorten up `runtests.jl` (and ease modification), but this is not strictly necessary, so explicit `Dict`s can be reintroduced if desired.
- There is an `auxiliary` file used for the helper function, but this might not be necessary since the tests are not student facing in the first place.
- The order of the problems has been shuffled, since this felt a bit more of a natural progression.
- Prereq is set as `pairs-and-dicts` due to the use of `Dict`s.

I have to admit that I didn't take in the push for a single comprehension as a solution in the original `instructions.md`:
> While you want to give lots of suggestions, you don't want to give bad suggestions, so you decide to use a comprehension since you can easily _generate_ outfit combinations, and _filter_ them by some criteria.

I'm not adverse to going with the original idea, since it does give "real world" backing to choosing this option. I also don't mind if we go with the original idea, since the idea here is following somewhat of a template for exercises that I seem to be using, so it might not be bad to step outside of that occasionally to provide some variety. That said, I do feel this version hits is more aligned with the information given in `introduction.md`.

We can go with this one, the other one, or some combination of the two.

All/any feedback is welcome!

### DIFF
--- a/config.json
+++ b/config.json
@@ -388,7 +388,8 @@
         ],
         "prerequisites": [
           "pairs-and-dicts"
-        ]
+        ],
+        "status": "beta"
       },
       {
         "slug": "old-elyses-enchantments",
--- a/config.json
+++ b/config.json
@@ -381,7 +381,7 @@
       },
       {
         "slug": "boutique-suggestions",
-        "name": "boutique-suggestions",
+        "name": "Boutique Suggestions",
         "uuid": "e8019ab1-f82a-4fd2-bcc3-b72c460a4db2",
         "concepts": [
           "comprehensions"
--- a/config.json
+++ b/config.json
@@ -387,9 +387,10 @@
           "comprehensions"
         ],
         "prerequisites": [
-          "pairs-and-dicts"
+          "pairs-and-dicts",
+          "multi-dimensional-arrays"
         ],
-        "status": "beta"
+        "status": "wip"
       },
       {
         "slug": "old-elyses-enchantments",
--- a/config.json
+++ b/config.json
@@ -380,6 +380,17 @@
         "status": "beta"
       },
       {
+        "slug": "boutique-suggestions",
+        "name": "boutique-suggestions",
+        "uuid": "e8019ab1-f82a-4fd2-bcc3-b72c460a4db2",
+        "concepts": [
+          "comprehensions"
+        ],
+        "prerequisites": [
+          "pairs-and-dicts"
+        ]
+      },
+      {
         "slug": "old-elyses-enchantments",
         "name": "Old Elyse's Enchantments",
         "uuid": "234466a0-08ce-45a1-aebc-5ec0b0cde24f",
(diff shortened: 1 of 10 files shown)


colinleach:
First impressions are that it looks good. However, I'm really tired this afternoon, so I'll look at it with (hopefully!) more mental energy tomorrow.

config.json
```diff
@@ -379,6 +379,18 @@
...
       },
+      {
+        "slug": "boutique-suggestions",
+        "name": "Boutique Suggestions",
+        "uuid": "e8019ab1-f82a-4fd2-bcc3-b72c460a4db2",
+        "concepts": [
+          "comprehensions"
+        ],
+        "prerequisites": [
+          "pairs-and-dicts"
```
colinleach, line 390: Do we also need `multi-dimensional-arrays` as a prereq?

exercises/concept/boutique-suggestions/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
+# Hints
+
+## 1. Create clothing item
+
+- Some functions can be used with the comprehension syntax to construct them.
+- `Dict` is just such a function, review **Generator expressions** in the introduction.
+- The category at index `i` matches with the quality at index `i`
```
colinleach, line 7: I can't remember if we mentioned `eachindex()` before. Maybe worth adding a link?

config.json
depial: That's definitely a good idea. I've added that prereq, but I've changed the status to "wip" in the hopes that we could merge this and then turn it to "beta" after I get the exercise for `multi-dimensional-arrays` done. 

I'm not sure how the exercise will interact with the website when it has a prereq which isn't fully implemented... Does this seem reasonable or should we just wait to merge and deal with the conflicts then?

exercises/concept/boutique-suggestions/.docs/hints.md
depial: While not necessary, I think it's a good place to mention it's existence

colinleach:
If I wasn't involved in this discussion, I would struggle to spot the extra bit in `runtests.jl` (even on a good day).

depial:
It looks like it linked up fine on the website with `pairs-and-dicts` and I was able to solve it, so that's good news. I'll try to get to an exercise for `multi-dimensional-arrays` this week.

> If I wasn't involved in this discussion, I would struggle to spot the extra bit in runtests.jl (even on a good day).

It's good to see that all my practice trying to be cute with one-liners is finally coming in handy :D

PR #997, opened 2025-08-21
modify concept exercise exercism-matrix

depial:
This modifies the concept exercise `old-exercism-matrix` to better align with `about.md` for the Concept `multi-dimensional-arrays`. Some notes:

- The two matrix rotation tasks were removed.
- Two new tasks were included: `colpixelcount()` and `render()`.
- These last two tasks are more general and move away from exclusively using the Exercism logo matrix.
- Use of parts of most of the sections from `introduction.md` is possible/encouraged, with the exception of `Generating evenly-spaced values`.
- Status is set to "wip".
- The old UUID is being used.

During review, special attention should be given to `instructions.md` and `hints.md` to check for clarity. Testing should be sufficient, but any suggestions for added tests is welcomed.


colinleach:
Excellent news! I had no idea where to go with this, so I am (obviously!) delighted to see you rescue the situation.

First impressions positive, but I think it would be worth going through it in more detail tomorrow. It's not late in the day here, but my head feels like it far post-peak.

colinleach (approved):
Looks good! I haven't found anything to comment on.

colinleach:
I got to it in the end, after a chaotic morning (emergency trip to the window blind store after one fell apart on me; neighbor panicking about a flat battery as she's trying to leave for a medical appointment, etc...)

depial:
Cheers! I'm glad it looks like everything is under control there :)

Looks like everything is working on the website.

I noticed that `vector-operations` is a leaf in the tree, but seems like it might be a prereq for this. What do you think?

Issue #998, opened 2025-08-22
Adjust syntax in tisbury-treasure-hunt instructions


nucleic-acid:
The instructions.md file in the tisbury-treasure-hunt exercise contains examples to demonstrate the expected outcome. The current instructions for section 3 read:

``` julia
julia> compare_records("Brass Spyglass", "4B"), ("Seaside Cottages", ('1', 'C'), "blue")
false

julia> compare_records("Model Ship in Large Bottle", "8A"), ("Harbor Managers Office", ('8', 'A'), "purple")
true
```

As the functions expect to be passed two tuples, there should be surrounding brackets, like so:

``` julia
julia> compare_records(("Brass Spyglass", "4B"), ("Seaside Cottages", ('1', 'C'), "blue"))
false

julia> compare_records(("Model Ship in Large Bottle", "8A"), ("Harbor Managers Office", ('8', 'A'), "purple"))
true
```

The same issue affects section 4. If it's fine, I can create a PR with the above-mentioned minor changes in both sections.

depial:
Thanks for finding this! That PR would be much appreciated

PR #999, opened 2025-08-22
Adds brackets to examples in section 3 and 4

nucleic-acid:
Corrects function calls in examples that demonstrate the expected function behavior.


nucleic-acid:
Solves #998

depial (approved):

PR #1005, opened 2025-09-24
Add Baffling Birthdays practice exercise

colinleach:
This is one of the new ones. Adding it is clearly much less urgent than extending the syllabus (both Julia and Kotlin), but also easier. I'm still not at my best  (a course of Azithromycin helped, but less than I hoped for such a good antibiotic).

Copying both C# and Elixir, I've set difficulty 6, and haven't (yet) included a `hints.md`. Sometime, I'll check which practice exercises currently lack hints - probably _a lot!_ I did this for R recently, and didn't find a single `hints.md` (a bit depressing...)

The tests cover randomness, so I set some statistical thresholds for us to argue about.


colinleach:
Well, great - CI can't pull the Julia test-runner!  Don't blame me - I only talked about changing this, I haven't yet altered anything.

colinleach:
> I think a hints.md would be able to cover most of the hiccups I came across while completing it

I'm adding this, will push soon.

> I would actually like to push for changing the percentages to probabilities in the tests.

Changed.

> I found it strange that leap years need be excluded

I agree (sort-of), but went with the canonical data and other tracks. Practice exercises are easier to create than Concept exercises, but at the cost of giving us less freedom. This week, I'm grateful for easy!

PR #1011, opened 2025-10-04
Swift Scheduling practice exercise

colinleach:
Another workout for skills with the `Dates` module.

I've set difficulty 4, the same as Python and Java. C# and Elixir chose 6, but I guess their library modules are less supportive than Julia's. The code ends up fairly simple, but the problem needs some careful thought and will no doubt confuse some students (_educational_ - the Concept docs clearly warn them that this subject isn't as simple as some people think).

The GH infrastructure problems delayed `baffling-birthdays`, the Exercism glitch delayed `intergalactic-transmission`. Third time lucky?

### DIFF
--- a/config.json
+++ b/config.json
@@ -1384,6 +1384,18 @@
         "difficulty": 4
       },
       {
+        "slug": "swift-scheduling",
+        "name": "Swift Scheduling",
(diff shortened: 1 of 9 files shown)


depial (approved):
Looks good to me!

depial:
> Third time lucky?

Looks like it!

PR #1013, opened 2025-10-06
Nth-prime practice exercise

colinleach:
Something simpler - at least conceptually, though avoiding timeouts can be a challenge.

I'm not sure why Julia is one of the few tracks missing this exercise, it seems mathematical enough to be our sort of thing.

I went with difficulty 3, though other tracks use every value from 2 to 6.

I included a couple of example solutions, though not the slow ones. My first attempt took 9.8 seconds locally, that's now down to 0.7 s. I'm sure you'll improve it further! The exercise also taught me that Julia tuning is a million miles from NumPy tuning...


depial (approved):
LG2M!

I'm fine with the difficulty being 3. I'll be interested to see if my naive solution will pass (scanning `sqrt(n/2)` vs. your `log(n)`), but it seems snappy enough on my machine.

colinleach:
> it seems snappy enough on my machine

Yeah - 0.1s on my machine, 7x faster than my best.

depial:
I quite like your idea since it's like bringing the Sieve of Eratosthenes into the problem, in a way.

PR #1022, opened 2025-10-11
Linear Algebra Basics concept draft

colinleach:
There's no Introduction at this stage. I want to get the About fully agreed before I chop it down.

This version is pretty closely based on the first 60% of the now-closed `linear-algebra` PR #1015. Edits are mainly attempts to clean up some wording and formatting, for clarity.

### DIFF
--- a/concepts/linear-algebra-basics/about.md
+++ b/concepts/linear-algebra-basics/about.md
@@ -146,6 +146,10 @@ There is no need to specify the adjoint, as this detail is handled automatically
 ```julia-repl
 julia> using LinearAlgebra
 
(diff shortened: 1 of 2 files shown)


depial:
Sounds like a plan! I'll continue to try to think of a story in case we can use the linear regression idea for one of them. In any event, it could make a practice exercise.

The other thing I was considering doing was to port/adapt `mecha-munch-management` for the `dict-functions` concept. Have you made any headway on that or should I have a go?

depial:
I'll take a look at it soon :)

PR #1026, opened 2025-10-16
add numbers to concept testset names

depial:
Added numbers to `@testset` names in anticipation of implementing `testrunner v3`.

- added two missing `@testset`s to make the numbers match up with the tasks: one in `encounters` and the other in `factory-sensors`. 
- added a hint to `annalyns-infiltration`, which just copies the hint before, since there's not much else to say there.


depial:
We can add `[no important files changed]` to the merge-commit message since the tests added are actually redundant.

colinleach (approved):

PR #1028, opened 2025-10-16
typos in regular-chatbot hints

glennj:
Just a missing backtick and a typo.

This would be a tiny amount of rep.


colinleach (approved):

colinleach:
I seem to remember there was a time when my fingers largely did what my brain intended. That doesn't survive the ageing process!

PR #1032, opened 2025-10-21
Update introduction.md

depial:
Minor change to try to force the website to recognize the concept. It is strange that the concept initially appeared there only to disappear immediately afterwards...

We can let this sit for a bit (before trying) and see if the website falls in line on its own.


colinleach:
The [concept page](https://exercism.org/tracks/julia/concepts/dict-functions) no longer 404s, but isn't showing up on the tree yet (as you say). I'm in favor of wait and see (for various reasons, including the detail that my window cleaners just showed up and the house is unusually busy).

PR #1034, opened 2025-10-21
Update runtests.jl

depial:
Small change to try to force update.

It looks like the concept and the exercise don't see each other, but we at least got the concept on the website with a previous change of this sort, so let's try again here.


colinleach (approved):

depial:
Looks like that did the trick!

PR #1035, opened 2025-10-22
Corrected description of file_range function output

JCheney20:
Error in the instruction.md of the chessboard exercise on the julia track, stating "range of integers" for the file instead of "range of characters"


depial:
Thanks for catching this!

PR #1036, opened 2025-10-24
Fix Mecha Munch Management stub

colinleach:
The stub has the arguments in the wrong order. This changes it to align with the tests and exemplar.


depial:
Thanks! I had that issue with the original `exemplar.jl` but I let the stub slip through!

PR #1038, opened 2025-11-01
Fix broken link

colinleach:
Tiny fix to `composite-types/about.md`, to make the link name match correctly.

I spotted this when re-reading the document while tackling the `rational-numbers` practice exercise - which in the Julia version is crazy difficult and goes ***far*** beyond the problem specs. I learned a lot (after cheating and looking at Sasha's example solution), but I'm reluctant to link this to the `multiple-dispatch` concept. We could drive away a lot of students!

### DIFF
--- a/concepts/composite-types/about.md
+++ b/concepts/composite-types/about.md
@@ -163,7 +163,7 @@ Field types will be checked for compatibility with the type definition, and an e
 
 Suppose we want constraints on the _values_ passed in, not just the _types_?
 
-Then we can include an `inner constructor` within the type definition, to carry out appropriate checks prom.
+Then we can include an `inner constructor` within the type definition, to carry out appropriate checks.
 
 For example, we have an abstract type `AbstractPoint`, intended to take `(x, y)` coordinates, but want a subtype with the constraint `y > x`.
 


depial (approved):

depial:
>  but I'm reluctant to link this to the multiple-dispatch concept.

I would agree. I found the hard exercises using structs (`custom-set`, `rational-numbers`, `circular-buffer` and `complex-numbers`) to be quite challenging when I first came across them. At present, I think they all require more concepts than we have covered to be completed, with the possible exception of `complex-numbers`. The concepts I think are potentially needed for each are:
- Custom Set: interfaces (iteration) 
- Rational Numbers: promotion, conversion
- Circular Buffer: interfaces (abstract arrays)
- Complex Numbers: promotion?

So, I guess we may be able to use `complex-numbers` as a practice exercise for `multiple-dispatch` since other concepts aren't necessary to pass the tests. The bonus tests add the need to use promotion, if I'm remembering correctly.

depial:
As an aside, I clicked on the `types` concept link in the syllabus and got a 500 error. I've gone through the rest of the syllabus and all of the links work (as well as the link directly to the `types` exercise).

Are you seeing the same thing?

Issue #1040, opened 2025-11-02
500 Error for Types Concept


depial:
When clicking on the `Types` concept, to access the `about.md`, the following error is encountered:

<img width="1100" height="424" alt="Image" src="https://github.com/user-attachments/assets/22665c18-6771-4fd4-bbe5-7f8ea4faadbe" />

This also occurs when navigating through the site logged out (as a "visitor") in a private browser.

All other concepts link to their `about.md` successfully.

We've attempted a slight change to `about.md` to see if this would resolve the issue, but it hasn't. A lack of development experience prevents me from making any other sort of suggestion for now.

Any ideas?

PR #1041, opened 2025-11-02
Fix types concept about

colinleach:
Attempts to address issue #1040.

Bad formatting of a URL means invalid Markdown, so our best guess is that the Exercism website can't render it. Probably caused by careless copy-paste from `links.json`, and an author brain too fried to notice...


colinleach:
Success! The page renders and looks good now.

Such a simple, careless thing...

depial:
I had just commented to my friend the other day on the unforgiving nature of programming and how when something goes wrong, I know it's because I gave it explicit instructions to do so. However, that doesn't stop me from yelling at my machine and telling it to do otherwise! :D

PR #1042, opened 2025-11-04
adjust for addition of Combinatoric.jl

depial:
This is a fix for when `Combinatorics.jl` is added to the docker image. This should be merged after the change, and we can include `[no important files changed]` in the merge message.

Checks need to be rerun after `Combinatorics` is added, to make sure everything is backwards compatible.


depial:
`Combinatorics` is live on the website now, so I'm not sure why this is still failing... Maybe we should give it another minute and force the checks again?

Luckily, there doesn't seem to be any problem in the namespace when including `using Combinatorics` and using `permutations()` at the same time in `alphametics`.

colinleach:
The failing checks are unrelated to the test runner. I think we could modify the [workflow](https://github.com/exercism/julia/blob/main/.github/workflows/exercise-tests.yml) on line 37, appending `Pkg.add("Combinatorics")`. Admittedly, I say that without much confidence.

depial:
```yml
      - name: Install test dependencies
        run: julia --color=yes --project -e "using Pkg; Pkg.instantiate()"
```

This seems like it should be automatic, but I'm not sure either, so I'm trying to find a place that the test dependencies would be listed.

colinleach:
There's a [`Project.toml`](https://github.com/exercism/julia/blob/main/Project.toml) in the root directory. Add `Combinatorics.jl` there?

`Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"`

colinleach:
Guessing, I think this small `Project.toml` is just for adding packages that are not part of a standard installation. Statistics is in the standard library.

colinleach:
> I'm not familiar with how the workflow functions, and am operating on the (uneducated) assumption that it would use the same image as the website.

Different repos. The website pulls from `julia-test-runner`, we're on `julia` now and the [1.10, 1.12, nightly] GHA stuff is local.

depial:
I only tried to cancel the test runner, but then everything canceled behind. I'm not sure if I triggered a domino effect or if they timed out. I'll rerun them.

In the meantime, I've also noticed that line 37 from the yml is repeated in line 53 in reference to the test runner. Could that cause a conflict of some kind?

depial:
Looks like that was the solution after all!

Thanks for clarifying all that :D

colinleach:
So much spending on useful books, so little progress in reading them (and remembering what I read).

depial:
Much better than me! I don't know if I'm even brave enough to buy those :D

PR #1044, opened 2025-11-12
updates to vectors and inventory-management

depial:
This addresses some of the following suggestions from the [forum](https://forum.exercism.org/t/inventory-management-return-the-entire-content-of-the-inventory/19823/20):

- The function `remove_item()` has been renamed `remove_item!()` in `inventory-management` to emphasize use of `delete!()`.
- `hints.jl` has been updated to provide more information in this regard.
- `runtests.jl` has been updated for backwards compatibility.
- the `about.md` and `introduction.md` in the concept `Vectors` and the matching concept exercise `elyses-enchantments` have been updated to include creation of empty vectors.

I hope I've covered all the bases here, but let me know if I've missed something.

### DIFF
--- a/exercises/concept/inventory-management/.meta/exemplar.jl
+++ b/exercises/concept/inventory-management/.meta/exemplar.jl
@@ -16,6 +16,6 @@ function decrement_items(inventory, items)
     inventory
 end
 
-remove_item(inventory, item) = delete!(inventory, item)
+remove_item!(inventory, item) = delete!(inventory, item)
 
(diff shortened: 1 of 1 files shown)


depial:
We can go with `[no important files changed]` in the merge-commit body since this is simply a function name change that should be backwards compatible.

On that note, `exemplar.jl` has not be updated for the initial commit to verify backwards compatibility. I'll update that to be in line with the new stub after CI has finished.

colinleach (approved):

PR #1053, opened 2025-12-02
update hints

depial:
I've added a reference to being able to use `for loops` to solve the first task. This is to help with some students who are potentially being overwhelmed by new information and thus forgetting the basics still apply.

Also: 
- `vector` has been changed to `Vector` in `instructions.md` to reflect the type better.
- "packing, unpacking" -> "packing (i.e. slurping), unpacking (i.e. splatting)" to make the connection to information in `introduction.md` more explicit.


colinleach:
I greatly admire your patience in forum conversations, but I'm afraid I can't match it. Not even close!

PR #1055, opened 2025-12-06
Improves backwards compatibility in alphametics after addition of Combinatorics.jl

depial:
In response to suggestion by @cmcaine [here](https://github.com/exercism/julia-test-runner/issues/131#issuecomment-3619724919).

- Restores `permutations.jl` to editor files, but with the code commented out so as to avoid any namespace conflicts.
- File can be used locally, with more details in `instructions.append.md` on how to use `permutations(a, t)` in all cases.
- Replaced brute force algorithm in `example.jl` with more performant pruned dfs approach.
- `using Combnatorics` has been changed to `using Combinatorics: permutations` to keep namespace as clean as possible.

This should make the previous change backwards compatible with all solutions while not complicating the namespace with multiple definitions of `permutations(a, t)`. Also, `permutations.jl` will be readable to all students, albeit it will be lacking syntax highlighting since the code is commented out.

### DIFF
--- a/exercises/practice/alphametics/runtests.jl
+++ b/exercises/practice/alphametics/runtests.jl
@@ -73,33 +73,35 @@ include("alphametics.jl")
         )
     end
 
-    @testset "puzzle with ten letters" begin
-        @test solve("AND + A + STRONG + OFFENSE + AS + A + GOOD == DEFENSE") == Dict(
-            'A' => 5,
-            'D' => 3,
-            'E' => 4,
-            'F' => 7,
-            'G' => 8,
-            'N' => 0,
-            'O' => 2,
-            'R' => 1,
-            'S' => 6,
-            'T' => 9
-        )
-    end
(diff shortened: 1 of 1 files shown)


colinleach (approved):

cmcaine:
Did you ever find out what was going on here, or why permutations() was performing so badly?

depial:
There was an ill-advised [change](github.com/JuliaMath/Combinatorics.jl/pull/122) to the algorithm back in 2023. This killed the performance of `Combinatorics.permutations()` on larger inputs. There was new version released (v1.1.0) in mid-December, which has a roundabout correction, but it's still not as performant as the version you had put in `permutations.jl` in the `alphametics` exercise.

I've opened a [PR](https://github.com/JuliaMath/Combinatorics.jl/pull/205) there to revert these regressions while trying to further speed things up, but I was told they lack reviewers, so things don't get done very quickly (hence the time lag between the bad update in Feb 2023 and the fix in Dec 2025).

PR #1057, opened 2025-12-08
Reinstate full alphametics testing

depial:
After downgrading to v1.0.2 of Combinatorics, this fix contains:

- Reinstated the original `example.jl` since it is a good test of the `permutations()` implementation in Combinatorics and is fast on the largest test.
- Reinstated the large tests.
- Updated documentation.
- Deleted `factorial()` implementation in `permutations.jl` editor file and replaced `Base.length()` with the current implementation that doesn't use `factorial`.

Please check if documentation is clear enough. I'm afraid it could get too wordy, but I've also added some admonition blocks to help highlight important info.

### DIFF
--- a/exercises/practice/alphametics/.docs/instructions.append.md
+++ b/exercises/practice/alphametics/.docs/instructions.append.md
@@ -6,12 +6,12 @@ You may (or may not!) want to call the function `permutations(a, t)` from [Combi
 [combinatorics]: https://juliamath.github.io/Combinatorics.jl/stable/api/#Combinatorics.permutations-Tuple{Any,%20Integer}
 ~~~~
 
-- If working either in the online editor or locally with `Combinatorics.jl` installed in your environment, `permutations(a, t)` is already in the namespace, but you would normally need to add one of: `using Combinatorics` or `using Combinatorics: permutations` to your solution file.
-- If working locally, without `Combinatorics.jl` installed in your environment, you can uncomment the relevant code in `permutations.jl` and add `include("permutations.jl")` at the top of your solution file to access a version of `permutations(a, t)`. You will also need to comment out or remove the line `using Combinatorics: permutations` in `runtests.jl`.
+- If working in the online editor, you can call `permutations(a ,t)` freely as it is already in the namespace. You can see the source code in the tab labeled `permutations.jl`.
+- If working locally, add `include("permutations.jl")` to the top of your solution file to access a version of `permutations(a, t)` located there. 
 
 ~~~~exercism/caution
 
-If you use Combinatorics locally, `permutations(a, t)` in v1.0.3 is likely too slow to pass the large tests. 
-Use either v1.0.2, or a release after v1.0.3.
+If you use Combinatorics.jl v1.0.3 locally, `permutations(a, t)` is very slow on the large tests.
+Use either the provided file `permutations.jl`, Combinatorics.jl v1.0.2, or, should it exist, a release after v1.0.3.
 
 ~~~~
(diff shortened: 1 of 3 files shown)


colinleach:
I'm kind of wishing we had never opened this, but I admit it was originally my suggestion, and here we are...

depial:
> Would it be simpler just to say...

Yeah, I'm not terribly concerned about it, so we can probably just skip it, since people who are consistently setting up virtual environments are likely to be able to handle the dependencies which we made clear in the instructions. 

Plus, I feel like they are working to push a new version out over at `Combinatorics.jl`, so this probably won't be much of an issue for long anyway.

In the end, CLI students would be able to run the tests locally with v1.0.3. It would just take a long time to complete. 

> We were kind of warned.

I wouldn't quite say that as this came well after the fact, and, to me, many of the objections seem to have the emphasis on making it as easy as possible for the student to complete this particular exercise. However, I think we can start looking forward towards an emphasis on the curriculum as a whole over individual exercises at this point. It seems to me that providing helper files with bits of code from common packages is not really in line with trying to provide students with environments they may see in the real world, so I see encouraging them to interact with packages as a good thing, not to mention that it could get unwieldy providing snippets of code for any exercise that needs an outside package. If it's proprietary boilerplate, it's one thing, if it's package code, it's another.

In the end, using packages:
- Provides a more realistic experience for students.
- Scales better with the big picture in mind.

This is the only exercise we're going to have to worry about this on, so I personally think we can just bite the bullet on it.

colinleach:
> This is the only exercise we're going to have to worry about this on

Depending on whether I get my act together, and go back to debugging Zebra Puzzle. I wrote a draft of it Aug 2024, got it to a stage where it runs to completion, but bizarrely returns the wrong answer. I'm sure there's a simple fix, but by that point I was heartily sick of it and wanted to do _just about anything_ else.

Meanwhile, what do we do about this PR? My inclination is the get CI running successfully, then we can merge once we're confident that the website is using 1.0.2.

colinleach:
There's a test runner version on the Track Status page, but I have no idea what it refers to. I wish I'd recorded the previous ID, so we would know if it changed.

image

colinleach:
I just ran this in the online editor:

```julia
using Combinatorics

function hello()
    "Hello, World! $(pkgversion(Combinatorics))"
end
```
Unexpectedly, `pkgversion(Combinatorics)` returned `nothing`. This worked fine in my local REPL.

Any votes for just removing Combinatorics until 1.0.4 is available?

depial:
> Any votes for just removing Combinatorics until 1.0.4 is available?

We can do that since it seems like we've not been able to get v1.0.2 actually running.

It wouldn't be hard to just modify this PR to restore `alphametics` to what it was before.

I guess the bigger hassle is dropping `Combinatorics` everywhere else?

colinleach:
> I guess the bigger hassle is dropping Combinatorics everywhere else?

Where do we have it, besides Alphametics? I can only think of the test runner and the CI, though my thinking isn't at its sharpest right now.

colinleach:
Is there any performance hit to leaving v1.0.3 in? Either the test runner, or CI, when we're not calling it?

depial:
I did just come across this:

---

### Equality specifier

Equality can be used to specify an exact version:

```toml
[compat]
PkgA = "= 1.2.3"  # [1.2.3, 1.2.3]
```

---

Which may (or may not) explain our issue because:

---

Similar to other package managers, the Julia package manager respects [semantic versioning](https://semver.org/) (semver).
As an example, a version specifier given as e.g. `1.2.3` is therefore assumed to be compatible with the versions `[1.2.3 - 2.0.0)` where `)` is a non-inclusive upper bound.

---

So, since we have `1.0.2` it's assumed to be compatible with `[1.0.2 - 2.0.0)`. Maybe it just automatically bumps it to the latest version?

PR #1059, opened 2025-12-11
Improve readability of permutations.jl

depial:
I've tried to improve the readability of `permutations.jl` while bringing the implementation more in line with what is currently in `Combinatorics.jl/main` mostly by renaming variables, but also some other minor style changes.


depial:
We should put `[no important files changed]` in the merge-commit message.

colinleach (approved):

PR #1061, opened 2025-12-14
Linear Equation Solving Concept

colinleach:
A bit of displacement activity. I'm not making much progress on things that are important, so I did something self-indulgent instead. This has been rumbling around since early October, so it seemed about time.

PR #1015 was closed as too long. #1022 was taken from the first part of #1015, this PR is expanded from the second part.

I've only gone as far as eigenvectors here. It's already quite long, so cramming factorizations into the same document seems ... _a bit much?_

### DIFF
--- a/concepts/linear-equation-solving/about.md
+++ b/concepts/linear-equation-solving/about.md
@@ -54,7 +54,7 @@ Any values where `x = 0.75y` will be a solution.
 These `(x, y)` values lie along a line in 2-D space.
 There are an infinite number of solutions, and they form the [nullspace][wiki-nullspace] of the matrix.
 
-The rows in the matrix are not [linearly independent][indep] in this case, and the [rank][rank] of the matrix is less than the row count.
+The rows in the matrix are not [linearly independent][indep] in this case, and the [rank][rank] of the matrix is less than the row or column count.
 
 ```julia-repl
 julia> A = [4 -3; 8 -6]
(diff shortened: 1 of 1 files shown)


colinleach:
>  I did something self-indulgent instead

Stop me if you've heard this before...

concepts/linear-equation-solving/about.md
```diff
@@ -0,0 +1,388 @@
...
+ 10.720153254455274
+
+# each column is an eigenvector
+julia> F.vectors
+2×2 Matrix{Float64}:
+ -0.497417  0.945605
+  0.867511  0.325317
+```
+
+In general, an `n×n` matrix will have `n` eigenvalues, though not always _distinct_ values.
```
depial: If it fits with the flow, we could mention that Julia returns normalized eigenvectors.

depial:
> I'm not making much progress on things that are important, so I did something self-indulgent instead.

I hear you! And I see nothing wrong with that :D I got sucked into a `Permutations` rabbit hole, so I haven't gotten around to looking at approaches... I should have a bit more free time for the coming weeks, so I'll try to get on it.

colinleach:
All good suggestions! I'll make the changes tomorrow.

depial:
> would it be worth adding a Further Reading section

I think this would be a fine idea! Maybe you could tack it on to the existing PR #1064?

As far as exercises go, I'll try to come up with something, but I'm feeling, for the concept exercises, we should lean towards the less complicated side of things to help ease the "mathiness" of the subject. Practice exercises might be a bit more liberal. Any thoughts on the matter?

PR #1064, opened 2025-12-15
Linear Equation Solving add introduction

colinleach:
Oops! I forgot to fill in the `introduction.md`, so it is showing as completely blank on the website.

This just copies the About, with no attempt to shorten it at this stage.


colinleach:
I checked the website - it looks a lot better now!

PR #1065, opened 2025-12-16
Add new test from problem-specifications

depial:
This addresses Issue #1062


depial:
I'm ambivalent about retesting solutions with these updates, so I'll leave it for you to decide.

PR #1068, opened 2025-12-28
Add Satellite practice exercise

colinleach:
Not the most urgent priority, but I wanted to tick off something I started long ago. Better to do something than nothing?

Other tracks use nested dictionary structures, but I tried that and it was clunky and hard to debug. We need composite types with type unions and recursion in our lives!

I've set difficulty 7 for now, but I wonder about 8 (we're short of Hard exercises). Other tracks vary widely, generally in the 6-9 range.

We could add it to the Type Unions concept, but my feeling is that we'll scare too many students.

Also: there's no Hints file at the moment. Partly because I haven't got round to it, partly from a feeling that harder exercises should push students to find their own strategy.

### DIFF
--- a/exercises/practice/satellite/.meta/example.jl
+++ b/exercises/practice/satellite/.meta/example.jl
@@ -14,11 +14,15 @@ function tree_from_traversals(preorder, inorder)
     length(preorder) == length(unique(preorder)) || 
         throw(ArgumentError("traversals must contain unique items"))
 
+    tree(preorder, inorder)
+end
+
+function tree(preorder, inorder)
     isempty(preorder) && return(nothing)
 
     root = preorder[1]
     inx = findfirst(==(root), inorder)
-    left = tree_from_traversals(preorder[2:inx], inorder[1:inx - 1])
-    right = tree_from_traversals(preorder[inx + 1:end], inorder[inx + 1:end])
+    left = tree(preorder[2:inx], inorder[1:inx - 1])
+    right = tree(preorder[inx + 1:end], inorder[inx + 1:end])
     Tree(root, left, right)
 end
--- a/exercises/practice/satellite/runtests.jl
+++ b/exercises/practice/satellite/runtests.jl
@@ -4,8 +4,8 @@ include("satellite.jl")
 
 @testset verbose = true "tests" begin
     @testset "Empty tree" begin
-        preorder = []
-        inorder = []
+        preorder = String[]
+        inorder = String[]
         @test isnothing(tree_from_traversals(preorder, inorder))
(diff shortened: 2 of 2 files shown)


exercises/practice/satellite/.meta/example.jl
```diff
@@ -0,0 +1,24 @@
...
+
+function tree_from_traversals(preorder, inorder)
+    length(preorder) == length(inorder) || 
+        throw(ArgumentError("traversals must have the same length"))
+
+    sort(preorder) == sort(inorder) || 
+        throw(ArgumentError("traversals must have the same elements"))
+
+    length(preorder) == length(unique(preorder)) || 
+        throw(ArgumentError("traversals must contain unique items"))
```
depial: Your answer is effectively verbatim to what I have.

However, I separated the input validation from my recursive function, since, otherwise, it'll be done at every recursive call. Something like:

```julia

function tree_from_traversals(preorder, inorder)
    length(preorder) == length(inorder) || 
        throw(ArgumentError("traversals must have the same length"))

    sort(preorder) == sort(inorder) || 
        throw(ArgumentError("traversals must have the same elements"))

    length(preorder) == length(unique(preorder)) || 
        throw(ArgumentError("traversals must contain unique items"))

    tree(preorder, inorder)
end

function tree(preorder, inorder)
    isempty(preorder) && return(nothing)

    root = preorder[1]
    inx = findfirst(==(root), inorder)
    left = tree(preorder[2:inx], inorder[1:inx - 1])
    right = tree(preorder[inx + 1:end], inorder[inx + 1:end])
    Tree(root, left, right)
end
```

This probably doesn't make much of a difference to CI running times, considering the small size of the test cases, but I just thought I'd mention it.

depial:
> Also: there's no Hints file at the moment.

I feel like there is only one practical way of doing this (hence our near verbatim answers), so I guess hints would have to be for those who are pretty new to the ideas. A few I can think of:

- Use recursion
- Find a relationship between the two arrays which can reduce the problem to a similar problem, but smaller.
- Find the base case

Admittedly, those seem pretty obvious to people who already know how to use recursion, but extremely vague to those who are not well versed in the concept...

colinleach:
I figured this would appeal to the sort of people who like Julia. I'm surprised so few tracks have implemented it. My initial assumption was that it's a fairly new exercise, but apparently it was added to `problem-specifications` 7 years ago.

I made the code changes you suggested.  Still no hints file: I just can't persuade myself that this exercise would benefit, in contrast to the more beginner-oriented ones. Feel free to disagree and add something!

PR #1069, opened 2025-12-28
add hints.md

depial:
This adds a `hints.md` to the `satellite` practice exercise, while also giving it a push so the website sees it.

I'm not sure if there's relevant enough information to be included, but I guess it won't hurt?


depial:
That looks to have done the trick! It's live on the site now :)

Issue #1043, opened 2025-11-05
New Concepts: sorting and/or mutating functions


depial:
We've yet to explicitly cover [sorting and related functions](https://docs.julialang.org/en/v1/base/sort/#Sorting-and-Related-Functions), which has a decent amount of material. Some possible inclusions:

- `sort()` / `sort!()`
- `sort(x, by=function)`
- `sortperm()`
- `partialsort()`

Since there's usually a mutating variant, we could also combine the concept of mutating functions (or also just do a separate concept on that).

I guess the question is where it would come in the syllabus:
- After `Vectors` as a leaf
- After `Ranges` (since `partialsort()` can take a range as an argument), either as a leaf or a prereq for `Vector Operations`
- Somewhere else?

I'm leaning towards combining the two ideas into a single concept, just we would have to include special emphasis on explaining mutating vs non-mutating functions, the default being non-mutating, and that they can be found for inputs other than just `Vectors`.

BethanyG:
Um, was passing by and saw this.  _**Why on this green earth would either of you feel guilty?!!?**_  You have gotten **so much** done in a short time!!  The fact that someone was drawn to the syllabus, did multiple exercises, and then had questions means things are working!!  It's great that you have more concepts to add, and students who want you to do so.  I remember times past when someone would start the track and then give up because there wasn't enough of a ramp to get them started.

Sorry.  Guilty 🙄 is simply not allowed.  💙

colinleach:
As an aside: I recently completed `python/swift-scheduling` (in student mode), and had to look back at my notes for [PR 3563](https://github.com/exercism/python/pull/3563) to remind myself how `datetime.datetime` works. I really ought to finish that PR sometime, but I'm assuming that Bethany would much prefer me to do it later rather than sooner!

depial:
> Performance tips could be hard to write

I agree, since they often seem to be rather specialized in their application due to the amount of heavy lifting the Julia compiler seems to do on its own.

PR #1073, opened 2026-01-03
Sorting concept

colinleach:
An attempt to address issue #1043.

Placing this in the syllabus will be an interesting challenge. We want it fairly early, to introduce students to mutating and non-mutating functions. On the other hand, the About touches on a wild variety of more advanced concepts, which we'll need to work around.

I'm tentatively thinking the sequence will be either `Ranges -> Sorting -> Vector Operations` or `Vector Operations -> Sorting -> Functions`.

Maybe this will be clearer once we have an exercise? I haven't found anything on another track that covers specifically this topic.

### DIFF
--- a/concepts/sorting/about.md
+++ b/concepts/sorting/about.md
@@ -91,7 +91,7 @@ true
 
 Now the input `Vector` is sorted in-place, without copying, and the original ordering is lost permanently.
 
-This can be efficient, it can be buggy and dangerous, it can be confusing to other people using your code.
+This can be efficient, it can be buggy and lead to unexpected [side-effects][wiki-side-effect], it can be confusing to other people using your code.
 _Choose carefully!_
 
 [`sort!()`][ref-sort!] is a [mutating function][ref-mutating-func], and by (_very strong_) convention, any function that mutates its inputs has a `!` appended to its name, as a warning to programmers.
@@ -156,7 +156,18 @@ julia> sort(j, by=uppercase)
  'J': ASCII/Unicode U+004A (category Lu: Letter, uppercase)
  'l': ASCII/Unicode U+006C (category Ll: Letter, lowercase)
  'u': ASCII/Unicode U+0075 (category Ll: Letter, lowercase)
+
+julia> tups = [(1, 3), (5, 2), (3, 4), (4, 1)];
+
+julia> sort(tups, by=last)
+4-element Vector{Tuple{Int64, Int64, Int64}}:
+ (4, 1)
+ (5, 2)
+ (1, 3)
+ (3, 4)
 ```
+
+The final example about will be explained in more detail in the [Tuples Concept][concept-tuples].
 
 For [multi-dimensional arrays][concept-mda], we can specify a `dims` keyword to control the direction of sorting.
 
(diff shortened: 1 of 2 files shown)


concepts/sorting/about.md
```diff
@@ -0,0 +1,313 @@
...
+ "Python"
+
+julia> langs[inxs[1:2]]  # slice the indices
+2-element Vector{String}:
+ "C++"
+ "Elixir"
+```
+
+Each index in `langs` is an integer in the [range][concept-range] `1:length(langs)`.
+Output from [`sortperm()`][ref-sortperm] is a [permutation][wiki-permutation] of these integers, corresponding to a sort of the input vector.
```
depial: I don't think there's any need to talk about subtleties of indices (e.g. `eachindex()`), since it seems like information overload at this stage, but just to cover our bases, we may want to include a bit more detail without actually saying anything more. Something like:


*Because `langs` is a `Vector`, each index is an integer in the [range][concept-range] `1:length(langs)`.
Output from calling [`sortperm()`][ref-sortperm] on a `Vector` is a [permutation][wiki-permutation] of these integers, corresponding to a sort of the input `Vector`.*

concepts/sorting/about.md
```diff
@@ -0,0 +1,313 @@
...
+
+Does your program need the entire input to be sorted?
+In cases where only the first few or last few entries are interesting, it can be significantly quicker to do a [partial sort][wiki-partial-sort].
+
+Julia provides the [`partialsort()`][ref-partialsort] and [`partialsort!()`][ref-partialsort!] functions, which guarantee that the specified entry or range of entries is correct, but provides no guarantees about the remainder of the input.
+
+The second argument can be either an integer or a range, representing the index/indices of the sorted input:
+
+```julia-repl
+julia> vsq = [3, 2, 4, 1].^2  # would be [1, 4, 9, 16] after a full sort
```
depial: I think this is the only example which makes reference to the `Vector Operations` concept, so if we got rid of the broadcasting, I think we could put this concept after `Ranges`. Maybe just change this line to:

```julia-repl
julia> vsq = [9, 4, 16, 1] # would be [1, 4, 9, 16] after a full sort
```

depial (approved):
This all looks good! Just a couple of minor suggestions and a possible switch in an example.

With the exclusion of broadcasting, I think that there is enough in `Ranges` to have this follow. I'd imagine that we'll end up deleting the bits on multi-dimensional sorting in the `introduction.md` but everything else seems to be at an appropriate level.

concepts/sorting/about.md
colinleach: I had already made a similar change to the Introduction, now (belatedly) copied back to About.

depial:
> I'd already done that.

Sorry about that! I was operating under the (unjustified) assumption that `introduction.md` was still just a copy of `about.md`, so I didn't even go through it...

colinleach:
The Introduction appears on the website under my own account, and the About in incognito - nice! I'll read through them again to see what typos slipped through.

concepts/sorting/about.md
```diff
@@ -0,0 +1,313 @@
...
+# j is from "Julia", as previously
+
+julia> sort(j, by=uppercase)
+5-element Vector{Char}:
+ 'a': ASCII/Unicode U+0061 (category Ll: Letter, lowercase)
+ 'i': ASCII/Unicode U+0069 (category Ll: Letter, lowercase)
+ 'J': ASCII/Unicode U+004A (category Lu: Letter, uppercase)
+ 'l': ASCII/Unicode U+006C (category Ll: Letter, lowercase)
+ 'u': ASCII/Unicode U+0075 (category Ll: Letter, lowercase)
+```
```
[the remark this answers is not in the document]
depial: Also, there might be a stray "about" in the last line:

> The final example about will be explained in more detail in the [Tuples Concept][concept-tuples].

PR #1075, opened 2026-01-13
add ordering-ordeal concept exercise

depial:
Here's a  first shot at an exercise for the `Sorting` concept. Being so early in the syllabus, I felt my hands are a bit tied as far as the allowable functionality and, therefore, the quality of the backstory.

Some notes:
- `sort!` and `sortperm` are the targeted functions.
- All sorting is in descending order. It is not mentioned in `introduction.md` that this is possible for `sortperm` but I don't see this necessarily as an issue.
- The order in which `Task 1` is described is intentional.
- Unsure if the inverse sort permutation part of `Task 3` is too complicated, but I felt like it could be interesting.
- Current testing is fairly limited.

In particular, please pay attention to the `instruction.md` and `hints.md` to see where clarity in these can be improved.

Possible options for this:
- Use as is.
- Expand to include more functionality (e.g. more sorting functions or conditionals).
- Rewrite completely.

### DIFF
--- a/config.json
+++ b/config.json
@@ -200,7 +200,7 @@
           "vector-operations"
         ],
         "prerequisites": [
-          "ranges"
+          "sorting"
         ],
         "status": "beta"
       },


exercises/concept/ordering-ordeal/.docs/instructions.md
```diff
@@ -0,0 +1,95 @@
...
+ 6
+ 5
+```
+
+## 2. Sort customer
+
+As stated above, you need to keep track of the customers in a defined order.
+
+Write a function `sortcustomer(cust, srtperm)` which takes a `Vector` of customers `cust` and a `Vector` sort permutation `srtperm`.
+This function returns a new array with the customers in the order proscribed by the sort permutation.
```
colinleach: "proscribe" means forbid, so I assume this should be "prescribed".

exercises/concept/ordering-ordeal/.docs/hints.md
```diff
@@ -0,0 +1,22 @@
...
+## 1. Sort quantity!
+
+- Your returned sort permutation needs to be in descending order.
+- Make sure to sort the quantities in descending order in-place.
+- The order in which you perform the two above operations is important.
+
+## 2. Sort customer
+
+- A new array with the customers ordered should be returned (i.e. do not sort in-place)
+- An sort permutation (i.e. index array) can be used like indices.
```
colinleach: A sort permutation...

colinleach (approved):

depial:
> How much effort is it?

It wouldn't be too much effort to do, but I do also feel it's contrived. If two pieces of information could be in the vector, why not all three? This is essentially the problem I was having trying to find a way to incorporate it.

If you think it's worth sticking in there, I'm happy to do it, but I'm fine leaving it out. 

If you'd rather leave it out, I've changed the prereq for `Vector Operations`, so you could merge at your leisure.

colinleach:
> If you'd rather leave it out

On balance, yes. We could force it in, but only by making it ugly.

PR #1076, opened 2026-01-14
update vector-operations w/ in-place broadcasting

depial:
Since we've formally introduced in-place operations in the `Sorting` concept (the prereq for `Vector Operations`), I think it would be good to include a small bit on in-place broadcasting in this concept.

Notes:

- The section beginning with "But be careful!" is not necessary (just useful) and could be deleted if space/long-windedness is a concern.
- The examples may be a bit overly pedantic, so maybe they could be streamlined a bit?
- The last sentence of the sub-section "Programmers familiar with..." now feels a bit out of context...
- In other words, this could likely use a "style-over" by the original author ;)

Feel free to modify/edit/stylize the `about.md` as you see fit, and if/when we agree on something, I can add it to `introduction.md`.

### DIFF
--- a/concepts/vector-operations/introduction.md
+++ b/concepts/vector-operations/introduction.md
@@ -98,29 +98,10 @@ This is at the heart of `broadcasting`.
 ### Broadcasting in-place
 
 If memory usage is a concern, then in-place operations are a common way to look to reduce allocations.
+
 However, the broadcasting operations in the examples above create a new `Vector` instead of modifying the original.
 
-```julia-repl
-julia> v = [1, 2, 3]
-3-element Vector{Int64}:
- 1
- 2
- 3
-
-julia> v .+ 1
-3-element Vector{Int64}:
- 2
- 3
- 4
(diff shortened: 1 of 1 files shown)


depial:
> This relates mainly to length mismatches, so I moved it up to the previous section (before in-place broadcasting).

Just for context, I had moved that sentence below the new section mostly because I had felt there was a smoother connection via memory usage in the sentence just before it.

In other news, I've added an abridged version to the `introduction.md`. Let me know if that is suitable or not. I think the example could be further cut down to just the second part if wanted. 

I don't think it needs to go into the exercise version, though.

Feel free to edit/merge as you see fit.

depial:
I've shortened the example in `introduction.md` to just the most important bit since the document is quite concise as is and so I feel this should be too if it is to be included.

colinleach:
I think it's as good as we'll easily get it, so I went ahead and merged. Future small tweaks are not ruled out.

PR #1078, opened 2026-01-15
update to ordering-ordeal

depial:
I realized that the final function `production_schedule` should have a `!` in the name, so I've corrected that while adding backwards compatibility.

I've added a couple of modifications to `instructions.md` and `hints.md` to add more clarity. I feel there are two points that may catch students:

1. The order they complete the operations in Task 1 is important.
2. The inverse sort permutation in Task 3 can be tricky to think about.

I think the more difficult is the inverse sort permutation, so I've added a final hint which all but gives the solution. I'm not sure if it is necessary to include that yet or if I've just been spooked by the very low completion rate from the first day, so I'm open to taking it back out for now.


colinleach (approved):
nice!

PR #1081, opened 2026-01-28
Update permutations.jl

depial:
Should be non-breaking, unless students were using the default constructor instead of the defined methods `permutations`.


depial:
`[no important files changed]` can be added to the merge-commit message.

colinleach (approved):

PR #1085, opened 2026-01-30
update runtests.jl

depial:
I found a couple extraneous `Fib` references and also saw that the `runtests.jl` was not in V3 format, so I've put something together which should work.


colinleach (approved):

depial:
All looks good on the website!

Issue #1088, opened 2026-02-01
Error in test set


MagnusLilledahl:
https://github.com/exercism/julia/blob/11b959d9bdd260de4bc7c9e72bf6f31d753ef39e/exercises/concept/chessboard/runtests.jl#L28

In testset 4. of the chessboard testset, the test tests the ranks() function instead of the files() function. Thus the fourth step in the exercise is never tested.

PR #1090, opened 2026-02-02
Fix issue #1088

colinleach:
See #1088


colinleach:
477 completions so far. It's marginal if we need to re-run these. What do you think?

PR #1110, opened 2026-04-15
Update config.json

depial:
Adding some practice exercises to `multiple-dispatch` and `interfaces`.

I'm not sure if this is a great idea, since these are so difficult, so I'm open to discussion.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1681,8 +1681,13 @@
         "slug": "custom-set",
         "name": "Custom Set",
         "uuid": "b47377f7-84af-4a10-83ef-c2444a4f8940",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "multiple-dispatch",
+          "interfaces"
+        ],
+        "prerequisites": [
+          "interfaces"
+        ],
         "difficulty": 8,
         "topics": [
           "iterators",
@@ -1694,8 +1699,12 @@
         "slug": "rational-numbers",
         "name": "Rational Numbers",
         "uuid": "0caf0209-be6e-4585-a90b-e41c94c5fa40",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "multiple-dispatch"
+        ],
+        "prerequisites": [
+          "multiple-dispatch"
+        ],
         "difficulty": 8,
         "topics": [
           "math",
@@ -1707,8 +1716,12 @@
         "slug": "circular-buffer",
         "name": "Circular Buffer",
         "uuid": "3dc9bbde-d344-4a36-a7e1-433637974dfc",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "multiple-dispatch"
+        ],
+        "prerequisites": [
+          "multiple-dispatch"
+        ],
         "difficulty": 8,
         "topics": [
           "data_structures",
@@ -1720,8 +1733,12 @@
         "slug": "complex-numbers",
         "name": "Complex Numbers",
         "uuid": "eca7e62c-7bbf-436c-8011-2920cb2ab1c1",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "multiple-dispatch"
+        ],
+        "prerequisites": [
+          "multiple-dispatch"
+        ],
         "difficulty": 10,
         "topics": [
           "math",


colinleach:
Apologies, but I can't feel comfortable linking such difficult exercises to a learning syllabus. It might be (just about) OK if these were implemented in a normal way (like most other tracks), but the Julia versions really pack in an extraordinary range of challenges. I struggled with these, and in one case needed to look at the example solution on GH before I could complete it.

For context, there was some [recent discussion](http://forum.exercism.org/t/learn-vs-practice-mode/38380/9) on the forum. I agree with Isaac when he said:

> I’m thinking linked exercises should probably all be easy exercises (and on the easier side of easy) with well defined prerequisites.

That post was also supported by András, another very senior maintainer. I've not seen any disagreement.

That said, it would be good to have _something_ to add, as practice for these important concepts. Unfortunately, I've currently no idea what - let's keep thinking about it.

depial:
That makes sense to me!

BethanyG:
> That post was also supported by András, another very senior maintainer. I've not seen any disagreement.

That's because my (very) strong disagreement wasn't going to improve where the discussion was going.  And since Python is virtually the only track that's attempted to link a bunch of exercises ... well, I just noped out of there.  If Isaac and Andras want to not do that on their tracks -- they are free to not do that.

I agree that linked exercises should be more focused on practicing the concept and not as wide open as many practice exercises are, but I disagree that they should all be easy.  I think there should be a range, and the student should be encouraged to take breaks and come back if an exercise feels too hard.  But making everything always easy and nice defeats the purpose of learning and practicing.  "Drill and Kill" only gets you so far.

All that said, my current thinking is to create more purposeful/focused track-only practice exercises to link to the 
syllabus rather than "shoehorn" the current practice exercises.  But by focused, I don't mean that they are going to repeat one thing over and over.  More like they're going to have an obvious solution space that uses the techniques in the prior concepts and the current concept.  But that takes time, so shoehorn it is for now.

depial:
> All that said, my current thinking is to create more purposeful/focused track-only practice exercises to link to the
syllabus rather than "shoehorn" the current practice exercises.

I think this is what we'll have to do, at least for these more Julian topics. However, I do agree with Colin here on this particular group of exercises being a bit too extreme to link, since they may do more to demotivate than anything, hence my initial hesitance in opening the PR and my final lack of opposition to closing it.

> But making everything always easy and nice defeats the purpose of learning and practicing

I'm definitely on the side of challenging students with practice exercises. It seems a bit odd to say they've "mastered" a concept after only completing easy exercises. Also, once we are in more advanced territory (e.g. `Interfaces`), it's really not possible to create "easy" practice exercises in an absolute sense, only easy relative to the topic. 

I guess it's a matter of finding the right balance between challenge and barrier.

colinleach:
The frustration is that some of these exercises could have been suitable if they had stuck to the canonical data. Anyone getting through 30+ concepts in Julia is probably quite mathematically literate, and would find Python's versions of Complex Numbers and Rational Numbers quite quick and easy. However, someone wanted to include (and test for) every super-advanced nuance of the standard library implementation, and here we are. I figure that if I can't finish a practice exercise after writing so much of the syllabus, and can barely understand the example solution, I'm reluctant to ask a student to attempt it. 

We will need new exercises to do this properly. I don't know if any of the 49 unimplemented practice exercises could be made suitable, with a bit of tweaking - something to check. Starting from scratch with a completely new exercise is not in my skill set, even on a good day.

depial:
I've been meaning to put together something for the Linear Algebra stuff, so I'll start mulling about some ideas for other exercises here too. I don't think it'll be too difficult to think of something for multiple dispatch and maybe I could adapt some of the code I wrote for `Combinatorics.jl` for an interfaces practice exercise.

I'm hoping to finish up with the project I'm working on by the end of next week, but I'm not sure how much time I'll have before mid-May due to going to visit family. I guess it'll give me plenty of time to think of ideas :D

BethanyG:
> Python's versions of Complex Numbers and Rational Numbers quite quick and easy.

This is the pleasure and pain of practice exercises being "wide open".  Python has built-ins for Complex and a core lib for Rational.  So re-implementing them feels a tad stupid.  

So I think the whole implementation for Python ended up focusing on practicing OOP and how to overload or make dunder methods that emulate numeric types - not heavy mathematics.  And many solutions simply call complex or rational inside the class, which is silly but does pass the tests.

I've actually debated nuking the two exercises in favor of something else, but they aren't _harming_ things.  They're just not particularly helpful either.

colinleach:
Is it possible, and worthwhile, to simplify Rational Numbers?
- Make the more difficult tests optional, with the mandatory ones following `canonical-data.json`.
- Write fairly detailed hints.

This exercise is a 5 in Python, officially 8 in Julia, and I'd rate it a 12 in its current form (no matter if Exercism allows this). It practices some useful-but-advanced techniques, but the instructions are sparse and the tests are syntactically opaque.

> I'm not sure how much time I'll have before mid-May

Me too. I have visitors here May 1st-6th, and there is some remedial housework to catch up on before then (tbh, quite a lot...)

depial:
> Is it possible, and worthwhile, to simplify Rational Numbers?

This doesn't seem unreasonable and it could be a fairly quick fix. I think Complex Numbers has some extra testing that students can turn on, so we could follow that model. The only thing I am wondering about is that I think "raise a real number to a rational" would require a promotion rule, but, either way, it might not be a big deal (i.e. some research required)

I actually ended up writing some extra tests for my solution, but I was just nerding out, so probably no need to include those :D

> I've actually debated nuking the two exercises in favor of something else, but they aren't harming things.

Yeah, I would agree to leave them for those who are interested in challenging themselves (or just nerding out). 

For example, there's a solution to Rational Numbers that is just:

```julia
RationaNumber = Rational
```
But a good number of the (posted) solutions look to be done in the spirit of the exercise.

colinleach:
> many solutions simply call complex or rational inside the class, which is silly but does pass the tests.

The student's loss, and not something I worry about. I guess Julia attracts a much smaller and more self-selecting student group than Python, so I've not noticed a lot of this sort of cheat (also with R).

PR #1111, opened 2026-04-24
[Tooling] Port of `generate_tests.py` to Julia

BethanyG:
_ooof_.  This was less of an exercise in programming and more an **Epic Yak Shaving experience<sup>(TM)</sup>**

Still not totally confident in everything, hence the "draft" status.  I am also displeased with the performance, which is slowed considerably by `juliacall`.  Will experiment with alternatives, but I don't think precompiling the formatter is exactly the way we want to go.  But we might want to try using Python's `subprocess` module to call Julia directly and invoke the Formatter.  My "googling" said that would be slower ... but as it stands it feels slow to me.  So.   

We may also need to  add more functions/macros/filters as we make more complicated templates.

I am also not confident that the Formatter is really adding value. Despite passing in what I think are the correct arguments, newlines are not always stripped.  I resorted to JinJa foo to remove newlines, and so I can't really say that the formatter is going to format the way we want.  _Sigh_.  Definitely a WIP.

Added the following scripts/files:

- [x] bin/data.py
- [x]  bin/githelp.py
- [x]  bin/generate_tests.py (_modified for Julia and JuliaFormatter & using Juliacall_)
- [x]  config/generator_macros.j2 (_modified for Julia_)
- [x] exercises/practice/acronym/.meta/template.j2 (_test template for acronym_)
- [x] exercises/practice/acronym/runtests.jl (_generated test file for acronym_)
- [x] requirements.txt (Python requirements for the generator)
- [x] Project.tom (_didn't know where else to put JuliaFormatter, so it is here but commented out_)
- [x] .JuliaFormatter.toml (_config file, but I also pass YASStyle as an argument to juliacall in the generator script_)

This was an irritating experience.  I ran the tests from my Exercism Python conda env, and _despite_ Julia and JuliaFormatter being on my path, juliacall saw fit to install its own Julia into the env.  Total waste of disk space.  😡 

Open to suggestions around JuliaFormatter, juliacall, and anything else you can think of.  Oh - -and this uses Python 3.13.5, although you can probably use 3.14.

### DIFF
--- a/exercises/practice/acronym/.meta/template.j2
+++ b/exercises/practice/acronym/.meta/template.j2
@@ -7,8 +7,9 @@
     @testset "{{ case["description"]}}" begin
         @test acronym("{{ case["input"]["phrase"] }}") == "{{ case["expected"] }}"
     end
-    {%- endmacro %}
-    {%- for case in cases -%}
+    {% endmacro %}
+
+    {% for case in cases -%}
     {{ test_case(case) }}
-    {%- endfor %}
+    {% endfor %}
 end
--- a/exercises/practice/acronym/.meta/template.j2
+++ b/exercises/practice/acronym/.meta/template.j2
@@ -1,0 +1,14 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.canonical_ref() }}
+
+{{ macros.header()}}
+{%- macro test_case(case) %}
+    {%- set input = case["input"] %}
+    @testset "{{ case["description"]}}" begin
+        @test acronym("{{ case["input"]["phrase"] }}") == "{{ case["expected"] }}"
+    end
+    {%- endmacro %}
+    {%- for case in cases -%}
+    {{ test_case(case) }}
+    {%- endfor %}
+end
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style = "yas"
--- a/Project.toml
+++ b/Project.toml
@@ -4,3 +4,5 @@ JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
+#JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
+
--- a/bin/data.py
+++ b/bin/data.py
@@ -1,0 +1,381 @@
+from enum import Enum
+from dataclasses import dataclass, asdict, fields
+import dataclasses
+from itertools import chain
+import json
+from pathlib import Path
+from typing import List, Any, Dict, Type
+
+# Tomli was subsumed into Python 3.11.x, but was renamed to to tomllib.
+# This avoids ci failures for Python < 3.11.2.
+try:
+    import tomllib
+except ModuleNotFoundError:
+    import tomli as tomllib
+
+
+
+def _custom_dataclass_init(self, *args, **kwargs):
+    # print(self.__class__.__name__, "__init__")
+    names = [field.name for field in fields(self)]
+    used_names = set()
+
+    # Handle positional arguments
+    for value in args:
+        try:
+            name = names.pop(0)
+        except IndexError:
+            raise TypeError(f"__init__() given too many positional arguments")
+        # print(f'setting {k}={v}')
+        setattr(self, name, value)
+        used_names.add(name)
+
+    # Handle keyword arguments
+    for name, value in kwargs.items():
+        if name in names:
+            # print(f'setting {k}={v}')
+            setattr(self, name, value)
+            used_names.add(name)
+        elif name in used_names:
+            raise TypeError(f"__init__() got multiple values for argument '{name}'")
+        else:
+            raise TypeError(
+                f"Unrecognized field '{name}' for dataclass {self.__class__.__name__}."
+                "\nIf this field is valid, please add it to the dataclass in data.py."
+                "\nIf adding an object-type field, please create a new dataclass for it."
+            )
+
+    # Check for missing positional arguments
+    missing = [
+        f"'{field.name}'" for field in fields(self)
+        if isinstance(field.default, dataclasses._MISSING_TYPE) and field.name not in used_names
+    ]
+    if len(missing) == 1:
+        raise TypeError(f"__init__() missing 1 required positional argument: {missing[0]}")
+    elif len(missing) == 2:
+        raise TypeError(f"__init__() missing 2 required positional arguments: {' and '.join(missing)}")
+    elif len(missing) != 0:
+        missing[-1] = f"and {missing[-1]}"
+        raise TypeError(f"__init__() missing {len(missing)} required positional arguments: {', '.join(missing)}")
+
+    # Run post init if available
+    if hasattr(self, "__post_init__"):
+        self.__post_init__()
+
+
+@dataclass
+class TrackStatus:
+    __init__ = _custom_dataclass_init
+
+    concept_exercises: bool = False
+    test_runner: bool = False
+    representer: bool = False
+    analyzer: bool = False
+
+
+class IndentStyle(str, Enum):
+    Space = "space"
+    Tab = "tab"
+
+
+@dataclass
+class TestRunnerSettings:
+    average_run_time: float = -1
+
+
+@dataclass
+class EditorSettings:
+    __init__ = _custom_dataclass_init
+
+    indent_style: IndentStyle = IndentStyle.Space
+    indent_size: int = 4
+    ace_editor_language: str = "python"
+    highlightjs_language: str = "python"
+
+    def __post_init__(self):
+        if isinstance(self.indent_style, str):
+            self.indent_style = IndentStyle(self.indent_style)
+
+
+class ExerciseStatus(str, Enum):
+    Active = "active"
+    WIP = "wip"
+    Beta = "beta"
+    Deprecated = "deprecated"
+
+
+@dataclass
+class ExerciseFiles:
+    __init__ = _custom_dataclass_init
+
+    solution: List[str]
+    test: List[str]
+    editor: List[str] = None
+    exemplar: List[str] = None
+
+
+    # practice exercises are different
+    example: List[str] = None
+
+    def __post_init__(self):
+        if self.exemplar is None:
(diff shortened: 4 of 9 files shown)


colinleach:
This sounds like we inflicted unintended pain on you. Apologies!

I'm pretty frazzled at this point, so defusing the bomb will need to wait for tomorrow: another day, hopefully more awake. Please don't expect miracles...

BethanyG:
> This sounds like we inflicted unintended pain on you. Apologies!

No worries.  Now that things are set up, I expect a whole lot less drama.  But VSCode and I are NOT friends right now.

colinleach:
_Oof_, reviewing this properly is going to take a while, and a clearer head than I currently have. All in good time...

Some initial observations (beyond the obvious "thank you for doing this"):
- I think you may be worrying too much about performance. This code isn't run often enough to justify a big tuning effort: some initial conversions of existing exercises, occasional dependabot stuff, addition of some more practice exercises. Getting it right is much more important than making it fast.
- I've not used `JuliaCall` and barely knew it existed. My guess is that `PyCall` is much more developed, because Julia depends on Python libraries much more than Python depends on Julia. Age and maturity, backed by some explicit policy ("why would we devote resources to rewrite BS4 when we can just call it from Python?").
- Trivial and totally non-urgent, but at some point we need to remember to remove comments about `air`.

I'm tempted to suggest that we approve and merge this fairly quickly, then keep kicking the tyres/tires to really get a feel for it. None of this is student-facing.

colinleach:
I haven't explored this yet, but I find myself wondering how reproducible the slowness is. In Julia, the first call to any function hits a delay for JIT-compilation (serious, optimizing-compiler stuff). Subsequent calls should be much faster, though I don't know how that plays with Python.

BethanyG:
Updated the air refs (😊).  Putting it in "ready to review" status.  I can always experiment with Python subprocess on a different branch.

> Oof, reviewing this properly is going to take a while, and a clearer head than I currently have.

FWIW, this is the same code the Python track uses and the R track uses ..... with the change of `juliacall`, `JuliaFormatter`, and JinJa macros modification.  Yes - that's still a lot (more than I had to do for R), but you don't need to go over the Python code that slices and dices `problem-specs`, if that helps.

colinleach:
I'm still getting `ERROR:generator:the JuliaFormatter utility must be installed` when I try to run `generate_tests.py`

It's installed in my system Julia, it's installed through `juliapkg` in my `.venv`, yet it's not being found. 

Any other suggestions?

colinleach:
> I ran the tests from my Exercism Python conda env, and despite Julia and JuliaFormatter being on my path, juliacall saw fit to install its own Julia into the env. 

FWIW, my .`venv` seems happy to find and use `~/.juliaup/bin/julia` like everything else. It's just a pity it can't see the installed JuliaFormatter package.

As you suggested, this is very off-pissing.

colinleach:
Current workaround is to install [`jlfmt`](https://github.com/domluna/JuliaFormatter.jl#command-line-tool) and run it from the command line. Admittedly, this is some distance from the Python integration we want.

BethanyG:
AARGH.  So sorry this is getting you too.  Very frustrating.  I am trying to think what I did for JuliaFormatter that might be different.  Lemme check my paths...perhaps I added something there and then forgot?  I was able to have things called fine from the script, so **_something_** is different....  🤔

BethanyG:
Hummmm....I think I introduced a bug.  I am getting the same error you are!  😱 

_sigh_.  Hold, plz while I un-F*&k the code.......

colinleach:
No rush. My old, tired head is telling me I'm done for the day.

colinleach:
Just a question (this is not my expertise): would shelling out from Python to the command line to run `jlfmt` be significantly worse than all the `juliacall` stuff? Maybe simpler, but it I guess it might really mess with error reporting?

BethanyG:
> Just a question (this is not my expertise): would shelling out from Python to the command line to run jlfmt be significantly worse than all the juliacall stuff? Maybe simpler, but it I guess it might really mess with error reporting?

I would have to try it using `subprocess`.  From what I can read, it has the effect of all the "startup" stuff for Julia every time you do it (plus compliation).  However, I am no expert here - the proof would be in actually testing it.

BTW:  fixed the issue.  I will push an update here in a sec.  I had named the check for the app "JuliaFormatter", but it turns out it needs to be "jlfmt" -- since that's the app that has the CLI.  So the CLI _as well as_ the lib need to be installed.  Sorry  about that!

BethanyG:
I have to add that after this ride on the struggle bus, I find it a tad offensive that JuliaFormatter pats itself on the back:

image

_with emoji_ no less!

colinleach:
Nope, I'm still getting the same error.

Something that will need to wait till tomorrow.

I'll approve it meanwhile, without merging. I'd like to give @depial a chance to comment (if he's around).

BethanyG:
Shelling out works and it's not slower.  But it is much simpler - so expect an update here shortly.  Extra bonus because we wouldn't need a config file at all this way - we just need to figure out the options we want to pass on to the formatter CLI!  🎉 

The only thing that's now bugging me is that the newline removal is really aggressive - so we'll have to tune it probably.

colinleach (approved):
Provisionally approved, accepting that there are still some formatting tweaks to sort out.

BethanyG:
> Nope, I'm still getting the same error.

That is seriously weird.  Can't figure out what might be the difference.  I am going to push an update to do the shell out instead (but I'll leave the Juliacall code in commented out).  Let's not merge until we figure it out.

colinleach:
I think it's a $PATH problem at my end. My `bashrc` adds `~/.julia/bin` as this contains `jlfmt`, but the `.venv` is not seeing it (but is seeing `.juiliaup/bin`, which contains the Julia starter).

All fixable, I'm sure. Just not today.

BethanyG:
> I think it's a $PATH problem at my end. My bashrc adds ~/.julia/bin as this contains jlfmt, but the .venv is not seeing it (but is seeing .juiliaup/bin, which contains the Julia starter).

> All fixable, I'm sure. Just not today.

I am using `.bash_profile`, as well as `bashrc` and `.profile` (_which I think might be a mac thing_).  Anyways, juliaup wrote into all three.  So ... it might be a difference between a login shell and a session shell?  Not sure.  Good speed tho, and I hope you feel better and get some good rest!

I should be around tomm, should you find more bugs.  😄

depial:
Sorry... It's getting down to crunch time for me here, so I'm not likely to have much time to review thoroughly. Let me see how much I can get done on what I'm working on today, and I'll see if I can start exploring soon. Of course, since you feel ready to merge, feel free to do so. I agree we can work out any details/bugs as we go.

In the meantime, I did take a quick look and noticed:
- I see the template is in line with the requirements for verbose testing (w/ the wrapping testset macro), so that's a go. 
- There's less white space than I'd expect (e.g. no newline after a testset), but I don't really mind (there's also the GitHub complaint about no new lines at the end of some files, which I also don't really care about).
- I'm unsure how we will deal with auxiliary files, but [`alphametics`](https://github.com/exercism/julia/blob/main/exercises/practice/alphametics/runtests.jl) is really a one-off and we should only use aux files with custom/concept exercises, rather than canonical exercises, so this concern can be ignored.

depial (approved):

colinleach:
> It's getting down to crunch time for me here

Good luck!

> There's less white space than I'd expect

Yes, kind of a known problem. We need to think how we might improve it, but meanwhile it's not a deal-breaker.

When I get chance, I'll restart a few things on my machine and see if I can get it working (with everything findable on a suitable path). Given that, I hope to merge later today.

colinleach:
Working now, I'm happy to say!

I'll press the Big Green Button once @BethanyG gives the OK (unless Erik manages to add maintainer rights before then).

colinleach:
Tuning the formatter might be the cleanest solution, but if that goes badly one fallback is a quick regex replace. For unnested tests, we just need a newline added after `end` and before an indented `@testset`.  I'd need to play more to see how this plays with nesting - first impressions are it's OK.

BethanyG:
> Tuning the formatter might be the cleanest solution, but if that goes badly one fallback is a quick regex replace. 

I think we can also turn off the newline removal in JuliaFormatter by passing `remove_extra_newlines=false`.  But then we need to be really careful about the formatting codes we use with JinJa.  That's pretty frustrating, but we could probably do a master template (or series) that we'd only have to think through once.  YMMV.

colinleach:
Should I merge meanwhile, or would you prefer to wait?

BethanyG:
I have some stuff to do this morning, but can make that change this afternoon.  If you'd like to experiment, the line that calls JuliaFormatter is **296** in `generate_tests.py`.  You can either add `"remove_extra_newlines=false"` as an argument after "-v", or remove `"--style=yas"`

The template I made for `acronym` has the newline formatting in it, so that is a starting example of what we can do in JinJa.  Be warned tho:  its very frustrating/fiddly to work with, but is probably less irritating than getting a regex correct.

BethanyG:
> Should I merge meanwhile, or would you prefer to wait?

Let's wait a couple hours.  I want to make sure things are a little tighter, and it's easier for me to push changes before we merge - if that's OK?

exercises/practice/acronym/.meta/template.j2
```diff
@@ -0,0 +1,14 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.canonical_ref() }}
+
+{{ macros.header()}}
+{%- macro test_case(case) %}
+    {%- set input = case["input"] %}
+    @testset "{{ case["description"]}}" begin
+        @test acronym("{{ case["input"]["phrase"] }}") == "{{ case["expected"] }}"
+    end
+    {%- endmacro %}
```
colinleach: Suggested change: `{%- endmacro` to `{% endmacro` to create a blank line between testsets.

colinleach:
As far as I can tell, `"remove_extra_newlines=false"` is the default. I got good results by removing one `-` in the template to actually create the blank lines.

Delightfully simple, _if true_.

BethanyG:
Apologies - I didn't feel well after lunch, so had a nap.  This is the latest (and last, unless you have issues) iteration.  Happy to have this one merged while we sus out any other bugs.  Removed the Juliacall code and imports and adjusted the `acronym` template.

exercises/practice/acronym/.meta/template.j2
BethanyG: Yup - that helped.  😄   As did adjusting the format call in `generate_tests.py`.  Looks like `JuliaFormatter` will remove one excess newline (so if there are two, it will remove one).  But it won't add newlines ... so if the template is too aggressive, then it will leave it.

colinleach:
Pressing the button...

And I hope you're feeling better.

PR #1115, opened 2026-05-25
Format all `instructions.append.md` to start with an H1 (required but ignored) and H2

IsaacG:
See related discussions [here](https://forum.exercism.org/t/grep-instructions-append-may-need-a-visible-header/53923) as well as https://github.com/exercism/docs/pull/592. Append files should all have an H1. Including an H2 helps break the flow from the non-append file to the append file so it does not read as the same stream of thought.

### DIFF
--- a/exercises/practice/accumulate/.docs/instructions.append.md
+++ b/exercises/practice/accumulate/.docs/instructions.append.md
@@ -1,5 +1,7 @@
 # Instructions append
 
+## Implementation
+
 You can use `collect(1:10)` as a quick way to create a collection but remember that
 under the hood, collection returns an array so you can also just make an array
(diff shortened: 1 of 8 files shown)


colinleach (approved):

colinleach:
Merging this triggered a CI run, which gave a couple of errors.
- Only with Julia 1.10 LTS, not the current 1.12
- Only in `captains-log`, which is part of the Randomness concept and explicitly non-deterministic.
- Unrelated to the Markdown files which just changed.

We can have a discussion sometime about revising thresholds on the tests (again), but for now I'm not worried about this. Past experience suggests that everything will be fine if we re-run CI.

PR #1118, opened 2026-06-09
Update instructions for locomotive-engineer

D3usXMachina:
Correct the julia-repl examples for task 4


depial (approved):
Thanks for catching this!

PR #1122, opened 2026-06-29
Fix atbash approach markdown

BNAndras:
See https://forum.exercism.org/t/incorrect-markdown-on-atbash-cipher-approach/62074. Mark PR as tiny rep if you don't mind.


colinleach (approved):

colinleach:
Thanks.

PR #1125, opened 2026-07-18
state-of-tic-tac-toe: sync tests and fix example.jl win/turn parity bug

Konovor:
# state-of-tic-tac-toe: Sync tests for playing after a win ([exercism/problem-specifications#2667](https://github.com/exercism/problem-specifications/pull/2667))
Sync `tests.toml` and `runtests.jl` with the two new canonical test cases that verify an error is thrown when a player makes a move after the game has already been won:

* O kept playing after X wins
* X kept playing after O wins

# Fix example.jl: enforce win/turn-count parity
The reference solution failed both new test cases — it validated turn order and rejected simultaneous double wins, but never checked that a win matched the expected move count, so a board where the wrong player had an extra move after a win was incorrectly accepted. Added the missing parity check (`X winning requires sum(M) == 1`, `O winning requires sum(M) == 0`). The three alternative implementations kept as comments shared the same gap and were fixed too.


colinleach (approved):

colinleach:
There are only 21 completions on this relatively recent exercise, so I'll let them re-test.

colinleach:
Thank you for fixing this!

PR #1126, opened 2026-07-27
[New practice exercise] Line Up

colinleach:
Much like the R version from earlier today.

### DIFF
--- a/config.json
+++ b/config.json
@@ -86,18 +86,6 @@
         "uuid": "5f181d2c-bb18-4f5d-b674-459674f7bd8e",
         "concepts": [
           "loops"
-    ]
-  },
-  "concepts": [
-    {
-      "uuid": "006ebce8-87cd-4695-87e6-8a7b8dc2f239",
-      "slug": "basics",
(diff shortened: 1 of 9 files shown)


config.json
```diff
@@ -1736,6 +1736,18 @@
...
+        "slug": "line-up",
+        "name": "Line Up",
+        "uuid": "656ea68f-30c4-41fe-ad1a-03e2969ac66e",
+        "practices": [
+          "strings"
+        ],
+        "prerequisites": [
+          "strings"
+        ],
+        "difficulty": 1
```
depial: Could we shift this in the list to be up with the other `difficulty: 1` exercises?

(This might also apply to the R version)

depial (approved):
Just the one suggestion!

config.json
colinleach: Thanks for spotting this. I forgot to clean up the `configlet create` addition.

R should be OK. I created those exercises in another (private) repo and copied them to the Exercism repo manually at the end.

PR #1127, opened 2026-07-28
Exercise sync, part 1

colinleach:
Running `configlet sync -u`  identified a few practice exercises that needed updating.

I added a template for Bob, but ran into errors when regenerating the tests: one of the string literals confuses the Julia parser.

```
ERROR: failed to parse input from /tmp/tmpe_xswoap: ParseError:
# Error @ line 86:38
    @testset "shouting with special characters" begin
        @test bob("""ZOMG THE %^*@#$(*^ ZOMBIES ARE COMING!!11!!1!""") == "Whoa, chill out!"
#                                    ╙ ── not a unary operator
```

A few potential workarounds failed, so I've marked that test as `include = false` in `tests.toml`. 

Conversely, a few older tests were excluded for reasons that no longer seem relevant, so I removed the `include = false` lines.

Finally, I added `.pyc` files (from the `generate_tests.py` runs) to `.gitignore`.

This seems enough complexity for one PR, so I'll template and update `connect`, `perfect-numbers` and `swift-scheduling` separately

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ tmp/
 Manifest.toml
 
 .idea/
+
+*.pyc
--- a/exercises/practice/bob/.meta/template.j2
+++ b/exercises/practice/bob/.meta/template.j2
@@ -1,0 +1,15 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.canonical_ref() }}
+
+{{ macros.header()}}
+{%- macro test_case(case) %}
+    {%- set input = case["input"] %}
+    @testset "{{ case["description"]}}" begin
+        @test bob("""{{ case["input"]["heyBob"] }}""") == "{{ case["expected"] }}"
+    end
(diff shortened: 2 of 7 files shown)


colinleach:
If you look at the old `bob/runtests.jl`, you'll see it creates all the tests as arrays at the top, then loops over them in testing. Very Julian (or very Sascha, for older test runner versions). Jeremy vocally hates this approach, so I replaced it with a more conventional linear test suite. We have quite a bit more of this we could do in the coming months, if you ever fancy a bit of gentle track housekeeping.

depial (approved):
LG2M!

The update you made for `runtests.jl` certainly makes a lot of sense for readability during troubleshooting, especially in the online editor.

colinleach:
I didn't prevent retesting (1026 completions for Bob): mainly because I forgot, but also these are big changes to the tests.

PR #1128, opened 2026-07-28
More templates: connect, swift-scheduling

colinleach:
Continues from #1127.

These were simpler than Bob, though displaying the board in `connect` took a few attempts to get Jinja2 to play nicely with `jlfmt`.

`perfect-numbers` had an update in #1127, but it turns out that was just a URL change in the documentation. Fortunate, because the current Julia implementation is wildly different from the `problem-specifications`. Three separate boolean functions here, versus a single function returning a text classifier in `canonical-data`. I'm not eager to step into that compatibility minefield.


depial (approved):

colinleach:
43 and 29 completions, respectively. We can afford to retest.

PR #1129, opened 2026-07-31
[New practice exercise] Simple Cipher

colinleach:
Templating this one is a real mess. It can be done entirely in Jinja2 (the Python track achieves this), but the template is then _really_ long and complex.

I auto-generate the easy ones, but provide hand-written tests for the non-standard ones.

R is missing a template for this exercise and I thought of adding one. I've now very much changed my mind about that.

### DIFF
--- a/exercises/practice/simple-cipher/.meta/example.jl
+++ b/exercises/practice/simple-cipher/.meta/example.jl
@@ -1,0 +1,26 @@
+using Random
+
+function generate_key()
+    Random.randstring('a':'z', 100)
+end
+
+encode(plaintext, key, enc=true) =
+    String([change(chr, inx, key, enc) for (inx, chr) in enumerate(collect(plaintext))])
+
+decode(ciphertext, key) = encode(ciphertext, key, false)
+
+function change(chr, inx, key, enc)
+    keys = collect(key)
+    while inx > length(keys)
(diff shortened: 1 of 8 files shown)


exercises/practice/simple-cipher/.meta/example.jl
```diff
@@ -0,0 +1,26 @@
+using Random
```
depial, line 1: I ended up with a broadcasting based solution. Here it is for reference:

```julia
generate_key() = join(rand('a':'z', 100))
fittedkey(text, key) = repeat(key, cld(length(text), length(key)))[1:length(text)]
numerictext(text) = Int.(collect(text))

function encode(plaintext, key)
    shiftednumeric = numerictext(plaintext) .+ numerictext(fittedkey(plaintext, key)) .- 194
    join(Char.(shiftednumeric .% 26 .+ 97))
end

function decode(ciphertext, key)
    shiftednumeric = numerictext(ciphertext) .- numerictext(fittedkey(ciphertext, key))
    join(Char.((x -> x < 0 ? 26 + x : x).(shiftednumeric) .+ 97))
end
```

depial (approved):
LG2M!

Good job! I likely would have just not even tried the template in this case 😅 It's good to have an example of how to do something like this.

PR #1133, opened 2026-08-02
[New practice exercise] Rectangles

colinleach:
I've gone with difficulty 4, but could be tempted to raise it to 5. Except Python (an outlier at 3), other tracks range between 4 and 7.

I think you may enjoy this one. Simple to state, but no single, obvious approach to solving it. My solution here borrows from Isaac's solution on Python, but changed so much that he would struggle to recognize it.

I hope my example is _somewhat_ like Julia. I've been using other languages so much in the last few months, my early drafts were just an ugly mixture of R and Python.

### DIFF
--- a/exercises/practice/rectangles/.meta/example.jl
+++ b/exercises/practice/rectangles/.meta/example.jl
@@ -1,0 +1,40 @@
+struct Pt
+    c::Integer  # column
+    r::Integer  # row
+end
+
+function rectangles(strings)
+    corners = Set{Pt}()
+    horiz = Set{Pt}()
+    vert = Set{Pt}()
+    for (y, line) in enumerate(strings)
+        for (x, char) in enumerate(line)
+            if char == '+'
+                push!(corners, Pt(x, y))
+                # corners are also valid for top & bottom
+                push!(horiz, Pt(x, y))
+                push!(vert, Pt(x, y))
+            elseif char == '-'
+                push!(horiz, Pt(x, y))
+            elseif char == '|'
+                push!(vert, Pt(x, y))
+            end
+        end
+    end
+
+    # get top-left and bottom-right corners
+    possibles = [(pt1, pt2) for pt1 in corners, pt2 in corners 
+                    if pt1.r < pt2.r && pt1.c < pt2.c
+                        # check bottom-left and top-right
+                        && Pt(pt1.c, pt2.r) ∈ corners 
+                        && Pt(pt2.c, pt1.r) ∈ corners]
+    
+    sides(pt1, pt2) = all([Pt(pt1.c, row) ∈ vert for row in pt1.r:pt2.r]) &&
+                        all([Pt(pt2.c, row) ∈ vert for row in pt1.r:pt2.r])
+
+    top_bottom(pt1, pt2) = all([Pt(col, pt1.r) ∈ horiz for col in pt1.c:pt2.c]) &&
+                            all([Pt(col, pt2.r) ∈ horiz for col in pt1.c:pt2.c])
+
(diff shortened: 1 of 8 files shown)


depial:
This does look like an interesting one! 

I'll have to give it a proper go tomorrow to get a better feeling for where I think the difficulty should be set.

exercises/practice/rectangles/.meta/example.jl
```diff
@@ -0,0 +1,40 @@
...
+                        && Pt(pt1.c, pt2.r) ∈ corners 
+                        && Pt(pt2.c, pt1.r) ∈ corners]
+    
+    sides(pt1, pt2) = all([Pt(pt1.c, row) ∈ vert for row in pt1.r:pt2.r]) &&
+                        all([Pt(pt2.c, row) ∈ vert for row in pt1.r:pt2.r])
+
+    top_bottom(pt1, pt2) = all([Pt(col, pt1.r) ∈ horiz for col in pt1.c:pt2.c]) &&
+                            all([Pt(col, pt2.r) ∈ horiz for col in pt1.c:pt2.c])
+
+    [(pt1, pt2) for (pt1, pt2) in possibles
```
depial, line 38: This could be made into a sum:

```julia
sum(sides(pt1, pt2) && top_bottom(pt1, pt2) for (pt1, pt2) in possibles)
```

depial (approved):
In the end, I could see this being a difficulty 5. Your comment about approaches being open is on point. I initially wanted to solve it with a graph traversal method, but I ended up going a different way:

<details>
<summary>My solution</summary>

```julia
function validcorners(strmatrix, invalid = " |")
    corners = []
    for column in 1:size(strmatrix, 2)
        valid = Set()
        for i in 1:size(strmatrix, 1)-1
            if strmatrix[i, column] == '+'
                for j in i+1:size(strmatrix, 1)
                    occursin(strmatrix[j, column], invalid) && break
                    strmatrix[j, column] == '+' && push!(valid, (i, j))
                end
            end
        end
        push!(corners, valid)
    end
    corners
end

function rectangles(strings)
    isempty(strings) && return 0
    strmatrix = stack(strings, dims=1)
    vcorners = validcorners(strmatrix, " -")
    hcorners = validcorners(permutedims(strmatrix))

    count = 0
    for i in 1:length(hcorners)-1
        for j in i+1:length(hcorners)
            sides = intersect(hcorners[i], hcorners[j])
            if !isempty(sides)
                count += sum((i, j) ∈ intersect(vcorners[side[1]], vcorners[side[2]]) for side in sides)
            end
        end
    end
    count
end
```

</details>

The overall strategy is to find the validly connected corners on the horizontal, find the validly connected corners on the vertical, and then compare (i.e. intersect) those two sets. 

Your `exemplar.jl` looks Julian to my eyes if that's worth anything :D
I also get the feeling that your implementation is less opaque than mine. If I could offer one minor suggestion, it would be to make the function names for `sides` and `top_bottom` a bit more descriptive. Maybe something like `valid_sides` or `connected_vert`? But this is not important, it could just make it a bit easier on first reading.

colinleach:
I increased the difficulty to 5, but after some tinkering decided to leave the example unchanged. Your suggestions were good, it's just that I'm in a grumpy mood after getting back from the supermarket: headache not helped by 105F temperatures and a thunderstorm approaching (plus age, decrepitude  and all the usual stuff...)

Meanwhile, I've been playing about with Crypto Square in Python and R. It looks like a useful one to add to Julia. Tomorrow...

Incidentally, I've no idea why GH thinks I started a review. I didn't, as far as I know.

depial:
Hope you feel better soon!

PR #1138, opened 2026-08-27
Add hint about approximate equality

depial:
I realized that I had forgotten to include a hint about approximate equality when dealing with floating point numbers in Task 3. This adds that hint.


depial:
I'm not sure if this hint is fully sufficient, since I don't specify what has to be within `1e-7` and, technically, there is more than one way to solve this... That said, I don't really know how to be more specific without pushing a particular solution and I used tests which don't really require much floating point precision (i.e. all the directional vector inputs are integers and those need to be compared with whole numbers).

If you have any thoughts for better wording, feel free to let me know. Otherwise, I think it's reasonable to just hope that people get the tolerances right without too much difficulty.

colinleach (approved):

colinleach:
> If you have any thoughts for better wording

Sorry, no

PR #1139, opened 2026-08-30
[New practice exercise] Affine Cipher

colinleach:
I went with difficulty 4, because Julia provides `gcd()` and `invmod()` as standard (the latter is what we called `mmi()` in the R version).

Unfortunately, quirky behavior of the `%` operator wasted significant time and energy. The R and Python solutions don't just copy across.

### DIFF
--- a/exercises/practice/affine-cipher/runtests.jl
+++ b/exercises/practice/affine-cipher/runtests.jl
@@ -1,0 +1,81 @@
+# These tests are auto-generated with test data from:
+# https://github.com/exercism/problem-specifications/tree/main/exercises/affine-cipher/canonical-data.json
+# File last updated on 2026-08-29
+
+using Test
+
+include("affine-cipher.jl")
+
+@testset verbose = true "tests" begin
+    @testset "encode" begin
+        @testset "encode yes" begin
+            @test encode("yes", 5, 7) == "xbt"
+        end
+
+        @testset "encode no" begin
+            @test encode("no", 15, 18) == "fu"
+        end
+
+        @testset "encode OMG" begin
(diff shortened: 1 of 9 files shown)


exercises/practice/affine-cipher/runtests.jl
```diff
@@ -0,0 +1,81 @@
...
+            @test encode("Truth is fiction.", 5, 17) == "iynia fdqfb ifje"
+        end
+
+        @testset "encode all the letters" begin
+            @test encode("The quick brown fox jumps over the lazy dog.", 17, 33) ==
+                  "swxtj npvyk lruol iejdc blaxk swxmh qzglf"
+        end
+
+        @testset "encode with a not coprime to m" begin
+            @test_throws ArgumentError encode("This is a test.", 6, 17)
```
depial, line 45: I noticed that Julia throws a `DomainError` when using non-coprime numbers in `invmod`, but I'm okay with `ArgumentError` as well.

depial (approved):
I actually took the time to solve it this time which made me think that this could use a `hints.md`. I think we were trying to include one with new practice exercises, right? I think it could be helpful here since some people may not know about `invmod` and `gcd` being available.

exercises/practice/affine-cipher/runtests.jl
colinleach: I've no strong feelings about which to use. ArgumentError may be a bit simpler, DomainError may be a bit more correct.

colinleach: After more thought, I'm inclined to stick with ArgumentError. It refers to our function's arguments, not `invmod`.

